PR #60, opened 2015-05-12
Make the property `searchTextInput` an alias of `textInput.text` to a…

tdunn:
…void having to maintain and synchronize the same text in two places.


ldanzinger:
looks good @tdunn. Can you please also submit a PR for the example update and for the v.next branch?

PR #61, opened 2015-05-12
Add and demonstrate ability to change starting search text in SearchBox.

tdunn:


ldanzinger:
looks good, thanks @tdunn

PR #64, opened 2015-06-12
Quartz version and namespace changes

michael-tims:
Assign to @ldanzinger 

Quartz version and namespace changes.


ldanzinger:
@michael-tims was this supposed to upgrade the toolkit to 1.1 as well or does that happen later?

michael-tims:
@ldanzinger I pushed up changes to update the toolkit version number.

PR #72, opened 2015-08-07
Have StyleButton use map as a property

jf990:
Changing NavigationToolbarExample to use mainMap caused the buttons to lose the reference to the map. I propose changing StyleButton to take the map as a parameter (property) instead of relying on it being a specific id

### DIFF
--- a/Examples/NavigationToolbarExample.qml
+++ b/Examples/NavigationToolbarExample.qml
@@ -40,6 +40,7 @@ Rectangle {
 
         NavigationToolbar {
             id: navigationToolbar
+            map: mainMap
             anchors {
                 top: parent.top
                 left: parent.left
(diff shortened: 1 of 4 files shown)


ldanzinger:
@jf990 do any of the examples that use navigation toolbar need to get updated to set the map property?

jf990:
yes, i missed it because running the sample take the toolkit from one location and the examples from a different, so the change missed the repo. I'll add it.

ldanzinger:
I think StyleTemplateExample might also use this. Can you update any other instances of it in the example?

jf990:
I made the correction to StyleTemplateExample and test it. I didn't find any other samples that reference it but if you notice anything else I missed LMK.

ldanzinger:
looks good, thanks @jf990

PR #85, opened 2016-08-06
Nand6028/callout

nandinirao:
Assigned to @michael-tims 

Show or dismiss callout based on CalloutData's visible property.


michael-tims:
Looks good.

PR #89, opened 2016-08-18
removing QML API import. 

ldanzinger:
The callout will work with either C++ or QML APIs

@michael-tims please review and merge

cc @nandinirao


nandinirao:
Makes sense. Thanks for catching this!

michael-tims:
Looks good.

PR #94, opened 2016-09-09
Changes based on Luke's feedback

ryankim86:
Edits include:
- modified buttons to use icon images instead of text
- slight radius to rectangles
- model declaratively set in example
- height and width now have default values
- documentation edits
- set a MouseArea around item so mouse actions do not propagate down

@ldanzinger please review and merge, thanks.


ldanzinger:
looks good

PR #101, opened 2016-09-30
The attribute list model now re-notified once the feature is loaded. …

JamesMBallard:
…We no longer need to assign the list model only once loaded. Also update the out of date identify methods.

Assign to @michael-tims 

We can use property binding to update the list model data now. The list model now updates if it's hooked up to a loadable feature and the feature finishes loading.

attributes


JamesMBallard:
Merging to `v.next`

michael-tims:
Looks good.

PR #104, opened 2016-10-20
Nand6028/callout ios fixes

nandinirao:
Assigned to @michael-tims 

Since rectWidth and rectHeight already had devicePixelRatio accounted for, it was putting the callout in the wrong location. I changed it a bit to use the calloutLayout's width and height instead only for ios. Other platforms will continue to use rectWidth and rectHeight. 

Ideally I don't want to have these platform specific checks but may be we can clean it up better after we release since this is only toolkit specific.

### DIFF
--- a/Imports/ArcGIS/Runtime/Toolkit/Controls.2.0/Callout.qml
+++ b/Imports/ArcGIS/Runtime/Toolkit/Controls.2.0/Callout.qml
@@ -299,7 +299,10 @@ Item {
         // these are some of the initial calculations
         // before creating the callout frame
         preCalculateWidthAndHeight();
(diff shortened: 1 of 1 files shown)


michael-tims:
I assume this is going to be a problem on Mac with retina display also.  Can you please check?

nandinirao:
Tested on my Mac book pro, iphone 6, android phone and windows. ios was the only one having the issue.

michael-tims:
>  Tested on my Mac book pro

Ok, and the Mac book pro was not hooked up to external monitor?

nandinirao:
Nope. Just on its own not on the thunderbolt.

michael-tims:
Awesome.  Change is good.

PR #105, opened 2016-10-24
James/target scale

JamesMBallard:
Assigned to @michael-tims 

The name of this property is `targetScale` which matches the C++ Api. I also fixed up an unrelated problems with the Callout example.

Please review. Merging to v.next.


michael-tims:
Looks good.

PR #118, opened 2017-01-10
James/adjust auto leader line

JamesMBallard:
Assign to @nandinirao. Please review.

This is adjusting how the Automatic leader line setting works. Now it will always try to default to `Bottom` and adjust from there if it won't fit in the frame. I had to switch making the vertical and horizontal adjustments so it would work in the upper corners.

This is for v.next.

cc: @ldanzinger

### DIFF
--- a/Imports/ArcGIS/Runtime/Toolkit/Controls.2.0/Callout.qml
+++ b/Imports/ArcGIS/Runtime/Toolkit/Controls.2.0/Callout.qml
@@ -323,6 +323,9 @@ Item {
                 adjustRelativePositionOfCanvasFrame(anchorPointx, anchorPointy, rectWidth, rectHeight);
         }
 
+        if (leaderPosition !== Enums.LeaderPosition.Automatic)
+            adjustedLeaderPosition = leaderPosition
+
         // paint now.
         canvas.createPathAndPaint = true;
         canvas.requestPaint(true);
--- a/Imports/ArcGIS/Runtime/Toolkit/Controls.2.0/Callout.qml
+++ b/Imports/ArcGIS/Runtime/Toolkit/Controls.2.0/Callout.qml
@@ -296,9 +296,9 @@ Item {
 
         // set the adjustedLeaderPosition
         if (leaderPosition !== Enums.LeaderPosition.Automatic)
-            adjustedLeaderPosition = leaderPosition
+            adjustedLeaderPosition = leaderPosition;
         else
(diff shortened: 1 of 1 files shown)


Imports/ArcGIS/Runtime/Toolkit/Controls.2.0/Callout.qml
```diff
@@ -603,6 +605,15 @@ Item {
 
         if (leaderPosition === Enums.LeaderPosition.Automatic) {
 
+            // Move leader vertically if vertical position isn't optimal
```
nandinirao: @JamesMBallard Last time when I tried checking vertical first, I was hitting other issues with right and left positioning not working ok. Have you tested all 8 positions?

nandinirao: And please make sure you also try the devices.

Imports/ArcGIS/Runtime/Toolkit/Controls.2.0/Callout.qml
```diff
@@ -296,7 +296,9 @@ Item {
 
         // set the adjustedLeaderPosition
         if (leaderPosition !== Enums.LeaderPosition.Automatic)
-            adjustedLeaderPosition = leaderPosition
+            adjustedLeaderPosition = leaderPosition;
+        else
+            adjustedLeaderPosition = Enums.LeaderPosition.Bottom;
```
nandinirao: Starting from bottom makes sense.

Imports/ArcGIS/Runtime/Toolkit/Controls.2.0/Callout.qml
JamesMBallard: @nandinirao I tested all the positions on Android and iOS Simulator. Android looks good but there are some issues with iOS where I can't test very well. I chatted with @ldanzinger and that's a known issue right now. It was certainly no worse on iOS.

JamesMBallard:
There is a bit of a problem on Android. I'll look into fixing that. When you move towards the edges of the screen the callout changed the leader line position way too early (edge detection calculations may be off).

JamesMBallard:
@nandinirao I've updated the PR and retested on Windows, Android and Mac. All looks good now.

Please give it another review.

nandinirao:
Yes, this looks good now. Thanks!

PR #124, opened 2017-02-02
Michael tims/callout with insets support

michael-tims:
Assign to @JamesMBallard 

Updates for Callout to support view insets.  A new Callout file was created which will be used for the 2.1 release of the Toolkit.  A snapshot of the Callout for 2.0 was created as Callout20.qml which will be used if the user imports `Esri.ArcGISRuntime.Toolkit.Controls 2.0`.  QDoc will only pick up the current Callout doc, the prior doc version of Callout should be hidden (internal).


JamesMBallard:
Code is good.

PR #130, opened 2017-02-24
Nand6028/callout ios workaround

nandinirao:
Assigned to @michael-tims 
FYI @ldanzinger 

Using a workaround suggested by Qt company for iOS. It seems the bug we see with the sizes is specific to using a renderTarget of Canvas.FrameBufferObject. 

I am using the default of Canvas.Image only for iOS and also cleaning up the other platform specific checks since we don't need that anymore.


michael-tims:
@nandinirao Thanks.  This looks good.

PR #131, opened 2017-03-16
changing versioning scheme to match the Runtime API

ldanzinger:
@michael-tims please review and merge


michael-tims:
@ldanzinger Is it correct that we should be renaming some of the files to contain `20` in their names?  I thought rename is only necessary if we had a new version of that type.  Or maybe the 100.0 versions are coming in later?

ldanzinger:
@michael-tims I thought the cleanest way to do it was to rename all of the 2.0 files to include `20` in their names, so there is no confusion going forward. Those 2.0 versions will be deprecated and remain as-is. It was an attempt to draw a separation between the 2.0 series and 100.0 series

michael-tims:
@ldanzinger Maybe we need 100.0 versions also?  I'm concerned with how the documentation will look for types containing "20" in their names.

michael-tims:
Thanks @ldanzinger , per our discussion, this versioning looks good.

PR #132, opened 2017-04-07
There is an issue on Linux with using the ternary operator here. It's…

JamesMBallard:
… unclear why but it causes a crash in some scenarios.

Assigned to @nandinirao. As discussed, we're seeing strange behavior on Linux with the ternary operator. It's unclear why it was crashing but it seemed to be crashing in qt internals.

Tested on all platforms for regressions by @ldanzinger.


nandinirao:
Yes, strange one. Thanks for fixing and testing @JamesMBallard @ldanzinger

PR #140, opened 2017-06-09
Plugin documentation.

JamesMBallard:
Document how the Toolkit may be imported directly into an application,
which will be required on iOS starting with Qt 5.9.0.


JamesMBallard:
Updated contents approved by @ldanzinger.

PR #144, opened 2017-09-01
James/common toolkit

JamesMBallard:
Assign to @michael-tims 

Merging to v.next. This is the initial implementation of the Coordinate Conversion tool, integrated into our existing toolkit. The new pieces are for the Cpp Api only.


.gitignore
```diff
@@ -140,3 +140,6 @@ ehthumbs.db
 Thumbs.db
 
 *.user.2.7pre1
+*.exp
+Solutions/CppApi/output/
```
michael-tims: `Solutions/CppApi/output/` can be cleaned up.

Imports/ArcGIS/Runtime/Toolkit/Controls/CppApi/CoordinateConversion.qml
```diff
@@ -0,0 +1,128 @@
...
+import Esri.ArcGISRuntime.Toolkit.CppApi 100.2
+
+Rectangle {
+    id: coordinateConversionWindow
+    height: 800
+    width: 600
+
+    property bool getFromMapMode: false
+    property int buttonWidth: 100
+    property int spacingVal: 5
```
michael-tims: I assume these are properties that we'd like to document what they do.  Since that's the case, they should probably have clearer names. Usually booleans don't start with `get`.  `spacingVal` should be spelled out `spacingValue`.  Also these pixel values need to have scaleFactor applied to them.  If the UI isn't fleshed out then we can spin these tweaks up as a separate issue.

Imports/ArcGIS/Runtime/Toolkit/Controls/CppApi/CoordinateConversion.qml
```diff
@@ -0,0 +1,128 @@
...
+
+    CoordinateConversionController {
+        id: controller
+        objectName: "coordinateConversionController"
+    }
+
+    ScrollView {
+        id: scrollView
+        anchors.fill: parent
+        anchors.margins: 10
```
michael-tims: Needs scaleFactor applied.  I'll stop commenting on these.

JamesMBallard:
Thanks for the feedback @michael-tims. I've applied the requested changes. The UI is simple at the moment and will be improved as we iterate and apply the general look and feel to all the tools.

PR #152, opened 2017-09-15
ArcGISCompass Inherits from AbstractTool

ryankim923:
Made `ArcGISCompass` inherit from AbstractTool and added the appropriate import statement to QML view.

@JamesMBallard Please review and merge. Thank you.

### DIFF
--- a/Plugin/CppApi/source/ArcGISCompassController.cpp
+++ b/Plugin/CppApi/source/ArcGISCompassController.cpp
@@ -113,12 +113,12 @@ double ArcGISCompassController::heading() const
 
 bool ArcGISCompassController::autoHide() const
 {
-    return m_autoHide;
+  return m_autoHide;
 }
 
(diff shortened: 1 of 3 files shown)


Plugin/CppApi/source/ArcGISCompassController.cpp
```diff
@@ -113,7 +113,12 @@ double ArcGISCompassController::heading() const
 
 bool ArcGISCompassController::autoHide() const
 {
-  return m_autoHide;
+    return m_autoHide;
+}
+
+QString ArcGISCompassController::toolName() const
+{
+    return "ArcGISCompass";
```
JamesMBallard: Please use consistent 2-space indent.

ryankim923: Good catch. Done.

PR #161, opened 2017-09-27
James/toolkit restructure

JamesMBallard:


JamesMBallard:
Move some recently added files into the new folder structure. 

cc @ryankim86

PR #166, opened 2017-09-29
add properties to AbstractTool interface

lsmallwood:
@michael-tims please review the update to the abstract tool interface

### DIFF
--- a/Plugin/CppApi/include/AbstractTool.h
+++ b/Plugin/CppApi/include/AbstractTool.h
@@ -17,6 +17,7 @@
 
 #include <QObject>
 #include <QString>
+#include <QVariantMap>
 
 namespace Esri
 {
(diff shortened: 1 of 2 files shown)


Plugin/CppApi/include/AbstractTool.h
```diff
@@ -39,11 +40,14 @@ class TOOLKIT_EXPORT AbstractTool : public QObject
...
   virtual bool handleClick(const Point& pos);
 
+  virtual void setProperties(const QVariantMap& properties);
+
   virtual void setActive(bool active);
   bool isActive() const;
 
 signals:
   void errorOccurred(const Error& error);
+  void propertyUpdated(const QString& propertyName, const QVariant& propertyValue);
```
michael-tims: @lsmallwood Can we call this `propertyChanged` instead?  Usually `*Changed` is what I typically see.

PR #171, opened 2017-10-03
James/coordinate conversion

JamesMBallard:
Updates for the coordinate conversion tool.

Code review: @michael-tims 
Doc review: @tdunn 

Deleted images are cleanup from a previous merge. They already exist in the [correct location](https://github.com/Esri/arcgis-runtime-toolkit-qt/tree/v.next/Import/Esri/ArcGISRuntime/Toolkit/Controls/CppApi/images).

Notes:
 * CoordinateConversionResults methods are all made private so it's clearer that the class is only meant to be used by the CoordinateConversionController class. This clears up some possible confusing with having `clearResults` methods on both the controller and the results class.
 * Internal helper class `Result` no longer exported (`TOOLKIT_EXPORT` removed). This shouldn't be accessed or used outside of these classes.
 * Doc added throughout.
 * Consolidation of the coordinate type strings like "Gars" and the others.
 * Extensive documentation added for CoordinateConversionOptions detailing which groups of options are valid (and what their valid ranges are) for all types.

cc: @khajra


Plugin/CppApi/source/CoordinateConversionController.cpp
```diff
@@ -27,11 +42,66 @@ namespace Toolkit
...
+      {
+        setSpatialReference(mapView->spatialReference());
+      });
+
+      connect(mapView, &Esri::ArcGISRuntime::MapQuickView::mouseClicked, this,
+      [mapView, this](QMouseEvent& mouseEvent)
+      {
+        setPointToConvert(mapView->screenToLocation(mouseEvent.x(), mouseEvent.y()));
+      });
+    }
```
michael-tims: `spatialReferenceChanged` and `mouseClicked` signals should be moved to the new `ToolResourceProvider`

Plugin/CppApi/source/CoordinateConversionController.cpp
```diff
@@ -262,39 +424,71 @@ void CoordinateConversionController::objectAppend(QQmlListProperty<QObject>* pro
...
 
+/*!
+  \qmlmethod void CoordinateConversionController::copyToClipboard(string text)
+
+  \brief Helper method to copy \a text to the clipboard.
+ */
 void CoordinateConversionController::copyToClipboard(const QString& text)
 {
-  QApplication::clipboard()->setText(text);
+  auto clipboard = QApplication::clipboard();
```
michael-tims: `QApplication` brings in Qt Widgets dependency.  Use `QGuiApplication::clipboard()` instead.

JamesMBallard:
@michael-tims updated.

PR #175, opened 2017-10-10
change to properties instead of Q_INVOKABLE.

khajra:
- change the compass size a bit
- remove the extras dependency
- modified the setters so that they are not Q_INVOKABLES

### DIFF
--- a/Plugin/CppApi/include/ArcGISCompassController.h
+++ b/Plugin/CppApi/include/ArcGISCompassController.h
@@ -41,10 +41,6 @@ signals:
   void autoHideChanged();
 
 public:
-  Q_INVOKABLE void setHeading(double rotation);
-  Q_INVOKABLE void setAutoHide(bool autoHide);
-
-public:
   ArcGISCompassController(QObject *parent = nullptr);
   ~ArcGISCompassController();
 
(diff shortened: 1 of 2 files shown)


Plugin/CppApi/include/ArcGISCompassController.h
```diff
@@ -62,6 +58,8 @@ class TOOLKIT_EXPORT ArcGISCompassController : public AbstractTool
   bool m_autoHide = true;
   Esri::ArcGISRuntime::MapQuickView* m_mapView = nullptr;
   Esri::ArcGISRuntime::SceneQuickView* m_sceneView = nullptr;
+  void setAutoHide(bool value);
+  void setHeading(double value);
```
JamesMBallard: The getters are public so the setters can be public too.

Plugin/CppApi/include/ArcGISCompassController.h
```diff
@@ -53,7 +49,10 @@ class TOOLKIT_EXPORT ArcGISCompassController : public AbstractTool
   bool setView(Esri::ArcGISRuntime::SceneQuickView* sceneView);
 
   double heading() const;
+  void setHeading(const double value);
+
   bool autoHide() const;
+  void setAutoHide(const bool value);
```
JamesMBallard: Both setters should just take args by copy, not const.

khajra: That's what I have.  Pushed incorrect commit :(

PR #177, opened 2017-10-16
signal fwds

khajra:
- add signals to forward so that other classes can handle accordingly.

### DIFF
--- a/Plugin/CppApi/include/ToolResourceProvider.h
+++ b/Plugin/CppApi/include/ToolResourceProvider.h
@@ -74,6 +74,7 @@ public slots:
   void onMousePressedAndHeld(QMouseEvent& mouseEvent);
   void onMouseDoubleClicked(QMouseEvent& mouseEvent);
   void onIdentifyGraphicsOverlayCompleted(QUuid id, IdentifyGraphicsOverlayResult* identifyResult);
+  void onscreenToLocationCompleted(QUuid taskId, Point location);
 
 signals:
   void sceneChanged();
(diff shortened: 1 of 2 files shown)


Plugin/CppApi/source/ToolResourceProvider.cpp
```diff
@@ -238,6 +238,15 @@ void ToolResourceProvider::onIdentifyGraphicsOverlayCompleted(QUuid id, Identify
   emit identifyGraphicsOverlayCompleted(id, identifyResult);
 }
 
+void ToolResourceProvider::onscreenToLocationCompleted(QUuid taskId, Point location)
+{
+  if (!m_geoView)
+    return;
+
+  if (m_scene)
+    emit screenToLocationCompleted(taskId, location);
```
michael-tims: I'm not sure the value in actually checking for `m_geoView` and `m_scene` here.

Plugin/CppApi/include/ToolResourceProvider.h
```diff
@@ -74,6 +74,7 @@ public slots:
   void onMousePressedAndHeld(QMouseEvent& mouseEvent);
   void onMouseDoubleClicked(QMouseEvent& mouseEvent);
   void onIdentifyGraphicsOverlayCompleted(QUuid id, IdentifyGraphicsOverlayResult* identifyResult);
+  void onscreenToLocationCompleted(QUuid taskId, Point location);
```
michael-tims: @khajra Sorry, missed this the first time around, but this should be `void onScreenToLocationCompleted(QUuid taskId, const Point& location);`

Plugin/CppApi/include/ToolResourceProvider.h
```diff
@@ -82,6 +83,7 @@ public slots:
   void spatialReferenceChanged();
   void identifyGraphicsOverlayCompleted(QUuid id, IdentifyGraphicsOverlayResult* identifyResult);
   void setMouseCursorRequested(const QCursor& cursor);
+  void screenToLocationCompleted(QUuid taskId, Point location);
```
michael-tims: And then... `void screenToLocationCompleted(QUuid taskId, const Point& location);`

PR #189, opened 2017-12-08
Fix incorrect links in doc. Doc change only!

tdunn:
@michael-tims - Please review and merge. Thank you.


Import/Esri/ArcGISRuntime/Toolkit/Controls/CppApi/ArcGISCompass.qml
```diff
@@ -8,11 +8,9 @@ import Esri.ArcGISRuntime.Toolkit.CppApi 100.2
     \qmltype ArcGISCompass
     \ingroup ToolArcGISCompass
     \inqmlmodule Esri.ArcGISRuntime.Toolkit.Controls
-    \since Esri.ArcGISRutime 100.2
+    \since Esri.ArcGISRutime 100.2.0
```
michael-tims: @tdunn We've been just doing `\since Esri.ArcGISRutime 100.2` for the API.  Should we do the same for the toolkit?

tdunn:
Good point. Fixed.

PR #196, opened 2018-01-17
Toolkit Identify:

lsmallwood:
@michael-tims please review the addition of the signals to the toolkit

### DIFF
--- a/Plugin/CppApi/include/ToolResourceProvider.h
+++ b/Plugin/CppApi/include/ToolResourceProvider.h
@@ -74,8 +74,8 @@ public slots:
   void onMouseReleased(QMouseEvent& mouseEvent);
   void onMousePressedAndHeld(QMouseEvent& mouseEvent);
   void onMouseDoubleClicked(QMouseEvent& mouseEvent);
-  void onIdentifyGraphicsOverlayCompleted(QUuid id, IdentifyGraphicsOverlayResult* identifyResult);
-  void onIdentifyGraphicsOverlaysCompleted(QUuid id, QList<IdentifyGraphicsOverlayResult*> identifyResults);
+  void onIdentifyGraphicsOverlayCompleted(QUuid taskId, IdentifyGraphicsOverlayResult* identifyResult);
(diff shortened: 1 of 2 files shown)


Plugin/CppApi/include/ToolResourceProvider.h
```diff
@@ -74,14 +75,20 @@ public slots:
   void onMousePressedAndHeld(QMouseEvent& mouseEvent);
   void onMouseDoubleClicked(QMouseEvent& mouseEvent);
   void onIdentifyGraphicsOverlayCompleted(QUuid id, IdentifyGraphicsOverlayResult* identifyResult);
+  void onIdentifyGraphicsOverlaysCompleted(QUuid id, QList<IdentifyGraphicsOverlayResult*> identifyResults);
+  void onIdentifyLayerCompleted(QUuid taskId, IdentifyLayerResult* identifyResult);
```
ldanzinger: should we keep the parameter consistently `id` throughout?

Plugin/CppApi/include/ToolResourceProvider.h
```diff
@@ -74,14 +75,20 @@ public slots:
...
+  void onIdentifyLayersCompleted(QUuid taskId, QList<IdentifyLayerResult*> identifyResults);
   void onScreenToLocationCompleted(QUuid taskId, const Point& location);
 
 signals:
   void sceneChanged();
   void geoViewChanged();
   void mapChanged();
   void spatialReferenceChanged();
-  void identifyGraphicsOverlayCompleted(QUuid id, IdentifyGraphicsOverlayResult* identifyResult);
```
ldanzinger: i don't think we want to change the signature of the signal, do we? Maybe leave this one and use `id` for all the others?

Plugin/CppApi/include/ToolResourceProvider.h
lsmallwood: i think taskId is more descriptive

ldanzinger:
taskId sounds good. merging

PR #215, opened 2018-02-02
Use the cached vars consistently.

JamesMBallard:
Assigned to @khajra.

Some code cleanup porting back from the release/100.2.1 PR. Good catch 👍


khajra:
👍

PR #226, opened 2018-02-15
Add API ref and remove tools when they are destroyed

lsmallwood:
@michael-tims, @ldanzinger - please review the update to the ToolManager to remove tools which have been destroyed

### DIFF
--- a/Plugin/CppApi/include/ToolManager.h
+++ b/Plugin/CppApi/include/ToolManager.h
@@ -42,6 +42,8 @@ public:
 
   void removeTool(const QString& toolName);
 
+  void removeTool(AbstractTool* tool);
+
   void clearTools();
 
   AbstractTool* tool(const QString& toolName) const;
--- a/Plugin/CppApi/include/ToolManager.h
+++ b/Plugin/CppApi/include/ToolManager.h
@@ -40,6 +40,10 @@ public:
 
   void addTool(AbstractTool* tool);
 
+  void removeTool(const QString& toolName);
+
+  void clearTools();
+
   AbstractTool* tool(const QString& toolName) const;
 
   template<class T>
--- a/Plugin/CppApi/source/ToolManager.cpp
+++ b/Plugin/CppApi/source/ToolManager.cpp
@@ -60,16 +60,7 @@ void ToolManager::addTool(AbstractTool* tool)
   // when a tool is destroyed, remove it from the manager
   QObject::connect(tool, &AbstractTool::destroyed, tool, [this, tool]()
   {
-    auto it = m_tools.begin();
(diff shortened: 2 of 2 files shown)


Plugin/CppApi/include/ToolManager.h
```diff
@@ -40,6 +40,10 @@ class TOOLKIT_EXPORT ToolManager {
 
   void addTool(AbstractTool* tool);
 
+  void removeTool(const QString& toolName);
```
ldanzinger: Do we want to have a version of this function that takes in an AbstractTool instance? I wonder if you'd want to ensure you are removing a specific instance of a tool and not just some instance of a tool that is in the tool manager (in the case that you have multiple AddLocalDataToolControllers, for example)

lsmallwood: I can certainly add that. At the moment the map only allows 1 tool per name so you can't have multiple AddLocalDataToolControllers...

ldanzinger: probably not necessary if that is how the map behaves. Is there a use case for having multiple instances of a tool controller? I was wondering this a while back. Maybe I want the Coordinate Conversion Tool (with UI) in my app, but then I want to also be able to directly use the Coordinate Conversion controller from the C++ side in a different tool?

lsmallwood: I'll add it anyway since it's essentially the lambda we run when a tool gets destroyed. I don't know about multiple tools. Tricky to see how that would work. I guess each could have a unique name/ID?

Plugin/CppApi/source/ToolManager.cpp
```diff
@@ -21,51 +21,120 @@ namespace ArcGISRuntime
...
 void ToolManager::addTool(AbstractTool* tool)
 {
   if (!tool)
     return;
 
+  // when a tool is destroyed, remove it from the manager
+  QObject::connect(tool, &AbstractTool::destroyed, tool, [this, tool]()
+  {
+    removeTool(tool);
+  });
```
michael-tims: @lsmallwood Were you able to test that this works?

Plugin/CppApi/source/ToolManager.cpp
```diff
@@ -21,51 +21,120 @@ namespace ArcGISRuntime
...
+
+/*! \brief Removes the \l AbstractTool \a tool from the manager.
+ */
+void ToolManager::removeTool(AbstractTool* tool)
+{
+  if (tool == nullptr)
+    return;
+
+  auto it = m_tools.begin();
+  auto itEnd = m_tools.end();
```
michael-tims: Should these be const iterators?

PR #229, opened 2018-02-16
update the ToolResourceProvider to clear members when they are destroyed

lsmallwood:
@michael-tims - please review the change to clean up potential dangling ptrs

### DIFF
--- a/Plugin/CppApi/source/ToolResourceProvider.cpp
+++ b/Plugin/CppApi/source/ToolResourceProvider.cpp
@@ -52,14 +52,33 @@ Map* ToolResourceProvider::map() const
   return m_map;
 }
 
+/*! \brief Sets the map to \a newMap.
+ *
+ * To clear the map pass \c nullptr.
+ *
+ * \sa mapChanged.
+ */
 void ToolResourceProvider::setMap(Map* newMap)
 {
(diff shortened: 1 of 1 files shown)


Plugin/CppApi/source/ToolResourceProvider.cpp
```diff
@@ -52,46 +52,104 @@ Map* ToolResourceProvider::map() const
...
     return;
 
   m_map = newMap;
 
   emit mapChanged();
+
+  if (m_map == nullptr)
+    return;
+
+  QObject* mapObject = dynamic_cast<QObject*>(m_map);
```
michael-tims: I don't think is necessary because Map is already a QObject

Plugin/CppApi/source/ToolResourceProvider.cpp
```diff
@@ -52,46 +52,104 @@ Map* ToolResourceProvider::map() const
...
     return;
 
   m_scene = newScene;
 
   emit sceneChanged();
+
+  if (m_scene == nullptr)
+    return;
+
+  QObject* sceneObject = dynamic_cast<QObject*>(m_scene);
```
michael-tims: Same here, Scene is already a QObject

PR #296, opened 2018-06-27
Fix some API doc build warnings

tdunn:
@michael-tims Please review and merge.

### DIFF
--- a/Plugin/CppApi/source/CoordinateConversion/CoordinateConversionController.cpp
+++ b/Plugin/CppApi/source/CoordinateConversion/CoordinateConversionController.cpp
@@ -381,11 +381,14 @@ bool CoordinateConversionController::isCaptureMode() const
 }
 
 /*!
-  \brief Sets the tool's capture mode to \a captureMode.
-
-  If \c true, the tool will convert a point set via a mouse click.
-  If \c false, the too will use the app's current location as the target point.
+  \qmlproperty bool CoordinateConversionController::captureMode
+  \brief The tool's capture mode
+  \list
+    \li \c true = convert a point set via a mouse click.
+    \li \c false = use the app's current location as the target point.
+  \endlist
  */
+
 void CoordinateConversionController::setCaptureMode(bool captureMode)
 {
   if (captureMode == m_captureMode)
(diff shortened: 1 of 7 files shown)


Plugin/CppApi/source/CoordinateConversion/CoordinateConversionController.cpp
```diff
@@ -381,11 +381,14 @@ bool CoordinateConversionController::isCaptureMode() const
...
-  If \c true, the tool will convert a point set via a mouse click.
-  If \c false, the too will use the app's current location as the target point.
+  \qmlproperty bool CoordinateConversionController::captureMode
+  \brief The tool's capture mode
+  \list
+    \li \c true = convert a point set via a mouse click.
+    \li \c false = use the app's current location as the target point.
+  \endlist
  */
+
```
michael-tims: @tdunn For the C++ toolkit components, I believe we need the public setters documented as well as the public properties.  So the setter doc can remain and then the `qmlproperty` doc added to the bottom of the file.

Plugin/CppApi/source/CoordinateConversion/CoordinateConversionController.cpp
```diff
@@ -399,9 +402,11 @@ void CoordinateConversionController::setCaptureMode(bool captureMode)
 }
 
 /*!
-  \brief Handles the mouse click at \a mouseEvent .
+  \fn void CoordinateConversionController::onMouseClicked(QMouseEvent& mouseEvent);
+  \brief Signal emitted when the mouse is clicked.
 
-  If the tool is active and in \l captureMode, this will be used as the input for conversions.
+  If the tool is active and is in \l captureMode, the clicked location will be used
+  as the input for conversions.
```
michael-tims: `onMouseClicked` is actually a slot.  So it should more be handled like a regular method that gets called when the view's `mouseClicked` signal occurs.  Something like that.

Plugin/CppApi/source/CoordinateConversion/CoordinateConversionController.cpp
```diff
@@ -415,9 +420,11 @@ void CoordinateConversionController::onMouseClicked(QMouseEvent& mouseEvent )
 }
 
 /*!
-  \brief Handles the app's location update to \a location.
+  \fn void CoordinateConversionController::onLocationChanged(const Point& location);
+  \brief Signal emitted when the \l location property changes.
 
-  If the tool is active and is not in \l captureMode, this will be used as the input for conversions.
+  If the tool is active and is not in \l captureMode, the location will be used
+  as the input for conversions.
```
michael-tims: Same with this slot.

PR #299, opened 2018-08-16
James/set settings earlier

JamesMBallard:
Assigned to @ldanzinger.

@lsmallwood and @ldanzinger to review.

Changes for **master** to allow the tool manager to notify watchers when tools are added. This goes along with https://github.com/Esri/dynamic-situational-awareness-qt/pull/283.


Plugin/CppApi/include/ToolManager.h
```diff
@@ -61,6 +63,9 @@ class TOOLKIT_EXPORT ToolManager {
   ToolsList::const_iterator begin() const;
   ToolsList::const_iterator end() const;
 
+signals:
+  void toolAdded(Esri::ArcGISRuntime::Toolkit::AbstractTool* tool);
```
lsmallwood: seems like we should maybe have a companion signal that notifies when a tool is removed as well?

JamesMBallard: @lsmallwood I don't know if that's reliable. We'd be relying on the QObject::destroyed signal, and that pointer can go away so I don't know if it's safe to send that pointer out on the signal, and having no information on the signal isn't helpful at all.

We do something related in `ToolManager::removeTool`, but all we need is the pointer address to remove it from our map, it doesn't matter if it's a dangling pointer or not at that point.

JamesMBallard: @lsmallwood it should be possible to have a removed signal with the tool name, but it wouldn't match the 'added' signal. What do you think? We can get the name of the tool from our `ToolsList` member when we're removing it from the ToolManager. It would look something like this:

```cpp
signals:
  void toolAdded(Esri::ArcGISRuntime::Toolkit::AbstractTool* tool);
  void toolRemoved(const QString& toolName);
```

We could document that the name is used since the pointer might be invalid at the time the signal is invoked.

PR #305, opened 2018-11-05
Updated UserCredentials Dialog to deal with QuickControls2

anmacdonald:
With the turning on of High-DPI mode, the user credentials dialog showed heavily inflated text.

This change has removed the scale-factor calculations, and upgraded the dialog to QuickControls 2. 

References to explicit widths/heights have been removed in favour of resizing to contents.


Import/Esri/ArcGISRuntime/Toolkit/Dialogs/UserCredentialsView.qml
```diff
@@ -227,3 +222,5 @@ Rectangle {
     }
 }
 
+
+
```
JamesMBallard: Please remove the empty lines.

anmacdonald:
Looks good on Android and Mac. Happy to test on iOS/Windows!

JamesMBallard:
>Looks good on Android and Mac. Happy to test on iOS/Windows!

@anmacdonald yes, please if you have time.

anmacdonald:
Hi @JamesMBallard, I think these are acceptable! 

## Android

android_usercredentials

## iOS

ios_usercredentials

## MacOS

macos_usercredentials

## Windows

windows_usercredentials

lsmallwood:
I think it would be worth putting a newline in before the URL so that (at least in the general arcgis.com case it is not split over 2 lines). So something like:

```
You need to sign in to access the
resource at:
'https://www.arcgis.com'
```

Also, any reason that we ha v e the URL in single quotes?

@ldanzinger, @JamesMBallard - any thoughts?

JamesMBallard:
Testing is good. Code changes look good but it would be good to see if we can prevent the url from wrapping across lines as @lsmallwood mentioned. If that gets too hairy we can put that into a new issue to work on.

anmacdonald:
If there is a consensus then shall add. Will add as a new ``Text { }`` to the layout rows. I don't trust newline characters in QML rendering based on how I've seen the callouts behave.

anmacdonald:
New look authentication view:

screen shot 2018-11-08 at 14 08 46

PR #310, opened 2018-12-14
Ldanzinger/controls2 upgrade

ldanzinger:
@JamesMBallard please review/merge. I am updating QtQuickControls to version 2, as well as upping all the other imports for 5.12 - many were quite old. I also removed the purple banner image in the dialogs to be a plain rectangle that uses a default color (the blue from the sample viewer). Users can then customize look and feel which is common in the Controls 2 via the Materials conf file, whereas the image was static.

### DIFF
--- a/Import/Esri/ArcGISRuntime/Toolkit/Dialogs/UserCredentialsView.qml
+++ b/Import/Esri/ArcGISRuntime/Toolkit/Dialogs/UserCredentialsView.qml
@@ -113,7 +113,6 @@ Rectangle {
                 Layout.alignment: Qt.AlignHCenter
                 Layout.fillWidth: true
                 Layout.columnSpan: 2
-                //Layout.margins: 1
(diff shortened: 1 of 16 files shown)


Import/Esri/ArcGISRuntime/Toolkit/Dialogs/UserCredentialsView.qml
```diff
@@ -111,22 +113,17 @@ Rectangle {
                 Layout.alignment: Qt.AlignHCenter
                 Layout.fillWidth: true
                 Layout.columnSpan: 2
-                Layout.margins: 1
+                //Layout.margins: 1
```
JamesMBallard: Is this still needed?

ldanzinger: nope. fixed now

PR #311, opened 2018-12-19
We don't need the full include here.

JamesMBallard:
Bringing in <QtQml> also brings in qendian.h which has compile
errors with Qt 5.12.0 using gcc 4.8.x series compilers.


ldanzinger:
🤘

PR #318, opened 2019-02-14
Stackview now more robust.

anmacdonald:
Push/pop logic was confusing. All we want
is a transition. We now use the replace function
to get this effect.

### DIFF
--- a/Import/Esri/ArcGISRuntime/Toolkit/Controls/PopupStackView.qml
+++ b/Import/Esri/ArcGISRuntime/Toolkit/Controls/PopupStackView.qml
@@ -286,35 +286,42 @@ Item {
 
     /*! internal */
     function nextPopup() {
+        if (popupStack.busy)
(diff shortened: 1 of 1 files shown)


JamesMBallard:
@anmacdonald can you give a brief explanation of what was going wrong? I'm still unsure why the old logic suddenly stopped working.

anmacdonald:
@JamesMBallard I'm not entirely sure this ever worked as intended for a number of items > 2.

_popup1_ and _popup2_ are used to generate transition animations when iterating over the selected items.
In the old system, clicking the next button, we would have a transition of - push popup1, push popup2, and then push popup1 again. I don't think a StackView allows for the same item to be pushed to the stack if it is already in the stack. Hence the QML error when iterating through the list on index >= 2.

Now the `StackView` contains at most 1 item, and we do a replace instead to gain the same transition.

PR #322, opened 2019-04-11
James/100 6

JamesMBallard:
@anmacdonald please review.

This is to change the version to 100.6 for v.next.

### DIFF
--- a/Import/Esri/ArcGISRuntime/Toolkit/Controls/CppApi/qmldir
+++ b/Import/Esri/ArcGISRuntime/Toolkit/Controls/CppApi/qmldir
@@ -17,9 +17,9 @@
 module Esri.ArcGISRuntime.Toolkit.Controls.CppApi
 
 CoordinateConversion 100.2 CoordinateConversion1002.qml
-CoordinateConversion 100.6 CoordinateConversion.qml
+CoordinateConversion 100.5 CoordinateConversion.qml
 ArcGISCompass 100.2 ArcGISCompass1002.qml
-ArcGISCompass 100.6 ArcGISCompass.qml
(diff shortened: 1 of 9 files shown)


Import/Esri/ArcGISRuntime/Toolkit/Controls/CppApi/qmldir
```diff
@@ -17,9 +17,9 @@
 module Esri.ArcGISRuntime.Toolkit.Controls.CppApi
 
 CoordinateConversion 100.2 CoordinateConversion1002.qml
-CoordinateConversion 100.5 CoordinateConversion.qml
+CoordinateConversion 100.6 CoordinateConversion.qml
 ArcGISCompass 100.2 ArcGISCompass1002.qml
-ArcGISCompass 100.5 ArcGISCompass.qml
+ArcGISCompass 100.6 ArcGISCompass.qml
```
anmacdonald: Hi James, you made need to undo the changes you've made  in this and the other `qmldir`.

The only version you need to upgrade is `CurrentVersion 100.6 CurrentVersion.qml`. This allows for 100.6 imports - and 100.5 imports will be counted as 100.6 imports thanks to this.

We only up the versions of the actual components when we make a change to them. Otherwise we'll lose the clear stratum between 100.5 and 100.6 when it comes to making edits.

JamesMBallard: Thanks @anmacdonald. I'll make these changes.

JamesMBallard: I've pushed these changes. Thanks for spotting that @anmacdonald.

PR #326, opened 2019-08-19
Ldanzinger/callout fixes

ldanzinger:
@JamesMBallard please review. This adds support for setting the calloutHeight, as well as specifying a Component to display inside of the QML via a Loader. You could put a WebView and display KML Balloon content, WMS identify results, or anything else you want.

customCallout


Import/Esri/ArcGISRuntime/Toolkit/Controls/Callout.qml
```diff
@@ -391,20 +413,50 @@ Item {
...
+                   } else {
+                       return parent.x + margin;
+                   }
+                y: if (adjustedLeaderPosition === Enums.LeaderPosition.Top ||
+                       adjustedLeaderPosition === Enums.LeaderPosition.UpperLeft ||
+                       adjustedLeaderPosition === Enums.LeaderPosition.UpperRight) {
+                       return parent.y + margin + leaderHeight;
+                   } else {
+                       parent.y + margin
+                   }
```
JamesMBallard: These are complex property assignment and I'm guessing they won't work with property binding (not sure about that). 

Should `x` and `y` recalculate any time `adjustedLeaderPosition` changes? I'm unclear on the expectation here.

ldanzinger: Correct, I started by just binding to the x and y, but unfortunately, some of our logic for the canvas size/shape creation seems to be off, so the x & y properties can't be relied upon. Everything looks fine when you use the normal existing workflows, but in reality, if you were to simply anchor a child rectangle to make it fill the parent of the callout, it would display outside the bounds of the shape we draw. I think this is most likely a bug in our canvas code, but I didn't want to go unwinding that to make this improvement.

X & Y now adjust automatically whenever adjustLeaderPosition changes. For example, if in automatic mode where the leader will auto change while panning, the content will move to the correct location seamlessly. It works well, but is more complicated than I would have liked it

JamesMBallard: Is it possible to rewrite this code to make it clearer that it should depend on adjustedLeaderPosition?

```qml
x: 0
y: 0

onAdjustedLeaderPositionChanged: {
    // all that logic combined to assign x and y
}
```

ldanzinger: @JamesMBallard not quite, because the main importance is to bind to the parent's x and y - the adjusted leader is needed just so we can slightly move the content up or down. I reworked it in my most recent commit and I think it is better. Let me know what you think

JamesMBallard: Thanks @ldanzinger, I think that logic is easier to follow now.

JamesMBallard:
Merging to v.next. If testing all looks good, let's get this in to master as well.

PR #328, opened 2019-09-11
Specify C++14

JamesMBallard:
@mfeigl please review.

This ups our C++ level to 14.


JamesMBallard:
@mfeigl copyright dates get set when new files are added to source control, but don't need to be updated for existing files.

PR #329, opened 2019-09-12
Support of the ARCore and ARKit

GuillaumeBelz:
Work in progress. First implementation of a new APIs to provides ARKit and ARCore features with ArcGISRuntime SDK for Qt.


Common/AR/AR.pri
```diff
@@ -0,0 +1,113 @@
...
+    AR_CORE_GRADLE_PATH = $$AR_CORE_PATH/gradle
+
+    HEADERS += \
+        $$AR_CORE_INCLUDE_PATH/arcore_c_api.h
+
+    INCLUDEPATH += $$AR_CORE_INCLUDE_PATH
+    DEPENDPATH += $$AR_CORE_INCLUDE_PATH
+
+
+    LIBS += -L"$$AR_CORE_LIBS_PATH/jni/arm64-v8a"
```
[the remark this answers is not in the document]
GuillaumeBelz: I talk with Koushik and yes, armv7 is supported. But all the tests I did were run on v8. We probably need to test on armv7 before release.

Common/AR/Ar.md
```diff
@@ -0,0 +1,116 @@
...
+
+There are some conflicts between the AR frameworks and the Qt's rendering thread.
+This lines of code enable the non-threaded render loops mode in Qt.
+See SceneView::renderFrame documentation and Qt's documentation
+https://doc.qt.io/qt-5/qtquick-visualcanvas-scenegraph.html#non-threaded-render-loops-basic-and-windows
+for more informations.
+
+```
+#if defined(Q_OS_IOS) && !defined(Q_OS_ANDROID)
+  qputenv("QSG_RENDER_LOOP", "basic");
```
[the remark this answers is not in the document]
GuillaumeBelz: For the moment, I didn't run tests for performances or memory leaks or things like that. I didn't see difference in Android with or without that.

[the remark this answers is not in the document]
GuillaumeBelz: I misunderstood your comment. In `main.cpp`, this option is enabled for both iOS and Android by default. It's a mistake in the doc (I didn't see it).

GuillaumeBelz:
@tdunn @DanCobb1111 Can you review the doc?
@ldanzinger Can you review the APIs and samples?

Notes:
- some functions contains the comment `not implemented` and will be implemented in the next PRs.
- I didn't try to generate the documentation for the moment. I will do that next week, to be sure there is no error.

Plugin/CppApi/source/ArcGISArView.cpp
```diff
@@ -0,0 +1,309 @@
...
+  \since Esri::ArcGISRuntime 100.6
+  \brief Render and tracks the camera.
+ */
+
+using namespace Esri::ArcGISRuntime;
+using namespace Esri::ArcGISRuntime::Toolkit;
+using namespace Esri::ArcGISRuntime::Toolkit::Internal;
+
+/*!
+  \brief A constructor that accepts an optional \a parent.
```
tdunn:
```suggestion
  \brief A constructor that accepts an optional \a parent object.
```

Plugin/CppApi/source/ArcGISArView.cpp
```diff
@@ -0,0 +1,309 @@
...
+ArcGISArView::ArcGISArView(QQuickItem* parent):
+  ArcGISArViewInterface(parent),
+  m_tmcc(new TransformationMatrixCameraController(this)),
+  m_initialTransformation(TransformationMatrix::createIdentityMatrix(this))
+{
+  connect(m_tmcc, &TransformationMatrixCameraController::originCameraChanged, this, &ArcGISArView::originCameraChanged);
+}
+
+/*!
+  \brief A constructor that accepts an optional \a parent.
```
tdunn: @GuillaumeBelz - Do you want to expose this constructor even though two of its parameters are not implemented? If so, what are these parameters for? Thanks.

DanCobb1111: @GuillaumeBelz  Guilluame - This constructor has unimplemented parameters, should this be marked as "internal" for now?

samples/QmlArExample/main.cpp
```diff
@@ -0,0 +1,171 @@
...
+#define kShowNormal                     "normal"
+#define STRINGIZE(x) #x
+#define QUOTE(x) STRINGIZE(x)
+
+//------------------------------------------------------------------------------
+
+int main(int argc, char *argv[])
+{
+  // There are some conflicts between the AR frameworks and the Qt's rendering thread.
+  // This lines of code enable the non-threaded render loops mode in Qt.
```
tdunn:
```suggestion
  // This code enables the non-threaded render loop mode in Qt.
```

samples/QmlArExample/main.cpp
```diff
@@ -0,0 +1,171 @@
...
+  // See SceneView::renderFrame documentation and Qt's documentation
+  // https://doc.qt.io/qt-5/qtquick-visualcanvas-scenegraph.html#non-threaded-render-loops-basic-and-windows
+  // for more informati ons.
```
tdunn:
```suggestion
  // for more information.
```

samples/QmlArExample/main.cpp
```diff
@@ -0,0 +1,171 @@
...
+
+  // Intialize application window
+  QQmlApplicationEngine appEngine;
+  appEngine.addImportPath(QDir(QCoreApplication::applicationDirPath()).filePath("qml"));
+
+  QString arcGISRuntimeImportPath = QUOTE(ARCGIS_RUNTIME_IMPORT_PATH);
+  QString arcGISToolkitImportPath = QUOTE(ARCGIS_TOOLKIT_IMPORT_PATH);
+
+#if defined(LINUX_PLATFORM_REPLACEMENT)
+  // on some linux platforms the string 'linux' is replaced with 1
```
tdunn:
```suggestion
  // On some Linux platforms the string 'linux' is replaced with 1
```

samples/QmlArExample/main.cpp
```diff
@@ -0,0 +1,171 @@
...
+  // fix the replacement paths which were created
```
tdunn:
```suggestion
  // Fix the created replacement paths
```

samples/QmlArExample/main.cpp
```diff
@@ -0,0 +1,171 @@
...
+  QString replaceString = QUOTE(LINUX_PLATFORM_REPLACEMENT);
+  arcGISRuntimeImportPath = arcGISRuntimeImportPath.replace(replaceString, "linux", Qt::CaseSensitive);
+  arcGISToolkitImportPath = arcGISToolkitImportPath.replace(replaceString, "linux", Qt::CaseSensitive);
+#endif
+
+  // Add the Runtime and Extras path
```
tdunn:
```suggestion
  // Add the Runtime and Extras paths
```

samples/CppArExample/main.cpp
```diff
@@ -0,0 +1,132 @@
...
+#define QUOTE(x) STRINGIZE(x)
+
+//------------------------------------------------------------------------------
+
+using namespace Esri::ArcGISRuntime;
+
+int main(int argc, char *argv[])
+{
+  // There are some conflicts between the AR frameworks and the Qt's rendering thread.
+  // This lines of code enable the non-threaded render loops mode in Qt.
```
tdunn:
```suggestion
  // These lines of code enable the non-threaded render loop mode in Qt.
```

samples/CppArExample/CppArExample.cpp
```diff
@@ -0,0 +1,353 @@
...
+void CppArExample::setSceneView(SceneQuickView* sceneView)
+{
+  if (!sceneView || sceneView == m_sceneView)
+    return;
+
+  m_sceneView = sceneView;
+  emit sceneViewChanged();
+}
+
+// properties for debug mode
```
tdunn: @GuillaumeBelz - Should this be?

`// Sets whether point clouds should be visible.`

samples/CppArExample/CppArExample.cpp
```diff
@@ -0,0 +1,353 @@
...
+  m_arcGISArView->setTranslationFactor(1.0);
+
+  changeScene(true);
+}
+
+// Creates a test scene
+// Mode: Full-Scale AR
+void CppArExample::createFullScaleTestScene()
+{
+  // create scene
```
tdunn: General comment: Capitalize the first word of each comment. Or don't. But choose one way to do it. I prefer capitals.

Plugin/QmlApi/source/QmlArcGISArView.cpp
```diff
@@ -0,0 +1,270 @@
...
+{
+  // to implmente
+}
+
+/*!
+  \internal
+
+  As the QML API headers is not available, it's not possible to use the concret types in the QmlArcGISArView
+  API. Then, this API uses QObject* and this function is used to verify than the type of the objects passed
+  in the parameters are correct (using the Qt's meta-object).
```
tdunn:
```
\brief Verify than \a object is of the type expressed in \a className.
\internal

This method exists because the QML API headers are not available, so it's not possible to use the concrete types in the \c QmlArcGISArViewAPI.

File: Plugin/QmlApi/source/QmlArcGISArView.cpp [hunk 14]
```diff
@@ -0,0 +1,270 @@
...
+    m_sceneView->setProperty("cameraController", QVariant::fromValue(m_tmcc));
+
+  emit transformationMatrixCameraControllerChanged();
+}
+
+/*!
+  \brief Sets the initial transformation used to offset the originCamera.
+
+  The initial transformation is based on an AR point determined via existing plane hit detection
+  from `screenPoint`. If an AR point cannot be determined, this method will return `false`.
```
tdunn: A `void` method cannot return anything.

Common/AR/source/Android/ArCoreWrapper.cpp
```diff
@@ -0,0 +1,933 @@
...
+/*!
+  \internal
+ */
+void ArCoreWrapper::setPointCloudColor(const QColor& pointCloudColor)
+{
+  if (pointCloudColor.isValid())
+  {
+    if (!m_arCorePointCloudRenderer)
+    {
+      m_arCorePointCloudRenderer.reset(new ArCorePointCloudRenderer(this));
```
JamesMBallard: Same question here about ownership. I won't comment on all of these. We need to figure out if we're doing QObject parent ownership or std::unique_ptr ownership; we shouldn't do both unless there is a reason to.

Common/AR/source/Android/ArCoreWrapper.cpp
```diff
@@ -0,0 +1,933 @@
...
+}
+
+/*!
+  \internal
+ */
+int ArCoreWrapper::pointCloudSize() const
+{
+  if (m_arCorePointCloudRenderer)
+    return m_arCorePointCloudRenderer->pointCloudSize();
+  return -1;
```
JamesMBallard: Add a line after the if-statement return. If there are more of these, I won't comment on them all; please fix them.

Edit:
>If there are more of these,

There are many more...

Common/AR/source/Android/ArCoreWrapper.cpp
```diff
@@ -0,0 +1,933 @@
...
+{
+  ArTrackableList_acquireItem(m_arSession, m_arPlaneList, index, &m_arTrackable);
+  ArPlane* arPlane = ArAsPlane(m_arTrackable);
+
+  ArTrackingState outTrackingState = AR_TRACKING_STATE_STOPPED;
+  ArTrackable_getTrackingState(m_arSession, m_arTrackable, &outTrackingState);
+
+  ArPlane* subsumePlane = nullptr;
+  ArPlane_acquireSubsumedBy(m_arSession, arPlane, &subsumePlane);
+  if (subsumePlane != nullptr)
```
JamesMBallard: I think in some cases it may not be obvious you're dealing with a pointer. In this case it's obvious.

```suggestion
  if (subsumePlane)
```

Common/AR/source/Android/ArCoreWrapper.cpp
```diff
@@ -0,0 +1,933 @@
...
+  {
+    releasePointCouldData();
+    return;
+  }
+
+  int32_t size = 0;
+  ArPointCloud_getNumberOfPoints(m_arSession, m_arPointCloud, &size);
+  if (size <= 0)
+  {
+    releasePointCouldData();
```
JamesMBallard:
```suggestion
    releasePointCloudData();
```

Common/AR/source/Android/ArCoreWrapper.cpp
```diff
@@ -0,0 +1,933 @@
...
+    return;
+  }
+
+  // gets the positions of the could points
```
JamesMBallard:
```suggestion
  // gets the positions of the cloud points
```

Plugin/QmlApi/source/QmlArcGISArView.cpp
```diff
@@ -0,0 +1,270 @@
...
+  QObject* currentViewpointCamera = m_sceneView->property("currentViewpointCamera").value<QObject*>();
+  if (!currentViewpointCamera)
+    return nullptr;
+
+  // converts the point from the screen space to the 3D space
+  const std::array<double, 7> hitResult = ArcGISArViewInterface::hitTestInternal(x, y);
+
+  // calculates the matrix transformations and creates the QmlPoint location.
+  QVariant location;
+  QMetaObject::invokeMethod(m_qmlWrapper, "screenToLocation", Qt::DirectConnection,
```
JamesMBallard: What object is `screenToLocation` being invoked on? I don't see any such method in the API that takes this many arguments.

Plugin/QmlApi/source/QmlArcGISArView.cpp
```diff
@@ -0,0 +1,270 @@
...
+ */
+void QmlArcGISArView::resetTrackingInternal()
+{
+  // to implmente
+}
+
+/*!
+  \internal
+
+  As the QML API headers is not available, it's not possible to use the concret types in the QmlArcGISArView
```
JamesMBallard:
```suggestion
  As the QML API headers is not available, it's not possible to use the concrete types in the QmlArcGISArView
```

Common/AR/Ar.md
```diff
@@ -0,0 +1,197 @@
...
+
+## ARCore installation
+
+For the details to uses `ARCore`, pleases refere to the [Google's documentation](https://developers.google.com/ar/).
+The details of the installation of the `ARCore` is describes in the
+[documentation](https://developers.google.com/ar/develop/c/enable-arcore).
+
+To create an application with the `ARCore` support:
+
+1. Downloads the sources of the [ArcGIS Runtime API Toolkit: https://github.com/Esri/arcgis-runtime-toolkit-qt.
```
ldanzinger:
```suggestion
1. Downloads the sources of the [ArcGIS Runtime API Toolkit](https://github.com/Esri/arcgis-runtime-toolkit-qt).
```

Plugin/QmlApi/source/QmlArcGISArView.cpp
```diff
@@ -0,0 +1,270 @@
...
+  API. Then, this API uses QObject* and this function is used to verify than the type of the objects passed
+  in the parameters are correct (using the Qt's meta-object).
+ */
+bool QmlArcGISArView::assertClassName(QObject* object, const QString& className) const
+{
+  return object && object->metaObject()->className() == className;
```
JamesMBallard: This is a nice helper 👍

Common/AR/Ar.md
```diff
@@ -0,0 +1,197 @@
...
+The `AR_TOOLKIT_SOURCE_PATH` variable refers to the path of the sources of the toolkit.
+
+2. Downloads and extracts the ARCore SDK for Android v1.10: https://github.com/google-ar/arcore-android-sdk/releases.
+This project contains the C header file (`arcore_c_api.h`) used to build the ARCore wrapper.
```
ldanzinger: So we need to download the SDK from the above link so that we can access `arcore_c_api.h`? This PR contains this header in `3rdparty` already. Is this step still required?

Common/AR/Ar.md
```diff
@@ -0,0 +1,197 @@
...
+The `AR_CORE_HEADER_PATH` variable refers to the path of the ARCore header files.
```
ldanzinger: Please clarify - does this mean we need to define this variable in our `.pro` and set to the `arcore-android-sdk-1.10.0/libraries/include` path?

Common/AR/Ar.md
```diff
@@ -0,0 +1,197 @@
...
+
+3. Downloads the ARCore `aar` file from the Maven repository: https://maven.google.com/com/google/ar/core/1.10.0/core-1.10.0.aar
+This archive contains the native libraries used by the ARCore wrapper. Renames the extention file from `aar` to `zip` and
+extracts it.
+The `AR_CORE_BINARIES_PATH` variable refers to the path of the files of the ARCore native libraries.
```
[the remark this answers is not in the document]
ldanzinger: I have the same question. And if we do need to set them, then what specifically should they be set to? Top level folder?

samples/CppArExample/CppArExample.cpp
```diff
@@ -0,0 +1,353 @@
...
+    SimpleMarkerSceneSymbol* symbol = new SimpleMarkerSceneSymbol(
+          SimpleMarkerSceneSymbolStyle::Sphere, color, 0.1, 0.1, 0.1, SceneSymbolAnchorPosition::Center, graphicsOverlay);
+    graphicsOverlay->graphics()->append(new Graphic(Point(x, y, 0.00001), symbol, graphicsOverlay));
+  };
+
+  for (int i = -30; i <= 30; ++i)
+  {
+    for (int j = -30; j <= 30; ++j)
+    {
+      createSymbol(0.00001 * i, 0.00001 * j, Qt::blue);
```
JamesMBallard: Implicit conversion here. For readability I would suggest this (here and below)

```suggestion
      createSymbol(0.00001 * i, 0.00001 * j, QColor(Qt::blue));
```

This is optional, the code is readable as it is.

samples/CppArExample/Android/Android.pri
```diff
@@ -0,0 +1,37 @@
+#-------------------------------------------------
```
ldanzinger: General comment - this should probably go under the `Examples` folder and not introduce a new `samples` folder. They are for the same purpose

samples/CppArExample/CppArExample.cpp
```diff
@@ -0,0 +1,353 @@
...
+          SimpleMarkerSceneSymbolStyle::Sphere, color, size, size, size, SceneSymbolAnchorPosition::Center, graphicsOverlay);
+    graphicsOverlay->graphics()->append(new Graphic(Point(x, y, 0.0), symbol, graphicsOverlay));
+  };
+
+  for (int i = -5; i <= 10; ++i)
+  {
+    createSymbol(0.0, 0.000001 * i, 0.1, Qt::blue);
+    createSymbol(0.000001 * i, 0.0, 0.1, Qt::red);
+  }
+  createSymbol(0.0, 0.0, 0.11, Qt::green);
```
JamesMBallard: QColor implicit conversion, see previous comment about that.

Common/AR/Ar.md
```diff
@@ -0,0 +1,197 @@
...
+
+```
+include($$AR_TOOLKIT_SOURCE_PATH/ArCppApi.pri) // for C++ API
+```
+or
+```
+include($$AR_TOOLKIT_SOURCE_PATH/ArQmlApi.pri) // for QML API
+```
+
+7. Copy the gradle files from the ARCore SDK for Android to the `Android` dir in our project.
```
ldanzinger: Specifics would help clarify here. It isn't obvious which dir to copy to/from. For example:

> Copy the gradle files from `toolkit/Common/AR/3rdparty/arcore/gradle` to the `Android` dir in your project.

samples/CppArExample/main.cpp
```diff
@@ -0,0 +1,132 @@
...
+  // See SceneView::renderFrame documentation and Qt's documentation
+  // https://doc.qt.io/qt-5/qtquick-visualcanvas-scenegraph.html#non-threaded-render-loops-basic-and-windows
+  // for more informati ons.
```
JamesMBallard:
```suggestion
  // for more information.
```

samples/QmlArExample/qml/scenes/TabletopTestScene.qml
```diff
@@ -0,0 +1,101 @@
...
+            x: 0.000004
+            y: 0.0
+            color: "red"
+        }
+        TabletopTestSceneGraphic {
+            x: 0.000005
+            y: 0.0
+            color: "red"
+        }
+
```
JamesMBallard: This empty line is inconsistent. Is it needed?

samples/QmlArExample/qml/scenes/TabletopTestScene.qml
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    GraphicsOverlay {
+        id: graphicsOverlay
+        TabletopTestSceneGraphic {
+            x: 0.0
+            y: 0.0
+            color: "green"
+        }
+
```
JamesMBallard: his empty line is inconsistent. Is it needed?

Common/AR/Ar.md
```diff
@@ -0,0 +1,197 @@
+
+# Augmented Reality (AR) toolkit for ArcGIS Runtime SDK for Qt
+
+The AR toolkit provides support for ARKit for iOS and ARCore for Android.
+This documentation describes the installation and uses of the AR toolkik.
+
+## Overview
+
+Augmented reality experiences are designed to "augment" the physical world with virtual content that
+respects real world scale, position, and orientation of a device. In the case of Runtime, a SceneView
```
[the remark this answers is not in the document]
GuillaumeBelz: "device" = the smartphone or the tablet

Common/AR/Ar.md
ldanzinger: Are there lines you need to add to your pro/pri files to reference the gradle/manifest files? Nothing mentioned here, but the examples have additions

Common/AR/AR.pri
```diff
@@ -0,0 +1,122 @@
...
+        $$AR_COMMON_SOURCE_PATH/Android/ArCorePointCloudRenderer.cpp \
+        $$AR_COMMON_SOURCE_PATH/Android/ArCorePlaneRenderer.cpp
+
+    INCLUDEPATH += $$AR_COMMON_INCLUDE_PATH/Android
+    DEPENDPATH += $$AR_COMMON_INCLUDE_PATH/Android
+
+    # Extra files from the ARCore framework (Apache 2.0 licence)
+    AR_CORE_PATH = $$AR_COMMON_PATH/3rdparty/arcore
+    AR_CORE_INCLUDE_PATH = $$AR_CORE_PATH/include
+    AR_CORE_LIBS_PATH = $$AR_CORE_PATH/libs
```
ldanzinger: this folder is not here in the repo. Is it expected that the user download the maven files and place in this location? It seems so being that those files are expected to be there in lines 109-121 below. If that is the case, these steps should be mentioned in the Ar.md setup instructions. Right now, they mention to download the files and set the `AR_CORE_LIBS_PATH`, but it seems like really you should download them and copy them inside your 3rdparty folder

samples/QmlArExample/main.cpp
```diff
@@ -0,0 +1,171 @@
...
+  QSettings::setDefaultFormat(kSettingsFormat);
+
+  // Initialize license
+
+#ifdef kLicense
+  QCoreApplication::instance()->setProperty("Esri.ArcGISRuntime.license", kLicense);
+#endif
+
+  // Register the QmlArcGISArView type
+  qmlRegisterType<Esri::ArcGISRuntime::Toolkit::QmlArcGISArView>("Esri.ArcGISArToolkit", 1, 0, "ArcGISArViewInternal");
```
ldanzinger: why is this registering `ArcGISArViewInternal`? Is this correct?

Common/AR/Ar.md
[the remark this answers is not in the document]
GuillaumeBelz: cc @DanCobb1111

samples/QmlArExample/main.cpp
JamesMBallard: It's not correct. I just ran into this problem on iOS when running this example app. I think it needs to be as follows:

```suggestion
  qmlRegisterType<Esri::ArcGISRuntime::Toolkit::QmlArcGISArView>("Esri.ArcGISArToolkit", 1, 0, "ArcGISArView");
```

Common/AR/source/ArKitUsage.cpp
```diff
@@ -0,0 +1,41 @@
...
+    \row
+      \li ArKitUsage::Unknown (0)
+      \li The status of the AR framework is unknown.
+    \row
+      \li ArKitUsage::Yes (1)
+      \li The AR framework is used.
+    \row
+      \li ArKitUsage::No (2)
+      \li The AR framework is not used.
+  \endtable
```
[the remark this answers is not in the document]
tdunn: I suggest replacing this file with `qdoc\ArKitUsage.qdoc`.

Common/AR/Ar.md
```diff
@@ -0,0 +1,197 @@
...
+- **Flyover**: Flyover AR allows you to explore a scene using your device as a window into the virtual world.
+A typical flyover AR scenario will start with the scene’s virtual camera positioned over an area of interest.
+You can walk around and reorient the device to focus on specific content in the scene.
+- **Tabletop**: Scene content is anchored to a physical surface, as if it were a 3D-printed model.
+- **Real-scale**: Scene content is rendered exactly where it would be in the physical world. A camera feed
+is shown and GIS content is rendered on top of that feed. This is used in scenarios ranging from viewing
+hidden infrastructure to displaying waypoints for navigation.
+
+The AR toolkit component is comprised of one class: `ArcGISArView`. This is a subclass of `QQuickItem` that
+contains the functionality needed to display an AR experience in your application. It uses `ARKit` (Apple) or
```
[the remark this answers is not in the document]
GuillaumeBelz: @JamesMBallard  Good point. I used `Android` and `Apple`, but it makes sense to use `Android` and `iOS` or `Google` and `Apple`. What do you prefer?

Common/AR/Ar.md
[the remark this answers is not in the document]
GuillaumeBelz: This part of documentation was removed.

Common/AR/Ar.md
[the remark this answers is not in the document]
JamesMBallard: I prefer `(iOS)` for consistency so we're talking about the underlying OS.

Common/AR/Ar.md
```diff
@@ -0,0 +1,197 @@
...
+```
+include($$AR_TOOLKIT_SOURCE_PATH/ArQmlApi.pri) // for QML API
+```
+
+7. Copy the gradle files from the ARCore SDK for Android to the `Android` dir in our project.
+
+8. In the Android manifest file `AndroidManifest.xml`, adds the following entries:
+
+```
+<meta-data android:name="com.google.ar.core" android:value="optional" />
```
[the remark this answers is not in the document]
GuillaumeBelz: Good point. The `ArcGISArView` should work in the future without ARKit/ARCore (https://devtopia.esri.com/runtime/queen-annes-revenge/issues/154), using the sensors of the device. For the moment, this is not implemented.

With `required`, the app is visible in the app stores only for the devices which support AR. With `optional`, the app is visible for all the devices.

If the user want to create an app which contains an AR scene and a "normal" map/scene, then this parameter must be `optional`. If not (this is the case of the samples), `required` is fine.

I suppose I can improve the doc to explains that.

Common/AR/include/Android/ArCoreFrameRenderer.h
```diff
@@ -0,0 +1,54 @@
...
+#ifndef ArCoreFrameRenderer_H
+#define ArCoreFrameRenderer_H
+
+#include <QOpenGLFunctions>
+#include <QOpenGLShaderProgram>
+
+namespace Esri {
+namespace ArcGISRuntime {
+namespace Toolkit {
+namespace Internal {
```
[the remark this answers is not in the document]
GuillaumeBelz: Yes. Except if I did a mistake, the public classes are `ArcGISArView`, `ArcGISArViewInterface` and `LocationDataSource`.

Common/AR/include/Android/ArCoreWrapper.h
```diff
@@ -0,0 +1,164 @@
...
+  void render();
+
+  void udpateArCamera();
+  void releaseArData();
+
+  bool renderVideoFeed() const;
+  void setRenderVideoFeed(bool renderVideoFeed);
+
+  // hit test for screen to location feature
+  std::array<double, 7> hitTest(int x, int y) const;
```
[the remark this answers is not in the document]
GuillaumeBelz: Yes. And this name is used by ARCore https://developers.google.com/ar/reference/java/arcore/reference/com/google/ar/core/Frame#hitTest(android.view.MotionEvent and ARKit https://developer.apple.com/documentation/arkit/arframe/2875718-hittest?language=objc

Perhaps I can add these links to the docs in the comments?

Common/AR/Ar.md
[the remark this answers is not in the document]
JamesMBallard: Ok, that's good to know about the app store restrictions.

I think for now, we should advise people of the limitations as you've mentioned (regarding app store restrictions), but this doc should show/recommend `android:value="required"` since it will not function without that.

Common/AR/include/Android/ArCoreWrapper.h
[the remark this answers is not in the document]
JamesMBallard: Nothing more needed, you're both right; this isn't too esoteric. It's good like this.

Common/AR/include/Android/ArCoreWrapper.h
```diff
@@ -0,0 +1,164 @@
...
+  std::unique_ptr<ArCorePlaneRenderer> m_arCorePlaneRenderer;
+  std::unique_ptr<ArCorePointCloudRenderer> m_arCorePointCloudRenderer;
+
+  void renderArFrame();
+  void renderArPlane();
+  void renderArPointCloud();
+
+  // When the apllication launches or enters an AR mode,
+  // it should call this method with user_requested_install = 1.
+  static int32_t m_installRequested;
```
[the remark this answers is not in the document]
GuillaumeBelz: This parameter is used by `ArCoreApk_requestInstall` and can take only the values 0 or 1. `ARCore` is a C API, there is no `bool` type.

For `atomic`, I guess it's not necessary, this function is not called (I guess) in different threads. This function is used when the app starts for the first time and the installation of the `ARCore` is required. In this case, the app is paused and the Google Play is started to install `ARCore`. When the installation is finished or canceled, the app become active and the `ArCoreApk_requestInstall` is called again. There is no concurrent access to `m_installRequested`.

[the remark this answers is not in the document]
JamesMBallard: That makes sense, no changes needed here.

Common/AR/include/ArWrapper.h
```diff
@@ -0,0 +1,90 @@
...
+class ArWrapper : public ArKitWrapper { using ArKitWrapper::ArKitWrapper; };
+
+#elif defined Q_OS_ANDROID
+
+// implementation of the AR core for Android.
+class ArWrapper : public ArCoreWrapper { using ArCoreWrapper::ArCoreWrapper; };
+
+#else
+
+// default implementation for desktop platforms.
```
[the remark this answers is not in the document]
GuillaumeBelz: I think it's useful to build and run the app in desktop, to see the scene and help the creation of the app. The default `ArWrapper` do nothing, then it's easy to maintains it.

And if the user want to create an optional AR view and what to run it on Desktop. That means he need to use `#ifdef platform` in the code.

I guess it's better to have an `ArWrapper` which do nothing on Desktop.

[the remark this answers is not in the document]
JamesMBallard: That's a good point, let's keep it as-is.

Common/AR/include/iOS/ArKitWrapper.h
```diff
@@ -0,0 +1,95 @@
...
+  // hit test for screen to location feature
+  std::array<double, 7> hitTest(int x, int y) const;
+
+  // low access to the ARKit objects
+  template<typename ArRawPtr>
+  ArRawPtr* arRawPtr() const;
+
+  // point cloud data
+  QMatrix4x4 viewProjectionMatrix() const;
+  std::vector<float> pointCloudData() const;
```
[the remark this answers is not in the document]
GuillaumeBelz: Not really better. It's just because it's an internal class and the first implementation used `const float*`. I used `std::vector` by default.

Common/AR/source/Android/ArCoreWrapper.cpp
```diff
@@ -0,0 +1,933 @@
...
+
+// C++ headers
+#include <array>
+
+/*!
+  \class Esri::ArcGISRuntime::Toolkit::ArcGISARView
+  \ingroup AR
+  \inmodule ArcGISQtToolkit
+  \since Esri::ArcGISRuntime 100.6
+  \brief Wrap the ARCore for Android.
```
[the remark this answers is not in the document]
GuillaumeBelz: Yes, this class is internal. I remove that.

Common/AR/source/Android/ArCoreWrapper.cpp
```diff
@@ -0,0 +1,933 @@
...
+  auto status = ArSession_create(jniEnvironment(), applicationActivity(), &m_arSession);
+  if (status != AR_SUCCESS || !m_arSession)
+  {
+    emit m_arcGISArView->errorOccurred("ARCore failure", "Fails to create the AR session.");
+    return;
+  }
+
+  // request camera permission
+  const QString permissionKey = QStringLiteral("android.permission.CAMERA");
+  auto permissions = QtAndroid::requestPermissionsSync({ permissionKey }, 5000);
```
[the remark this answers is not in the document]
GuillaumeBelz: I need to verify if the permission is just reading the `AndroidManifest.xml` file or if this is a request to the user.

Common/AR/source/Android/ArCoreWrapper.cpp
```diff
@@ -0,0 +1,933 @@
...
+void ArCoreWrapper::setRenderVideoFeed(bool renderVideoFeed)
+{
+  m_renderVideoFeed = renderVideoFeed;
+
+  // create ArCoreFrameRenderer if necessary
+  if (renderVideoFeed)
+  {
+    if (!m_arCoreFrameRenderer)
+    {
+      m_arCoreFrameRenderer.reset(new ArCoreFrameRenderer(this));
```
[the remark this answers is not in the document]
GuillaumeBelz: No, `ArCoreFrameRenderer` is not a `QObject`. There is a argument `this` because `ArCoreFrameRenderer` must request the frame and texture parameters to the wrapper.

Common/AR/source/Android/ArCoreWrapper.cpp
GuillaumeBelz: Same. All the `XxxRenderer` classes are not `QObject`, they need to access to the wrapper to get the data.

It's a little bit tricky, because on Android, the renderer is called by the wrapper (and then the data are sent when the function is called) and it's not necessary to keep a pointer to the wrapper in the renderer. And on iOS, the renderer is called by the `QQuickItem` and the renderer calls the wrapper functions (and then the wrapper need to keep the data until the renderer use it).
I think it's ok for the concurrent access. In `ArKitWrapper`, the property `textureDataUsed` is used to avoid to delete the frame texture if the `ArKitFrameRenderer` still use the texture. This is the reason I'm using `std::vector` in parameter (to copy the data) rather than using `float*`.

Common/AR/Ar.md
```diff
@@ -0,0 +1,197 @@
...
+```
+<uses-permission android:name="android.permission.CAMERA"/>
+```
+
+9. The project is ready to be build and run in an Android device. You need to verify the compatibility of your device
+with the ARKit.
+
+## Usage
+
+Using the C++ API:
```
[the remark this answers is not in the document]
GuillaumeBelz: I updated the comment but without `flyover`. The `flyover` mode depends of the content of the scene and the camera parameters. Both are not in the code example. Should we add that? Or samples are enough?

Common/AR/Ar.md
```diff
@@ -0,0 +1,197 @@
...
+    id: arcGISArView
+    anchors.fill: parent
+    sceneView: sceneView
+    locationDataSource: LocationDataSource {}
+}
+
+SceneView {
+    id: sceneView
+    anchors.fill: parent
+    scene: sceneLoader.item ? sceneLoader.item.scene : null
```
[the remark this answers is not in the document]
GuillaumeBelz: Sorry, copy and paste from the sample.

Common/AR/source/LocationDataSource.cpp
```diff
@@ -0,0 +1,232 @@
...
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ ******************************************************************************/
+
+#include "LocationDataSource.h"
+#include <QGeoPositionInfoSource>
+#include <QCompass>
+
+/*!
+  \class Esri::ArcGISRuntime::Toolkit::LocationDataSource
```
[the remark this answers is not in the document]
GuillaumeBelz: The issue is the `LocationDataSource` class is not public in the Qt's Runtime API. And the AR's `LocationDataSource` don't need to be shared with the core runtime. For the AR point of view, it's just a wrapper to `QGeoPositionInfoSource` and `QCompass`, it's totally independent of the core runtime.

I think it's probably better to change the name of this class rather than using `QRTLocationDataSource` in this class.

Common/AR/Ar.md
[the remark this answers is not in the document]
ldanzinger: This is good, thanks

Common/AR/source/LocationDataSource.cpp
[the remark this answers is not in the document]
ldanzinger: Are other Runtime APIs using the LocationDataSource from core or are they implementing a new toolkit version like what you did?

samples/CppArExample/CppArExample.pro
```diff
@@ -0,0 +1,75 @@
...
+    qml/*.qml \
+    qml/*.js
+
+#-------------------------------------------------------------------------------
+# AR configuration
+
+# The path to the ArcGIS runtime toolkit for Qt sources, corresponding to the files downloaded
+# from the GitHub repo: https://github.com/Esri/arcgis-runtime-toolkit-qt
+
+AR_TOOLKIT_SOURCE_PATH = # must be set to the path to toolkit sources
```
ldanzinger: you could consider adding a qmake message for if this is empty so it is obvious to a user that opens the example that they need to set this. Otherwise, it is a bit buried and might be easy to miss

samples/CppArExample/qml/main.qml
```diff
@@ -0,0 +1,66 @@
...
+        anchors {
+            top: parent.top
+            left: parent.left
+            margins: 5
+        }
+
+        onStartTrackingClicked: arcGISArView.startTracking();
+        onStopTrackingClicked: arcGISArView.stopTracking();
+        onResetTrackingClicked: arcGISArView.resetTracking();
+        // onCalibrationClicked: not implemented
```
ldanzinger: should this be removed?

samples/QmlArExample/QmlArExample.pro
```diff
@@ -0,0 +1,85 @@
...
+    qml/scenes/*.qml \
+    qml/*.js
+
+#-------------------------------------------------------------------------------
+# AR configuration
+
+# The path to the ArcGIS runtime toolkit for Qt sources, corresponding to the files downloaded
+# from the GitHub repo: https://github.com/Esri/arcgis-runtime-toolkit-qt
+
+AR_TOOLKIT_SOURCE_PATH = # must be set to the path to toolkit sources
```
ldanzinger: same comment here as with c++ .pro

Common/AR/Ar.md
```diff
@@ -0,0 +1,197 @@
...
+## ARKit installation
+
+For the details to uses `ARKit`, pleases refere to the [Apple's documentation](https://developer.apple.com/augmented-reality).
+
+To create an application with the `ARKit` support:
+
+1. Downloads the sources of the [ArcGIS Runtime API Toolkit](https://github.com/Esri/arcgis-runtime-toolkit-qt).
+Sets the `AR_TOOLKIT_SOURCE_PATH` variable to the path of the sources of the toolkit.
+
+2. Installs the ArcGIS Runtime SDK for Qt version 100.6
```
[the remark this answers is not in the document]
GuillaumeBelz: I removed the version and add a new line in the end of overview:
```
The minimal version of the ArcGIS Runtime SDK for Qt is 100.6.
```

Common/AR/source/LocationDataSource.cpp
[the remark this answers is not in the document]
GuillaumeBelz: They use the LocationDataSource from core but this class is public in other SDKs, not in Qt.

Plugin/CppApi/include/ArcGISArView.h
```diff
@@ -0,0 +1,93 @@
...
+
+  // methods
+  Q_INVOKABLE void setInitialTransformation(const QPoint& screenPoint);
+  Q_INVOKABLE Point screenToLocation(const QPoint& screenPoint) const;
+
+signals:
+  void originCameraChanged();
+  void sceneViewChanged();
+
+public: // internals, used by AR wrappers
```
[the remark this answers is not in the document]
GuillaumeBelz: I would have liked to use `private` and `friend` but the there are 2 classes which need to access to that: `ArCoreWrapper` and `ArKitWrapper`, then this must be platform dependant. Maybe it's easier to use `public`?

Plugin/CppApi/source/ArcGISArView.cpp
```diff
@@ -0,0 +1,309 @@
...
+
+  \list
+    \li \a screenPoint - The screen point to determine the `initialTransformation` from.
+  \endlist
+ */
+void ArcGISArView::setInitialTransformation(const QPoint& screenPoint)
+{
+  // Use the `hitTestInternal` method to get the matrix of `screenPoint`.
+  const std::array<double, 7> hitResult = hitTestInternal(screenPoint.x(), screenPoint.y());
+  if (hitResult[3] == 0) // quaternionW shouldn't be 0
```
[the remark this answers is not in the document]
GuillaumeBelz: `quaternionW` parameter can't be 0, except when the array is default initialized (ie when the function return with error).

The best should be something like `std::optional<std::array>` or une structure with a `bool` and `array`. But it's simpler to just verify the `quaternionW` parameter.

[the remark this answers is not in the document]
JamesMBallard: That makes sense. I suggest a simple comment then indicating that, something like this perhaps?

```c++
// quaternionW can never be 0, this indicates an error occurred
if (hitResult[3] == 0) 
```

Plugin/QmlApi/include/QmlArcGISArView.h
```diff
@@ -0,0 +1,98 @@
...
+  void initialTransformationChanged(double quaternionX, double quaternionY, double quaternionZ, double quaternionW,
+                                    double translationX, double translationY, double translationZ);
+  void transformationMatrixChanged(double quaternionX, double quaternionY, double quaternionZ, double quaternionW,
+                                   double translationX, double translationY, double translationZ);
+  void fieldOfViewChanged(double xFocalLength, double yFocalLength,
+                          double xPrincipal, double yPrincipal,
+                          double xImageSize, double yImageSize);
+  void resetTracking();
+
+public: // internals, used by AR wrappers
```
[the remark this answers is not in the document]
GuillaumeBelz: Same as `ArcGISArView.h`

Plugin/QmlApi/source/QmlArcGISArView.cpp
GuillaumeBelz: This can be used by the users to place an 3D object in the 3D scene. For the moment, the sample for that is not implemented, it's in the list for the next PR.

samples/CppArExample/Android/Android.pri
GuillaumeBelz: I will move the examples in the correct directory in the next PR, to simplify the following of the comments.

samples/QmlArExample/qml/scenes/TabletopTestScene.qml
GuillaumeBelz: I added comments to explain why there is blue and red spheres

Common/AR/Ar.md
```diff
@@ -0,0 +1,197 @@
...
+include($$AR_TOOLKIT_SOURCE_PATH/ArCppApi.pri) // for C++ API
+```
+or
+```
+include($$AR_TOOLKIT_SOURCE_PATH/ArQmlApi.pri) // for QML API
+```
+
+7. Copy the gradle files from the ARCore SDK for Android to the `Android` dir in our project.
+
+8. In the Android manifest file `AndroidManifest.xml`, adds the following entries:
```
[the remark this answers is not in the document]
GuillaumeBelz: Right, I created it using QtCreator. I will update the doc

samples/CppArExample/qml/main.qml
GuillaumeBelz: This will be added in the next PR

Common/AR/Ar.md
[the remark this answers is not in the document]
GuillaumeBelz: I changed the value to `required` and add the following details:

```
The value can be `optional` or `required`. An AR optional app can be installed and run on devices
that don't support `ARCore` and the `Google Play Services for AR` is not automatically installed
with the app. An AR required app is available only on devices that support ARCore and the
`Google Play Services for AR` is automatically installed.

For details, see [Enable ARCore documentation](https://developers.google.com/ar/develop/java/enable-arcore#manifest).
```

Plugin/CppApi/include/ArcGISArView.h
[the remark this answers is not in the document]
JamesMBallard: Yes, keep it as public.

Common/AR/source/ArWrapper.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+// implemented in ArCoreWrapper
+
+#else
+
+// default implementation for desktop platforms.
+
+ArWrapper::ArWrapper(void*)
+{
+}
```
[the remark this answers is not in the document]
GuillaumeBelz: For an AR optional app (see my comment about `optional` vs `required` in `AndroidManifest.xml`), I guess it can be interesting to build and run on Desktop. And this can be helpful to debug the scene.

For the implementation of `ArcGISArView`, it's easier to have an `ArWrapper` with the same API than `ArCoreWrapper` and `ArKitWrapper`.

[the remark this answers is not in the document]
JamesMBallard: That makes sense to be able to prototype and run on Desktop. Let's leave this as-is.

samples/QmlArExample/main.cpp
GuillaumeBelz: There is a little confusion here.

There is an C++ class called `QmlArcGISArView`. This class is registred to QML engine with the name `ArcGISArViewInternal`. And the QML component `ArcGISArView` is implemented in the file `ArcGISArView.qml` and use`ArcGISArViewInternal` internally. The users can't use directly the C++ class.

Maybe the names are not the best one? Maybe `QmlArcGISArView` should be renamed to `ArcGISArViewInternal`?

samples/CppArExample/Android/Android.pri
GuillaumeBelz: @ldanzinger the `Examples` folder is not a folder with a list of examples but it contains one project called `Examples`. If the AR examples are moved to this folder, I suppose the `Examples` project should be rename and moved to a subdir. Which name do you want to use?

ldanzinger: @GuillaumeBelz I'd say leave the example project as-is, and then add a sub dir called "AR" and put your AR samples in there

tdunn:
@GuillaumeBelz @JamesMBallard - I have a working API doc build now. The output needs some tweaking, but it's good enough for the moment. I'm continuing to work on it. I can make additional changes in another branch if you are otherwise ready to merge this PR.

samples/QmlArExample/main.cpp
JamesMBallard: Ideally users shouldn't have to register any class with internal in the name. Is that possible?

GuillaumeBelz: I added `QmlArcGISArView::qmlRegisterTypes()` to do that. In the future, If we provide the binaries of the AR toolkit, this can be replaced by `qmldir` and `.qmlfiles` in a QML plugin.

Common/AR/source/Android/ArCoreWrapper.cpp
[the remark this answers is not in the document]
GuillaumeBelz: This is a request to the user and it seems ok to block the app until the user send the permission.

Common/AR/Ar.md
GuillaumeBelz: To be sure the explanations are correct, I implemented the "Redlands Fire Hydrants" samples in new projects from scratch and notes all the steps.

GuillaumeBelz:
@ldanzinger I updated the explanations to install and configure AR in `AR.md`. Can you review to check if this is better? Thank you.

Common/AR/Ar.md
```diff
@@ -0,0 +1,197 @@
...
+For the details to uses `ARCore`, pleases refere to the [Google's documentation](https://developers.google.com/ar/).
+The details of the installation of the `ARCore` is describes in the
+[documentation](https://developers.google.com/ar/develop/c/enable-arcore).
+
+To create an application with the `ARCore` support:
+
+1. Downloads the sources of the [ArcGIS Runtime API Toolkit: https://github.com/Esri/arcgis-runtime-toolkit-qt.
+The `AR_TOOLKIT_SOURCE_PATH` variable refers to the path of the sources of the toolkit.
+
+2. Downloads and extracts the ARCore SDK for Android v1.10: https://github.com/google-ar/arcore-android-sdk/releases.
```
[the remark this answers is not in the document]
GuillaumeBelz: For the moment, as we provide the ARCore libs, we need to update them ourselves. But we should to remove these files and update the gradle scripts to download and extract automatically these files.

Common/AR/Ar.md
```diff
@@ -0,0 +1,197 @@
...
+arcGISArView->setLocationDataSource(new LocationDataSource(this));
+
+// Start tracking our location and device orientation
+arcGISArView->startTracking();
+```
+
+Using the QML API:
+
+```
+// create the AR view
```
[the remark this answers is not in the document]
GuillaumeBelz: Documentation is updated.

Common/AR/Ar.md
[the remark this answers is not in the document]
GuillaumeBelz: Done in the last version of the documentation.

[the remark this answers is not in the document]
GuillaumeBelz: Done.

Common/AR/Ar.md
```diff
@@ -0,0 +1,340 @@
...
+1. Open the `Info.plist` file and add the following key at the end of the file, just before `</dict>`,
+to give the access to the camera to the `ARKit`.
+```
+<key>NSCameraUsageDescription</key>
+<string>Camera access is needed for AR testing</string>
+```
+
+See [NSCameraUsageDescription Apple's documentation](https://developer.apple.com/documentation/bundleresources/information_property_list/nscamerausagedescription?language=objc) for details.
+
+2. The project is ready to be build and run in an iOS device. You need to verify the compatibility of your device
```
ldanzinger: Is there a link to any doc that explains which devices are compatible for iOS?

Common/AR/Ar.md
```diff
@@ -0,0 +1,340 @@
...
+
+3. Open the `build.gradle` file. In `dependencies`, add the ARCore dependency:
+```
+dependencies {
+    implementation fileTree(dir: 'libs', include: ['*.jar', '*.aar'])
+    implementation 'com.google.ar:core:1.11.0'
+}
+```
+
+4. The project is ready to be built and run on an Android device. You need to verify the compatibility of your device
```
ldanzinger: Is there a link to any doc that explains which devices are compatible for Android?

Common/AR/Ar.md
```diff
@@ -0,0 +1,340 @@
...
+with the ARKit.
```
ldanzinger:
```suggestion
with ARCore.
```

Common/AR/Ar.md
```diff
@@ -0,0 +1,340 @@
...
+
+## Note for performance issues
+
+There are some conflicts between the AR frameworks and the Qt's rendering thread.
```
ldanzinger:
```suggestion
There are some conflicts between the AR frameworks and Qt's rendering thread.
```

Common/AR/Ar.md
```diff
@@ -0,0 +1,340 @@
...
+## Creating an new AR application using ArcGIS Runtime AR toolkit
+
+### Creating a new C++ application with AR support
+
+An example of C++ application with AR support can be found in the folder `Examples\AR\CppArExemple` in
+the toolkit repo.
+
+In the following explaination, `<ProjectName>` refere to the project name used to create the Qt's project.
+
+1. Install the ArcGIS Runtime SDK for Qt.
```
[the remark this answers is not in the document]
GuillaumeBelz:
```
1. Install the ArcGIS Runtime SDK for Qt. See
[ArcGIS Runtime SDK for Qt](https://developers.arcgis.com/qt/latest/qml/guide/arcgis-runtime-sdk-for-qt.htm)
for details.
```

Common/AR/Ar.md
GuillaumeBelz:
```

2. The project is ready to be build and run in a compatible iOS device. AR requires an iOS or iPadOS device
with iOS 11 and an A9 processor or later. The list of compatible devices can be found in the end of the
[Augmented Reality page](https://www.apple.com/lae/ios/augmented-reality/)
```

Common/AR/Ar.md
GuillaumeBelz:
```

4. The project is ready to be build and run in a compatible Android device. ARCore requires Android 7.0 or later.
The list of compatible devices can be found in
[ARCore supported devices](https://developers.google.com/ar/discover/supported-devices)
```

JamesMBallard: Should be built not build.

Plugin/QmlApi/source/QmlArcGISArView.cpp
```diff
@@ -0,0 +1,269 @@
...
+void QmlArcGISArView::resetTrackingInternal()
+{
+  emit resetTrackingChanged();
+}
+
+/*!
+  \brief Verify than \a object is of the type expressed in \a className.
+  \internal
+
+  This method exists because the QML API headers are not available, so it's not possible to use the concret
```
JamesMBallard:
```suggestion
  This method exists because the QML API headers are not available, so it's not possible to use the concrete
```

PR #333, opened 2019-10-10
Ar screen to location

GuillaumeBelz:
Add demo for screenToLocation function. WIP.


GuillaumeBelz:
There still a bug in the matrix transformation and I continue to work on that but I guess the code can be reviewed. This PR will not be merged until the bug will be fixed.

The function is already in the APIs then there is not a lot of doc to review. This PR is to add example for this function and fix the behavior.

cc @JamesMBallard @tdunn @ldanzinger

samples/CppArExample/CppArExample.cpp
```diff
@@ -310,6 +317,45 @@ void CppArExample::createTabletopTestScene()
...
+    if (!graphicsOverlay)
+    {
+      graphicsOverlay = new GraphicsOverlay(this);
+      m_sceneView->graphicsOverlays()->append(graphicsOverlay);
+    }
+    Q_CHECK_PTR(graphicsOverlay);
+
+    // Create and place a graphic at the real world location.
+    SimpleMarkerSceneSymbol* sphere = new SimpleMarkerSceneSymbol(
+          SimpleMarkerSceneSymbolStyle::Sphere, QColor("yellow"), 0.25, 0.25, 0.25,
```
JamesMBallard: Prefer QColor enum constructor when possible.

```suggestion
          SimpleMarkerSceneSymbolStyle::Sphere, QColor(Qt::yellow), 0.25, 0.25, 0.25,
```

samples/CppArExample/CppArExample.h
```diff
@@ -34,6 +34,7 @@ class CppArExample : public QObject
   Q_PROPERTY(Esri::ArcGISRuntime::Toolkit::ArcGISArView* arcGISArView READ arcGISArView WRITE setArcGISArView
              NOTIFY arcGISArViewChanged)
   Q_PROPERTY(Esri::ArcGISRuntime::SceneQuickView* sceneView READ sceneView WRITE setSceneView NOTIFY sceneViewChanged)
+  Q_PROPERTY(bool screenToLocationMode MEMBER m_screenToLocationMode NOTIFY screenToLocationModeChanged)
```
JamesMBallard: When using `MEMBER` and you change `m_screenToLocationMode` in your C++ code, you must manually emit `screenToLocationModeChanged()`.

I don't see that happening anywhere, so not sure if that's needed or not.

samples/CppArExample/CppArExample.cpp
```diff
@@ -310,6 +317,45 @@ void CppArExample::createTabletopTestScene()
   changeScene();
 }
 
+// If m_screenToLocationMode is true, create and place a 3D sphere in the scene. Otherwise set the
+// initial transformation based on the screen position.
+void CppArExample::onTouched(QTouchEvent& event)
+{
+  Q_CHECK_PTR(m_arcGISArView);
+
+  const QPoint screenPoint = event.touchPoints().first().lastScreenPos().toPoint();
```
JamesMBallard: This looks dangerous since it assumes `first()` returns a valid object, but I think you're safe since `onTouched` would only be invoked for an actual touch event.

Do you think there's any need to validate this for safety? I think it's ok like this.

samples/QmlArExample/qml/main.qml
```diff
@@ -39,7 +41,44 @@ ApplicationWindow {
         id: sceneView
         anchors.fill: parent
         scene: sceneLoader.item ? sceneLoader.item.scene : null
-        onMousePressed: arcGISArView.setInitialTransformation(mouse.x, mouse.y); // for touch screen events
+
+        // If m_screenToLocationMode is true, create and place a 3D sphere in the scene. Otherwise set the
```
JamesMBallard:
```suggestion
        // If screenToLocationMode is true, create and place a 3D sphere in the scene. Otherwise set the
```

samples/QmlArExample/qml/main.qml
```diff
@@ -39,7 +41,44 @@ ApplicationWindow {
...
+            if (screenToLocationMode)
+            {
+                // Get the real world location for screen point from AR view.
+                const point = arcGISArView.screenToLocation(mouse.x, mouse.y);
+                if (!point)
+                    return;
+
+                // Get or create graphic overlay
+                var graphicsOverlay = graphicsOverlays.get(0);
+                if (!graphicsOverlay)
```
JamesMBallard: I think this code would be cleaner if you check the graphicsOverlays size/count first before calling `.get(0)`.

This same comment applies to the C++ version.

samples/QmlArExample/qml/main.qml
```diff
@@ -39,7 +41,44 @@ ApplicationWindow {
...
         anchors.fill: parent
         scene: sceneLoader.item ? sceneLoader.item.scene : null
-        onMousePressed: arcGISArView.setInitialTransformation(mouse.x, mouse.y); // for touch screen events
+
+        // If m_screenToLocationMode is true, create and place a 3D sphere in the scene. Otherwise set the
+        // initial transformation based on the screen position.
+        onMousePressed: {
+            // If "screenToLocation" mode is enabled.
+            if (screenToLocationMode)
+            {
```
JamesMBallard: Brace on previous line for consistent Qml style. This comment applies to many places in this file.

samples/QmlArExample/qml/main.qml
```diff
@@ -39,7 +41,44 @@ ApplicationWindow {
...
+                const sphereParams = {
+                    style: Enums.SimpleMarkerSceneSymbolStyleSphere, color: "yellow",
+                    width: 0.2520, height: 0.25, depth: 0.25,
+                    anchorPosition: Enums.SceneSymbolAnchorPositionBottom };
+                const sphere = ArcGISRuntimeEnvironment.createObject("SimpleMarkerSceneSymbol", sphereParams);
+
+                const sphereGraphicParams = { geometry: point, symbol: sphere };
+                const sphereGraphic = ArcGISRuntimeEnvironment.createObject("Graphic", sphereGraphicParams);
+
+                graphicsOverlay.graphics.append(sphereGraphic);
```
JamesMBallard: I like the usage of the params objects in this block. That is a good idea to help clarify the `ArcGISRuntimeEnvironment.createObject` calls.

samples/CppArExample/CppArExample.h
GuillaumeBelz: `m_screenToLocationMode` is not changed in C++, this is the reason I used `MEMBER` this time. But we don't use `MEMBER` usually, then I can change to setter/getter if you prefer.

JamesMBallard: No, this is fine. In this case `MEMBER` is a lot easier and less code.

samples/CppArExample/CppArExample.cpp
GuillaumeBelz: Doc says it's possible to "cancel" a touch point. I don't know how to do that on mobile device. But you're right, it's safer to check if the list is not empty.

GuillaumeBelz:
@JamesMBallard This PR is ready for the review!

samples/CppArExample/CppArExample.cpp
```diff
@@ -361,8 +400,28 @@ void CppArExample::createSurfaceWithElevation()
   m_scene->setBaseSurface(baseSurface);
 }
 
+// Get or create graphic overlay
+Esri::ArcGISRuntime::GraphicsOverlay* CppArExample::getOrCreateGraphicsOverlay() const
```
JamesMBallard: You don't need the full namespace here.
```suggestion
GraphicsOverlay* CppArExample::getOrCreateGraphicsOverlay() const
```

GuillaumeBelz:
If it's fine for everyone, I will merge this PR.

PR #336, opened 2019-10-21
AR toolkit doc updates

tdunn:
Remaining Toolkit AR API doc updates.


Common/AR/source/LocationDataSource.cpp
```diff
@@ -336,8 +337,8 @@ void LocationDataSource::updateObjectsAndConnections()
  */
 
 /*!
-  \fn void LocationDataSource::geoPositionSourceChangedChanged();
```
GuillaumeBelz: good catch 👍

PR #367, opened 2020-05-04
100.8 final

JamesMBallard:


JamesMBallard:
@GuillaumeBelz please review. This is merging v.next to master once 100.8 is finalized.

As we discussed, there were some merges to master that were not migrated to v.next which caused a minor merge conflict. I've addressed that and we'll be in sync again once this is merged.

Examples/AR/CppArExample/qml/Message.qml
```diff
@@ -1,6 +1,6 @@
 import QtQuick 2.6
 import QtQuick.Controls 2.2
-import Esri.ArcGISRuntime 100.6
+import Esri.ArcGISRuntime 100.8
```
GuillaumeBelz, line 3: @JamesMBallard I suppose you're using a script to find the occurence of "100.7". Do you think it can be interesting to verify if there is no old version like this one? Maybe using a regex to find the string "100.x but not 100.8".

JamesMBallard: @GuillaumeBelz in this case, this is just merging our 100.8 work in v.next to master.  Your comment makes sense for https://github.com/Esri/arcgis-runtime-toolkit-qt/pull/366, which is a regex just as you described.

PR #369, opened 2020-05-14
Update wildcarding for Qt 5.15.0

ecmorris:
Toolkit update for 5.15.0 wildcarding, only uses it in the one file.


JamesMBallard:
@ecmorris please prepare a PR for the v.next branch as well. It's a good idea to fix this on master as well for anyone using Qt 5.14/5.15.

PR #375, opened 2020-06-25
Added old Callout. Confirmed working in C++/QML.

anmacdonald:
Added the Callout. I have not made any attempts to change the Callout beyond getting it running. So bear in mind you are reviewing the old Callout.


import/Callout.qml
```diff
@@ -0,0 +1,1089 @@
+/*******************************************************************************
+ *  Copyright 2012-2018 Esri
```
lsmallwood, line 2: 2020 i guess

anmacdonald:
Well if and when we get round to widgets maybe we can fix it then

lsmallwood:
> Well if and when we get round to widgets maybe we can fix it then

I'm not sure we could do that without breaking users code.

ldanzinger:
I like that we are not trying to fix all the known issues or embark in new enhancements, but as @lsmallwood mentioned, I wonder if we should think more about following the model/view/controller pattern somehow. It would be nice if we had a similar pattern for this tool as the other tools. Can we call a meeting to discuss this?

anmacdonald:
Yeah, I am happy to chat!

PR #376, opened 2020-07-02
Update AuthenticationView.qml

jared-2016:
@ldanzinger Ready for review. I think this should be the last of the sampleviewer related connection warnings. Thanks!


jared-2016:
I didn't realize there are a few more of these in the toolkit repo that need to be fixed. I'll open up another PR once they have been fixed.

PR #380, opened 2020-07-27
Changed the import story for module imports.

anmacdonald:
* Import statement is now `import Esri.ArcGISRuntime.Toolkit 100.9`
instead of `import "qrc://Esri.ArcGISRuntime.Toolkit" as Toolkit`.
* "+cpp_api" folder and boilerplate are no longer required.
* C++ Toolkit now no longer emebeds the QML toolkit's source files into the
application executable.
* User no longer needs to provide the QML engine to the registration
function in the C++ API.

One change is that controllers now exist in the
"Esri.ArcGISRuntime.Toolkit.Controller" module. This was not a
build/runtime requirement but is there to fix Qt creator and its
autocompletion/module discovery issues.


anmacdonald:
@lsmallwood 

Just updated the README now, thanks for that.

@ldanzinger 

In terms of future work/maintenance, I would like to ignore Qt's recommendations entirely. [Qt-6 is removing QML versioning semantics](https://www.qt.io/blog/2019/08/07/technical-vision-qt-6) because of the maintenance burden you describe. So I would expect (hope) we'll be purging versions in 2 or 3 runtime releases anyway.

I'd err on making maintenance as easy for ourselves as possible to maintain in the meantime. I would push for: 
  * We don't keep hold of older sources in the same tree, 
  * When we update the toolkit version we just increase the version number of everything across the board. 

IMO, now that Toolkit wouldn't be released with the SDK anymore, the developer has full control over upgrading/using the toolkit using git tags.

PR #381, opened 2020-07-29
Modern Qt 5.15 syntax for connections.

anmacdonald:
Removing the warnings that come with using Qt 5.15, which is our new minimum.


anmacdonald:
Only within `Connections` type AFAIK, because the slots available are dictated by the target's signals.

PR #382, opened 2020-08-03
Android builds working with Qt 5.15.

anmacdonald:
Android builds are Qt 5.15 compliant.

### DIFF
--- a/Plugin/CppApi/ArcGISRuntimeToolkit.pro
+++ b/Plugin/CppApi/ArcGISRuntimeToolkit.pro
@@ -62,8 +62,6 @@ CONFIG(release, debug|release) {
   BUILDTYPE = debug
 }
 
-!android:!win32: PLATFORM_OUTPUT = $$PLATFORM
-android: PLATFORM_OUTPUT = $$PLATFORM/$$ANDROID_ARCH
 win32: {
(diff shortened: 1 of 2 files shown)


Plugin/CppApi/ArcGISRuntimeToolkit.pro
```diff
@@ -62,17 +62,15 @@ CONFIG(release, debug|release) {
...
   }
 }
+else: {
+    PLATFORM_OUTPUT = $$PLATFORM
+}
 
 DESTDIR = $$PWD/output/$$PLATFORM_OUTPUT
-OBJECTS_DIR = $$DESTDIR/$$BUILDTYPE/obj
-MOC_DIR = $$DESTDIR/$$BUILDTYPE/moc
-RCC_DIR = $$DESTDIR/$$BUILDTYPE/qrc
```
JamesMBallard, line 78: Why remove these?

anmacdonald: With these lines, the object/moc/rcc files for the different android architectures overwrite each other and the linking fails.

I'm not sure why we every tried to enforce in-source builds in the first place. Aside from being bad practice, the default location for build files seems perfectly servicible.

PR #387, opened 2020-10-09
Add font to title and attachment headers

sbiswas96:
This PR sets the `font` property of the title and attachment headers of a PopupView. Previously, updating `popupView.font` would only update the font of the "Close" button.


anmacdonald (approved):
Makes sense, to me, thank-you!

PR #388, opened 2020-10-14
Beta update10

JamesMBallard:
@ldanzinger please review.

The purpose of this PR is to:
 1. Bring the beta branch up to date with v.next so it will eventually merge cleanly into v.next.
 2. Integrate Augmented Reality into the beta branch.

As discussed, AR is its own thing. It does not attempt to integrate or know about the rest of the tools. I have tested both of the AR examples with the new folder structure and they still work. I've restructured the AR files since they no longer integrate with our previous toolkit, which included other tools like CoordinateConversion.

FYI @sbiswas96 @anmacdonald @GuillaumeBelz 

Other, unrelated things I also added (in case you notice them):
 * Update example minimum Qt version to 5.15.1
 * Explicitly restrict Android ABI set to arm64. That is the only one we support for Android.
 * While bringing back AR, none of the other legacy C++ toolkit pieces were brought over, like AbstractTool.h.
 * Add an exemption in .gitignore for the ARCore shared libraries so they are easier to work with in git.


JamesMBallard:
I should add that aside from the folder structure and minor .pri file changes, the AR contents are unmodified. There is no need to rigorously review the AR files and logic.

Import/Esri/ArcGISRuntime/Toolkit/Controls/CppApi/CoordinateConversion.qml
```diff
@@ -0,0 +1,689 @@
+/*******************************************************************************
```
ldanzinger, line 1: I don't think any of these `Import/*` qml files should be included - these are from the old pattern

toolkit.pro
```diff
@@ -0,0 +1,34 @@
+################################################################################
```
ldanzinger, line 1: Do we want this? The folders referred to in OTHER_FILES look to reference the old pattern

ldanzinger (changes requested):

.gitignore
```diff
@@ -34,6 +34,9 @@ html/
 *.so
 *.dylib
 
+# exceptions for AR
+!libarcore*.so
```
anmacdonald, line 38: It may be worth keeping the top-level `.gitignore` minmal, and add two specific `.gitignore` files in the uitools and AR folders if they have different requirements for what needs ignored.

uitools/import/Esri/ArcGISRuntime/Toolkit/Callout.qml
```diff
@@ -280,7 +280,7 @@ Item {
     Connections {
         id: calloutConnection
         target: calloutData
-        onScreenPointChanged: {
```
anmacdonald, line 283: good fix

anmacdonald (commented):
Glad to hear AR importing work!

Please delete the top level `toolkit.pro` and `Import` folder which has been introduced in this PR.

uitools/import/Esri/ArcGISRuntime/Toolkit/Callout.qml
JamesMBallard: I didn't change this, so must be someone else's good work.

.gitignore
JamesMBallard: For now I would like to have a single .gitignore for the entire repo at the root. We've never needed to have different ones, so that sounds like a potential pitfall keeping them in sync, or figuring out why changes to the root .gitignore aren't working (since they're being overridden by a separate one).

If we need more exceptions in the future we can start splitting them out.

JamesMBallard:
>Please delete the top level toolkit.pro and Import folder which has been introduced in this PR.

Good call, I will fix that. Thanks for spotting that one @anmacdonald.

JamesMBallard:
@anmacdonald @ldanzinger ready for review again.

I accidentally brought back some "old toolkit" files but they're gone now.

uitools/import/Esri/ArcGISRuntime/Toolkit/Callout.qml
ldanzinger: @jared-2016 💪

ldanzinger (approved):
LGTM

anmacdonald (approved):
LGTM.

One note: `uitools` is all lower case, maybe `AugementedReality` should be snake_case?

JamesMBallard:
>One note: uitools is all lower case, maybe AugementedReality should be snake_case?

@ldanzinger thoughts or preferences on that? It's a good point. I have no strong preference. I like "AugmentedReality" but it's not consistent with the other naming.

ldanzinger:
Not a strong preference, but perhaps they should all be lower case? The plugins in the Qt kits are all lowercase. 

image

JamesMBallard:
I like it. We will follow suit then. I will make the changes, retest, push and merge.

PR #393, opened 2020-10-23
Update Callout.qml

ldanzinger:
@JamesMBallard please review


JamesMBallard:
This should get picked up when we merge beta_update10 into v.next since it's already in v.next. Should this go into "beta" for current beta users?

JamesMBallard (approved):
👍  all good, but not sure if this is the correct branch to merge to.

PR #397, opened 2020-10-30
James m ballard patch 1

JamesMBallard:


JamesMBallard:
This should fix the mac build errors. My mac is currently committed to other work, so I am testing on Windows. I'll merge as long as the Windows build is good.

PR #399, opened 2020-11-11
adding integration info to the readme

ldanzinger:
@anmacdonald please review. While testing the toolkit integration, I thought it would be helpful to have some high level instructions for each paradigm.


ldanzinger:
I guess I totally missed that this info is in a readme inside the uitools folder. Closing...

anmacdonald:
Hi Lucas, it might be worth linking to the internal README from the outer README!

PR #413, opened 2021-04-06
Revert "workaround for 5.15.1 regression"

ldanzinger:
Reverts Esri/arcgis-runtime-toolkit-qt#408

@JamesMBallard please review/merge

This was a workaround, and now the behavior seems to be fixed in Qt 5.15.2+

I tested with the current v.next and ran into the issue reported by the release team, and with this change, the sample and test app works as expected


JamesMBallard (approved):
This is great that we can remove workarounds!

PR #425, opened 2021-07-09
Updating the styling of BasemapGallery.

anmacdonald:
Based on feedback from a week ago. Here is the styling update to put the text below/next to the image instead of above it.

image
image


ldanzinger:
@anmacdonald 2 things:

1) I believe the text on side vs under is backwards from what the JS implementation is.
2) Can we soften the selection color a bit like the JS implementation?

image

image

https://developers.arcgis.com/javascript/latest/sample-code/widgets-basemapgallery/

anmacdonald:
Updated the text order:

image
image

ldanzinger (approved):
Looks good.

PR #426, opened 2021-07-28
fixing docs from doc build

ldanzinger:
@anmacdonald I ran the doc build and noticed a few issues. There are still several AR related doc build issues, but this addresses most of the UI Tool issues.


anmacdonald (approved):
LGTM. Thanks for doing this cleanup.

PR #428, opened 2021-08-11
Anmacdonald/basemap geomodel

anmacdonald:
This has come as some last-minute feedback from .NET to change the BasemapGallery design. 

The following changes have been made:

* BasemapGallery no longer makes uses of GeoView but instead takes a GeoModel.
* GeoModel is an optional parameter, the gallery will still emit signals if a new basemap is selected, but will only apply updates to the geoModel if one is provided.

### DIFF
--- a/uitools/cpp/Esri/ArcGISRuntime/Toolkit/BasemapGalleryController.cpp
+++ b/uitools/cpp/Esri/ArcGISRuntime/Toolkit/BasemapGalleryController.cpp
@@ -53,16 +53,16 @@ namespace Toolkit {
       map/scene basemapChanged signal is fired.
      */
     template <typename T>
-    void connectToBasemap(BasemapGalleryController* self, T* mapOrScene)
+    void connectToBasemap(BasemapGalleryController* self, T* geoModel)
     {
       static_assert(std::is_base_of<GeoModel, T>::value, "Must be a GeoModel.");
 
-      if (!mapOrScene)
+      if (!geoModel)
         return;
 
-      auto listenToLoadSignals = [self, mapOrScene]
+      auto listenToLoadSignals = [self](Basemap* basemap)
       {
-        if (auto basemap = mapOrScene->basemap())
+        if (basemap)
         {
           if (basemap->loadStatus() != LoadStatus::Loaded)
             QObject::connect(basemap, &Basemap::doneLoading, self, &BasemapGalleryController::currentBasemapChanged);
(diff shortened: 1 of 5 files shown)


uitools/cpp/Esri/ArcGISRuntime/Toolkit/BasemapGalleryController.cpp
```diff
@@ -606,7 +536,7 @@ namespace Toolkit {
   /*!
    \internal
    \brief Given a \a basemap, returns whether the spatial reference of its layers
-   match the spatial reference of the GeoView (and therefore if it appropriate to apply
+   match the spatial reference of the GeoModel (and therefore if it appropriate to apply
```
ldanzinger, line 539:
```suggestion
   match the spatial reference of the GeoModel (and therefore if it is appropriate to apply
```

uitools/cpp/Esri/ArcGISRuntime/Toolkit/BasemapGalleryController.cpp
```diff
@@ -647,40 +565,74 @@ namespace Toolkit {
...
   }
 
-/*!
-  \fn void Esri::ArcGISRuntime::Toolkit::BasemapGalleryController::geoViewChanged()
-  \brief Emitted when the geoView has changed.
+  /*!
+   \brief Convenience function for QML/C++ users which allows the map/scene to be extracted from a
+   SceneView or MapView assigned to \a view in QML code.
+
+  This is only a concern as [Map/Scene]QuickView does not expose a [Map/Scene] property in QML.
```
ldanzinger, line 572:
```suggestion
   This is only a concern as [Map/Scene]QuickView does not expose a [Map/Scene] property in QML.
```

uitools/cpp/Esri/ArcGISRuntime/Toolkit/BasemapGalleryController.cpp
```diff
@@ -647,40 +565,74 @@ namespace Toolkit {
                        });
   }
 
-/*!
-  \fn void Esri::ArcGISRuntime::Toolkit::BasemapGalleryController::geoViewChanged()
-  \brief Emitted when the geoView has changed.
+  /*!
+   \brief Convenience function for QML/C++ users which allows the map/scene to be extracted from a
```
ldanzinger, line 569: Good thinking

ldanzinger (approved):
LGTM. Good cleanup. Also good thinking for the Qt Quick C++ addition to allow a view to be passed through instead. Maybe in the future we could add a Q_PROPERTY on MapView for Map and SceneView for Scene in the Runtime API so that users can just pass the model through and we don't need to worry about the view

uitools/cpp/Esri/ArcGISRuntime/Toolkit/BasemapGalleryController.cpp
```diff
@@ -99,73 +72,62 @@ namespace Toolkit {
...
+      }
+
+      if (auto scene = qobject_cast<Scene*>(geoModel))
+      {
+        connectToBasemap(self, scene);
+      }
+      else if (auto map = qobject_cast<Map*>(geoModel))
+      {
+        connectToBasemap(self, map);
+      }
```
JamesMBallard, line 117: Make a note. This can likely be cleaned up next release once the basemapChanged signal is moved to GeoModel.

uitools/cpp/Esri/ArcGISRuntime/Toolkit/BasemapGalleryController.cpp
```diff
@@ -494,52 +435,41 @@ namespace Toolkit {
...
+      return;
+
+    m_currentBasemap = basemap;
+    emit currentBasemapChanged();
+
+    if (m_geoModel)
     {
-      auto scene = sceneView->arcGISScene();
-      if (scene)
+      if (auto scene = qobject_cast<Scene*>(m_geoModel))
```
JamesMBallard, line 466: Is this still needed (the casting)? `setBasemap` is common on GeoModel.

JamesMBallard (approved):
👍

PR #431, opened 2021-08-18
Fixing linux build failure with some gcc compilers.

anmacdonald:
Fix for certain supported GCC compilers not parsing initalizer syntax correctly.


ldanzinger (approved):
tested on linux and this fixes the issue for me. i also built successfully on macos. haven't tried the other compilers

PR #438, opened 2021-09-21
Update Show Popup sample UI

Gela:
This pull request updates the existing show popup sample. Currently the sample has "next" and "previous" buttons that are not functional and are always greyed out. Additionally, the "Close" button used to close the popup covered the title so I moved it to the bottom of the popup and it no longer covers any text. 

Supplement PR in the SDK samples repo: 

Please let me know if I should assign anyone else to review.

**Old UI:**

old_UI

**New UI:**

popup_new_UI

### DIFF
--- a/uitools/import/Esri/ArcGISRuntime/Toolkit/PopupView.qml
+++ b/uitools/import/Esri/ArcGISRuntime/Toolkit/PopupView.qml
@@ -90,7 +90,7 @@ Control {
        Defaults to setting visible to false.
      */
     property var closeCallback: function () {
-        popupView.visible = false
+        popupView.visible = false;
     }
 
 
--- a/uitools/import/Esri/ArcGISRuntime/Toolkit/PopupView.qml
+++ b/uitools/import/Esri/ArcGISRuntime/Toolkit/PopupView.qml
@@ -1,3 +1,5 @@
+
+
 /*******************************************************************************
  *  Copyright 2012-2020 Esri
  *
@@ -22,6 +24,7 @@ import QtQuick.Dialogs 1.2
 import QtQuick.Window 2.11
 import QtQuick.Layouts 1.3
 
+
 /*!
    \qmltype PopupView
    \ingroup ArcGISQtToolkit
@@ -29,7 +32,7 @@ import QtQuick.Layouts 1.3
    \inqmlmodule Esri.ArcGISRuntime.Toolkit
    \since Esri.ArcGISRuntime 100.10
    \brief A view for displaying and editing information about a feature.
-  
+
    A PopupView can be used to display information for any type that
    implements the PopupSource interface. For example, FeatureLayer
    implements PopupSource. This means that it has a PopupDefinition,
@@ -61,21 +64,24 @@ import QtQuick.Layouts 1.3
 Control {
     id: popupView
 
+
     /*!
        \brief The PopupManager that controls the information being displayed in
        the view.
-       
+
        The PopupManager should be created from a Popup.
        \qmlproperty PopupManager popupManager
      */
     property var popupManager: null
-    
+
+
     /*!
       \qmlproperty PopupViewController controller
       \brief the Controller handles reading from the PopupManager and monitoring
       the list-models.
     */
-    property var controller: PopupViewController { }
+    property var controller: PopupViewController {}
+
 
     /*!
        \brief Callback function called when the close button is clicked. When
@@ -83,9 +89,10 @@ Control {
        the close button is clicked the function in this property is called.
        Defaults to setting visible to false.
      */
-    property var closeCallback: function() {
-        popupView.visible = false;
+    property var closeCallback: function () {
+        popupView.visible = false
     }
+
 
     /*!
        \qmlsignal PopupView::attachmentThumbnailClicked(var index)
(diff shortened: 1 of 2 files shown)


JamesMBallard:
The next and previous buttons are functional if you are able to select multiple items at once and navigate through them. I don't know if that particular sample supports that, but it is a supported workflow.

tanneryould:
> The next and previous buttons are functional if you are able to select multiple items at once and navigate through them. I don't know if that particular sample supports that, but it is a supported workflow.

I was just about to say that. I think the sample may be worth refactoring to return multiple items at once.

tanneryould:
Screen Shot 2021-09-21 at 11 15 18 AM

tanneryould:
Screen Shot 2021-09-21 at 11 37 55 AM

Alright, I resolved the issue and I'll open another PR against this one for you to merge in @Gela

anmacdonald:
I can confirm that the next/previous buttons are required in the multiple case. I've tried your branch in the DSA, where popup stack view is used extensively, and I am happy with the results.

https://user-images.githubusercontent.com/42203545/134360133-b9532eb9-4e74-492f-8992-e5e576f349ad.mp4

The removal of the buttons in the one-selection case I think is a nice addition.

uitools/import/Esri/ArcGISRuntime/Toolkit/PopupView.qml
```diff
@@ -216,6 +213,20 @@ Control {
...
             }
         }
+
+        Button {
+            anchors {
+                top: fieldsLayout.bottom
+                topMargin: 20
+                horizontalCenter: parent.horizontalCenter
+            }
+            Layout.alignment: Qt.AlignHCenter
```
anmacdonald, line 223: :x: If you are removing the button from the Layout I don't think this line is required.

uitools/import/Esri/ArcGISRuntime/Toolkit/PopupView.qml
```diff
@@ -216,6 +213,20 @@ Control {
                 }
             }
         }
+
+        Button {
+            anchors {
```
anmacdonald, line 218: :thought_balloon: `GridLayout` exists to both size and poition child content. I see you've removed it from the Layout. I think in this case, this is fine as you only need to position the close button and not size it.

uitools/import/Esri/ArcGISRuntime/Toolkit/PopupView.qml
```diff
@@ -216,6 +213,20 @@ Control {
...
+                top: fieldsLayout.bottom
+                topMargin: 20
```
anmacdonald, line 220: :x: Try not hard-coding margin/padding values into toolkit QML. Sometimes this can make tools look "off" when the developer applies a QML style on top of the toolkit item. I'm attempting to actively expunge the hard-coded values that still actively exist. 

If you are attempting to put some space between two content items, a better approach is to set `topMargin: popupView.spacing`. The property [spacing](https://doc.qt.io/qt-5/qml-qtquick-controls2-control.html#spacing-prop) is designed for this use case.

You can set the default spacing of the `PopupView` to 20 if you wish - but this at least gives the developer a meaningful chance to tweak the value if the styling doesn't apply right.

anmacdonald (changes requested):
👍 This is a good quality of life improvement, thanks for improving the PopupView. I have some minor observations about the style properties.

uitools/import/Esri/ArcGISRuntime/Toolkit/PopupView.qml
Gela: I was not aware of that, but that totally makes sense. I'm removing it for this PR and will avoid this in the future

Gela: @anmacdonald Removing the hardcoded value of 20 and replacing it with spacing does not seem to work and so the close button ends up awkwardly close to the text above it. Is there a different property I could try? I'm going to play around with the qml and see if I can figure it out but you may have the answer a lot quicker :) 

spacing

anmacdonald: Sorry Gela, that's my fault for not being clear.

I'll take the chance to go into a bit more detail. The default spacing of `Control`  is 0, and it's up to the inherited types like `Button`, `DialogButtonBar` etc to set sensible defaults. You can see this happen under `$QT_INSTALL/qml/QtQuick/Controls.2`

For example, with Button, we can see this happen in `$QT_INSTALL/qml/QtQuick/Controls.2/Button.qml`:

```qml
T.Button {
    id: control

    implicitWidth: Math.max(implicitBackgroundWidth + leftInset + rightInset,
                            implicitContentWidth + leftPadding + rightPadding)
    implicitHeight: Math.max(implicitBackgroundHeight + topInset + bottomInset,
                             implicitContentHeight + topPadding + bottomPadding)

    padding: 6
    horizontalPadding: padding + 2
    spacing: 6
    ...
```

In this case, `Popup` also inherits directy from `Control` (my fault - it would be better if it it inherited from `Frame`), so you also need to set a sensible `spacing` default.

So I'd put `spacing: 20` somewhere on the `PopupStackView`, and as an inherited propery it will allow the developer to change the spacing value on `PopupStackView` without having to go in and edit the source code.

uitools/import/Esri/ArcGISRuntime/Toolkit/PopupView.qml
```diff
@@ -107,18 +112,10 @@ Control {
...
-            color: palette.shadow
-            width: 2
-        }
-        radius: 2
-    }
-
     contentItem: Flickable {
         id: flickable
         clip: true
+        topMargin: popupView.spacing
```
anmacdonald, line 118: Should this be in line 213?

anmacdonald (changes requested):
Almost there! Sorry, just one final query.

PR #440, opened 2021-09-23
set the implicit width and height to 32

mrwillyees:
fix to default height and width were 0. set to 32
it might be too small 32x32
image

### DIFF
--- a/uitools/import/Esri/ArcGISRuntime/Toolkit/NorthArrow.qml
+++ b/uitools/import/Esri/ArcGISRuntime/Toolkit/NorthArrow.qml
@@ -27,7 +27,7 @@ import QtQuick 2.12
    heading of the SceneView. 
    Double clicking on the NorthArrow triggers the heading of the connected
    GeoView to be orientainted to 0.
-   \note default width and height is 32.
+   \note default width and height is 48.
  */
 
 Item {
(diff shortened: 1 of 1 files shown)


anmacdonald:
I see what you mean about the size being too small. My bad for picking the numbers out of a hat.

Looking at the actual image size it is 48x48: https://github.com/Esri/arcgis-runtime-toolkit-qt/blob/main/uitools/images/compass.png

I think that would be more sensible?

mrwillyees:
changed to 48x48 as original png
image

uitools/import/Esri/ArcGISRuntime/Toolkit/NorthArrow.qml
```diff
@@ -27,6 +27,7 @@ import QtQuick 2.12
    heading of the SceneView. 
    Double clicking on the NorthArrow triggers the heading of the connected
    GeoView to be orientainted to 0.
+   \note default width and height is 32.
```
anmacdonald, line 30: ❌  Remember to update the documentation too!

anmacdonald (changes requested):
👍  Good stuff

Please test this change with both a Qt Quick C++ App, and a QML App to confirm.

After updating the documentation, if both apps work, feel free to merge using the `squash and merge` option. No need to re-review.

PR #441, opened 2021-09-24
Andrewbladon/basemap gallery set basemap fix

AndrewBladon:
This PR adds a check to see if the Scene/Map view has been initialised prior to passing an argument into the setBasemap() method. If the Scene/Map view has been initialised, this is passed as an argument. If the view has not been initialised, null is passed as an argument.

A misspelling of "connections" in geoModelConnections has also been resolved.

@anmacdonald, please could you review these minor changes?


anmacdonald (approved):
👍 Looks great. Thank you for doing this.

PR #445, opened 2021-09-27
default size widget northarrow

mrwillyees:
setting also for the widget qt northarrow an inital size as 48x48. if the png is changed later on, the size will still be 48x48


anmacdonald (approved):
LGTM. Thank you for looking into this.

PR #447, opened 2021-09-29
BasemapGallery tooltips: enable links and adjust timing

AndrewBladon:
Html links included in tooltip's in the Basemap Gallery cannot be clicked. This PR adds a parser to remove any html tags [e.g. `<html>` and `</html>`] from tooltip text. This means the text appears as plain text and does not appear as a link.

### DIFF
--- a/uitools/import/Esri/ArcGISRuntime/Toolkit/BasemapGallery.qml
+++ b/uitools/import/Esri/ArcGISRuntime/Toolkit/BasemapGallery.qml
@@ -246,6 +246,34 @@ Pane {
                 anchors.fill: parent
                 hoverEnabled: true
                 onClicked: controller.setCurrentBasemap(modelData.basemap)
+
+                // When mouse enters thumbnail area, use timer to delay showing of tooltip.
+                onEntered: {
+                    timerOnEntered.start();
+                }
+                // When mouse exits thumbnail area, use timer to delay hiding of tooltip.
+                onExited: {
+                    timerOnExited.start();
+                }
+
+                Timer {
+                    id: timerOnEntered
+                    interval: 1000
+                    repeat: false
+                    onTriggered: {
+                        if (allowTooltips && mouseArea.containsMouse && modelData.tooltip !== "")
+                            basemapDelegate.ToolTip.visible = true;
+                    }
+                }
+
+                Timer {
+                    id: timerOnExited
+                    interval: 1000
+                    repeat: false
+                    onTriggered: {
+                        basemapDelegate.ToolTip.visible = false;
+                    }
+                }
             }
         }
         highlightFollowsCurrentItem: false
--- a/uitools/import/Esri/ArcGISRuntime/Toolkit/BasemapGallery.qml
+++ b/uitools/import/Esri/ArcGISRuntime/Toolkit/BasemapGallery.qml
@@ -202,7 +202,9 @@ Pane {
             Connections {
                 target: basemapDelegate.ToolTip.toolTip.contentItem
                 enabled: basemapDelegate.ToolTip.visible
-                function onLinkActivated(link){Qt.openUrlExternally(link)}
+                function onLinkActivated(link){
+                    Qt.openUrlExternally(link)
+                }
             }
             ToolTip.text: modelData.tooltip
 
@@ -248,17 +250,17 @@ Pane {
                 onClicked: controller.setCurrentBasemap(modelData.basemap)
 
                 // When mouse enters thumbnail area, use timer to delay showing of tooltip.
-                onEntered: {
-                    timerOnEntered.start();
-                }
+                onEntered: timerOnEntered.start();
+
                 // When mouse exits thumbnail area, use timer to delay hiding of tooltip.
                 onExited: {
+                    timerOnEntered.stop();
                     timerOnExited.start();
                 }
 
                 Timer {
                     id: timerOnEntered
-                    interval: 1000
+                    interval: Qt.styleHints.mousePressAndHoldInterval * 2
                     repeat: false
                     onTriggered: {
                         if (allowTooltips && mouseArea.containsMouse && modelData.tooltip !== "")
@@ -268,11 +270,9 @@ Pane {
 
                 Timer {
                     id: timerOnExited
-                    interval: 1000
+                    interval: Qt.styleHints.mousePressAndHoldInterval * 1.9
                     repeat: false
-                    onTriggered: {
-                        basemapDelegate.ToolTip.visible = false;
-                    }
+                    onTriggered: basemapDelegate.ToolTip.visible = false;
                 }
             }
         }
--- a/uitools/import/Esri/ArcGISRuntime/Toolkit/BasemapGallery.qml
+++ b/uitools/import/Esri/ArcGISRuntime/Toolkit/BasemapGallery.qml
@@ -202,7 +202,7 @@ Pane {
             Connections {
                 target: basemapDelegate.ToolTip.toolTip.contentItem
                 enabled: basemapDelegate.ToolTip.visible
-                function onLinkActivated(link){
+                function onLinkActivated(link) {
                     Qt.openUrlExternally(link)
                 }
             }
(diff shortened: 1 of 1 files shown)


ldanzinger:
Is it possible to use something like `onLinkActivated: Qt.openUrlExternally(link)` 

https://stackoverflow.com/questions/12536416/qml-text-element-hyperlink

AndrewBladon:
Thanks for your comments @anmacdonald and @ldanzinger.

Using a `Connection` we have managed to use the `Qt.openUrlExternally(link)` method to enable the link in the tooltip to be opened.

With this change, the timing of the tooltip needed to be modified.  Currently, if the mouse leaves the thumbnail `MouseArea`, the tooltip will disappear. This means that a user cannot click the link in the tooltip because the tooltip disappears. Two properties are available with attached tooltips: `delay` and `timeout`, however, the timeout duration appears to be from the time the tooltip appears rather than the time the mouse leaves the `MouseArea` and is not suitable for this. As such, delays to the appearance and disappearance of the tooltips were manually created using two `Timer` objects and the `onEntered` and `onExited` slots of the `MouseArea` object. 

Please could you review these updated changes?

**NOTE:** I have modified the title of the PR from `Remove html links from BasemapGallery tooltips` to `BasemapGallery tooltips: enable links and adjust timing`. This is to reflect the changes to the purpose of the work.

ldanzinger (approved):
LGTM

uitools/import/Esri/ArcGISRuntime/Toolkit/BasemapGallery.qml
```diff
@@ -241,6 +246,34 @@ Pane {
...
+                    repeat: false
+                    onTriggered: {
+                        if (allowTooltips && mouseArea.containsMouse && modelData.tooltip !== "")
+                            basemapDelegate.ToolTip.visible = true;
+                    }
+                }
+
+                Timer {
+                    id: timerOnExited
+                    interval: 1000
```
anmacdonald, line 271: :thought_balloon: What's the look and feel of this? There were complaints previously that the tooltip appeared to quickly, and 1 second feels relatively short. it might be worth comparing to:

```qml
Qt.styleHints.mousePressAndHoldInterval * 2
```

uitools/import/Esri/ArcGISRuntime/Toolkit/BasemapGallery.qml
```diff
@@ -241,6 +246,34 @@ Pane {
...
                 hoverEnabled: true
                 onClicked: controller.setCurrentBasemap(modelData.basemap)
+
+                // When mouse enters thumbnail area, use timer to delay showing of tooltip.
+                onEntered: {
+                    timerOnEntered.start();
+                }
+                // When mouse exits thumbnail area, use timer to delay hiding of tooltip.
+                onExited: {
+                    timerOnExited.start();
```
anmacdonald, line 256: 🔧  Should probably kill the `timerOnEntered` timer here too.

uitools/import/Esri/ArcGISRuntime/Toolkit/BasemapGallery.qml
```diff
@@ -195,12 +195,17 @@ Pane {
...
+            id: basemapDelegate
             width: view.cellWidth
             height: view.cellHeight
             enabled: controller.basemapMatchesCurrentSpatialReference(modelData.basemap)
-            ToolTip.visible: allowTooltips && mouseArea.containsMouse && modelData.tooltip !== ""
-            ToolTip.delay: Qt.styleHints.mousePressAndHoldInterval * 2
+            Connections {
+                target: basemapDelegate.ToolTip.toolTip.contentItem
+                enabled: basemapDelegate.ToolTip.visible
+                function onLinkActivated(link){Qt.openUrlExternally(link)}
```
anmacdonald, line 205: :x: add some newlines here.

uitools/import/Esri/ArcGISRuntime/Toolkit/BasemapGallery.qml
```diff
@@ -241,6 +246,34 @@ Pane {
                 anchors.fill: parent
                 hoverEnabled: true
                 onClicked: controller.setCurrentBasemap(modelData.basemap)
+
+                // When mouse enters thumbnail area, use timer to delay showing of tooltip.
+                onEntered: {
```
anmacdonald, line 251: 🙃 Consider making this one line. Don't need the braces.

uitools/import/Esri/ArcGISRuntime/Toolkit/BasemapGallery.qml
```diff
@@ -241,6 +246,34 @@ Pane {
...
+                    timerOnEntered.start();
+                }
+                // When mouse exits thumbnail area, use timer to delay hiding of tooltip.
+                onExited: {
```
anmacdonald, line 255: 🙃 Consider making this one line. Don't need the braces.

uitools/import/Esri/ArcGISRuntime/Toolkit/BasemapGallery.qml
```diff
@@ -241,6 +246,34 @@ Pane {
...
+                    timerOnExited.start();
+                }
+
+                Timer {
```
anmacdonald, line 259: :thought_balloon: This is not a hard requirement. But with this implementation there are now 2 timers per basemapgallery item which is pretty costly. 

Since we only show 1 tooltip, there is possibly a saving to be made here by only having two timers for the entire gallery, if we keep them outside the gallery.

It would require passing a reference to the currently hovered basemap to the timer somehow.

anmacdonald (changes requested):
👍 Well done for braving the world of QML timers. It's a good solution. Couple of comments to think about.

uitools/import/Esri/ArcGISRuntime/Toolkit/BasemapGallery.qml
AndrewBladon: The `Qt.styleHints.mousePressAndHoldInterval` is 800ms.

To be consistent with the comments received previously, the interval of the two timers have been changed to `Qt.styleHints.mousePressAndHoldInterval * 2`.

AndrewBladon:
Thanks for your feedback @anmacdonald and for your assistance making the changes. The changes are ready for re-review.

uitools/import/Esri/ArcGISRuntime/Toolkit/BasemapGallery.qml
```diff
@@ -255,6 +282,26 @@ Pane {
...
     }
 
+    Timer {
+        id: timerOnEntered
+
+        property var showTooltipFn: null;
+
+        interval: Qt.styleHints.mousePressAndHoldInterval * 2
+        repeat: false
+        onTriggered: showTooltipFn();
```
anmacdonald, line 292: 🔧 For completeness might be worth adding a null check here to be safe.

uitools/import/Esri/ArcGISRuntime/Toolkit/BasemapGallery.qml
```diff
@@ -255,6 +282,26 @@ Pane {
...
+    }
+
+    Timer {
+        id: timerOnExited
+
+        property var hideTooltipFn: null;
+
+        interval: Qt.styleHints.mousePressAndHoldInterval * 1.9
+        repeat: false
+        onTriggered: hideTooltipFn();
```
anmacdonald, line 302: 🔧 For completeness might be worth adding a null check here to be safe.

uitools/import/Esri/ArcGISRuntime/Toolkit/BasemapGallery.qml
```diff
@@ -195,12 +195,19 @@ Pane {
...
+            id: basemapDelegate
             width: view.cellWidth
             height: view.cellHeight
             enabled: controller.basemapMatchesCurrentSpatialReference(modelData.basemap)
-            ToolTip.visible: allowTooltips && mouseArea.containsMouse && modelData.tooltip !== ""
-            ToolTip.delay: Qt.styleHints.mousePressAndHoldInterval * 2
+            Connections {
+                target: basemapDelegate.ToolTip.toolTip.contentItem
+                enabled: basemapDelegate.ToolTip.visible
+                function onLinkActivated(link){
```
anmacdonald, line 205: 🙃 Add a space here.

anmacdonald (approved):
👍 Looking good! Some final minor comments. No need to re-review though.

PR #449, opened 2021-09-29
Add onLoadStatusChanged to TimeSliderController

Gela:
This fix addresses a bug where the time slider in QML does not refresh when the time extent changes.


anmacdonald (approved):
LGTM thanks for doing this.

PR #451, opened 2021-10-05
Remove brackets from hideTooltipFn null check

AndrewBladon:
This PR removes the brackets from `hideTooltipFn()` in a null check expression.

@anmacdonald, please could you confirm that the null check is now correct? Thanks.


anmacdonald (approved):
LGTM

PR #452, opened 2021-10-05
Cache timeExtent prior to assigning to geoView

Gela:
This change addresses a bug in the QML toolkit where the map does not redraw as we step through the time aware data using the play button.

This is a work around for a known existing JS bug.


anmacdonald:
@Gela Based on recent findings, is this okay to close out?

Gela:
Yes I believe this change is no longer needed. Closing.

PR #453, opened 2021-10-07
Awilliams/basemap widg highlight

mrwillyees:
Branch encapsulating the feature to disable the highlight of not supported basemaps.
Achieved by overriding the .flag() in GenericModel and added flags in the GalleryItem.

- Unfortunately, just setting the flags as 'notselectable | notenabled', the listview will let the click go through, so a check on the item flag currently set is needed to be done in the view part of the component.
- Not available basemaps are shown as grayed out icons, but there is still a blue background around them, not helping in hinting that they are not available. If a user clicks, nothing happens, so they should get the idea of disabled


mrwillyees:
This PR was incorporated into the #456, don't need to going forward from here

PR #463, opened 2021-11-10
Awilliams/calcite changes

mrwillyees:
Text and Rectangle changes.
A few Rectangles were not changed into Pane because need the radius property which is missing in the Pane.
Any other "low level" components to scout for?

### DIFF
--- a/uitools/import/Esri/ArcGISRuntime/Toolkit/Callout.qml
+++ b/uitools/import/Esri/ArcGISRuntime/Toolkit/Callout.qml
@@ -17,6 +17,7 @@
 import QtQuick 2.11
 import QtQuick.Window 2.11
 import QtQuick.Layouts 1.3
+import QtQuick.Controls 2.15
 import "LeaderPosition.js" as Enums
 
 /*!
@@ -490,7 +491,7 @@ Item {
                         }
                     }
 
-                    Text {
+                    Label {
                         id: title
                         text: calloutData ? calloutData.title : ""
                         wrapMode: Text.NoWrap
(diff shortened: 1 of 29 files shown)


uitools/import/Esri/ArcGISRuntime/Toolkit/Callout.qml
```diff
@@ -473,11 +488,17 @@ Item {
                     rows: 2
                     columnSpacing: 7
 
-                    Rectangle {
+                    Pane {
```
anmacdonald, line 491: ❌  Honestly, I won't touch Callout at all. There's too much else wrong with it.

Plus if we are setting the Pane, then hard-overwriting the color, then there was no point setting Pane at all, we should have left this as a Rectangle.

uitools/import/Esri/ArcGISRuntime/Toolkit/SearchView.qml
```diff
@@ -327,12 +327,19 @@ Pane {
...
                 height: list.currentItem ? list.currentItem.height : 0
-                color: searchView.palette.highlight
+                background: Rectangle {
+                    color: searchView.palette.highlight
+                }
+                leftPadding: 0
+                rightPadding: 0
+                topPadding: 0
+                bottomPadding: 0
+
```
anmacdonald, line 342: ❌  I'd leave this as a rectangle if we are customizing the color.

💭  The real problem I believe is we're using the wrong kind of Delegate, we should be using an `ItemDelegate` and doing the highlighting via that's inbuilt highlighting methods, see the BasemapGallery for example. This is not a request to change the rendering of SearchView, more just a comment of what the "proper" fix should be.

anmacdonald (changes requested):
Looking good, thanks for updating PopupView.

I don't think the Callout changes are entirely appropriate however, or the change to highlight in SearchView.

anmacdonald (approved):
LGTM

PR #464, opened 2021-11-10
changed group search icon to match ios version

mrwillyees:
changed group icon to match the ios version. At the moment using the same icon as the search icon. Below is a screen shot showing a different icon if preferred.
groupicons


anmacdonald (approved):
LGTM

ldanzinger (commented):
Couple questions/comments for you and @anmacdonald :

* do we still need pins.svg or can it be removed?
* it seems like the icon and text are vertically top aligned vs centered aligned. 
* the icons seem oversized compared to the text, especially compared to .NET and iOS where the icons are about the same height as text, or maybe slightly smaller

mrwillyees:
smaller icon and centered text
image

previous version:
groupicons

ldanzinger (approved):
LGTM

PR #466, opened 2021-11-17
Example apps

anmacdonald:
image

This PR covers the creation of the Toolkit QML demo and and Toolkit C++ demo app. There are no changes to the toolkit components themselves within this PR.

It is recommended the reviewer goes through this PR commit-by-commit. 

Currently the README.md is empty which we plan to fill in in a seperate PR.


uitools/examples/cpp_quick/src/demos/OverviewMapDemo.qml
```diff
@@ -0,0 +1,52 @@
...
+            id: view
+            OverviewMap {
+                anchors {
+                    top: parent.top
+                    left: parent.left
+                    margins: 10
+                }
+                geoView: parent
+            }
+            OverviewMapDemoModel {
```
lsmallwood, line 47: 💭 I'm not quite sure that the term "model" is right for these types in a MVC sense... they know about the controller and are present in the ui layer which is not what i'd expect from a model.

uitools/examples/cpp_quick/src/demos/PopupViewDemo.qml
```diff
@@ -0,0 +1,66 @@
...
+            PopupView {
+                id:popupView
+                anchors {
+                     left: parent.left
+                     top: parent.top
+                     bottom: parent.bottom
+                }
+                visible: false
+                popupManager: model.popupManager
+
```
lsmallwood, line 56: empty line

uitools/examples/cpp_quick/src/demos/PopupViewDemoModel.cpp
```diff
@@ -0,0 +1,119 @@
...
+}
+
+void PopupViewDemoModel::setUp()
+{
+  apply([this](auto geoView)
+        {
+          using ViewType = std::remove_pointer_t<decltype(geoView)>;
+          connect(geoView, &ViewType::mouseClicked, this, [this, geoView](QMouseEvent& mouse)
+                  {
+                    m_featureLayer = static_cast<FeatureLayer*>(geoModel()->operationalLayers()->first());
```
lsmallwood, line 69: you can check the `LayerType` of the feature layer before static casting. Looks like this sample uses an online web map so if someone was to change that map this will crash

uitools/examples/cpp_quick/src/demos/TimeSliderDemoModel.cpp
```diff
@@ -0,0 +1,14 @@
...
+#include "Scene.h"
+
+TimeSliderDemoModel::TimeSliderDemoModel(QObject* parent) :
+  BaseDemoModel(parent)
+{
+}
+
+TimeSliderDemoModel::~TimeSliderDemoModel()
+{
+}
```
lsmallwood, line 14: a lot of files don't have a new line at the end. That causes some IDEs and things to grumble...

uitools/examples/cpp_quick/src/main.cpp
```diff
@@ -0,0 +1,97 @@
...
+    // https://links.esri.com/arcgis-runtime-security-auth.
+
+    // 1. ArcGIS identity: An ArcGIS named user account that is a member of an
+    // organization in ArcGIS Online or ArcGIS Enterprise.
+
+    // 2. API key: A permanent key that gives your application access to Esri
+    // location services. Create a new API key or access existing API keys from
+    // your ArcGIS for Developers dashboard (https://links.esri.com/arcgis-api-keys).
+
+    const QString apiKey = QString("");
```
lsmallwood, line 49: 💭 @AndrewBladon added some code to the samples recently that lets you set an api key via the cmd line. That means you don't have to compile it in which is quite nice

uitools/examples/cpp_quick/src/proxies/EnumsProxy.h
```diff
@@ -0,0 +1,26 @@
...
+class EnumsProxy : public QObject
+{
+  Q_OBJECT
+public:
+  Q_INVOKABLE explicit EnumsProxy(QObject* parent = nullptr);
+  ~EnumsProxy() override;
+
+  enum class LoadStatus 
+  {
+    LoadStatusLoaded = static_cast<int>(Esri::ArcGISRuntime::LoadStatus::Loaded),
```
lsmallwood, line 17: we should just get these exposed to qml from the api if users will need them

uitools/examples/qml_quick/src/demos/CalloutDemo.qml
```diff
@@ -0,0 +1,55 @@
...
+                accessoryButtonHidden: true
+                leaderPosition: leaderPositionEnum.Automatic
+
+                property Point location;
+            }
+            onMouseClicked : {
+                if (callout.calloutVisible) {
+                    callout.dismiss();
+                } else {
+                    xClickLoc = mouse.mapPoint.x.toFixed(2);
```
lsmallwood, line 47: i'm surprised you don't need to do `screenToLocation` with the view  in QML?

lsmallwood (commented):
Looks really cool @anmacdonald and @mrwillyees. I only did a quick skim through so just a few comments

uitools/examples/qml_quick/src/demos/CalloutDemo.qml
anmacdonald: I believe we get away with this one because we expose our own `MouseEvent` in QML, which returns a`Point` for us to use: https://developers.arcgis.com/qt/qml/api-reference/qml-esri-arcgisruntime-mouseevent.html#mapPoint-prop

uitools/examples/cpp_quick/src/proxies/EnumsProxy.h
anmacdonald: I'd love to see this proxies folder removed in its entireity. Currently, I think `ArcGISRuntimeEnvironment.apiKey` is the lowest hanging fruit. Enumerations might require a bit of consideration in how to expose to QML because of how we define them in C++.

uitools/examples/cpp_quick/src/demos/TimeSliderDemoModel.cpp
anmacdonald: Are you sure about that one? I'm not seeing the problem in git, and they have the newlines in QtCreator.

uitools/examples/cpp_quick/src/demos/OverviewMapDemo.qml
anmacdonald: I'm happy to go through and change the suffixes about. I'm not convinced they are _not_ models but I understand the concern. 

All the QML files could have the name `Form` appended, and `Model` could be removed from the C++ files, this would be closer to the templates, I think.

uitools/examples/cpp_quick/src/demos/TimeSliderDemoModel.cpp
lsmallwood: maybe github hides the newlines now 🤷‍♂️

uitools/examples/cpp_quick/cpp_quick.pro
```diff
@@ -43,11 +43,11 @@ include($$PWD/../../toolkitcpp.pri)
 INCLUDEPATH += $$PWD/src
 
 HEADERS += \
-    src/BaseDemoModel.h
+    src/BaseDemol.h
```
lsmallwood, line 46: type here

anmacdonald: Ack. Litreally must have occured before commit because it tested fine. Good catch!

uitools/examples/cpp_quick/src/BaseDemo.cpp
```diff
@@ -0,0 +1,132 @@
...
+    {
+      setGeoView_(static_cast<MapQuickView*>(geoView));
+    }
+    else if (geoView->geoViewType() == GeoViewType::SceneView)
+    {
+      setGeoView_(static_cast<SceneQuickView*>(geoView));
+    }
+    else
+    {
+      throw std::runtime_error("Unknown type of GeoView");
```
JamesMBallard, line 58: 🔧  perhaps `Q_ASSERT_X`? Uncommon to throw exceptions in public-facing Runtime Qt code. Qt is exception agnostic.

anmacdonald: Good point, I've replaced with `qFatal` as that is what `Q_ASSERT` uses under the hood when a failure occurs.

PR #468, opened 2021-11-18
Awilliams/qdocs code snippets

mrwillyees:
Here I have enhanced the docs for the tools that were added into the demo apps.
I have added an image or gif to the docs and I am extracting snippets of code from the QML version of the demo app.
To make it easier I have uploaded the docs and filtered to the one of interest. How do you see it? The gifs might be too much and be annoying, if so I can change it to png (I am missing 2 images at the moment: overview and conversion).
https://esriis-my.sharepoint.com/:f:/g/personal/ale11971_esri_com/ElMZ9zdKpiFOqcF3Jq6dFZUBybxqahDr4A5IomsciVwi0g?e=EiILtj
To be noted is a new folder inside the /images called docs in which I put all the images and gifs needed to be built into the documentation.

After this, I will need to come up with a solution to also show snippets for the cpp version since it sometime need more work to be done in the cpp backscene, but I am unsure where these snippets should be put because they both point to the same qml files?


uitools/examples/qml_quick/src/demos/BasemapGalleryDemo.qml
```diff
@@ -72,5 +73,6 @@ DemoPage {
                 }
             }
         }
+        //! [Set up BasemapGallery QML]
```
anmacdonald, line 76: This snippet is QML API specific because of the `Connections` object. Because the `BasemapGallery` documentation is shared amongst the C++ and QML API this snippet may be confusing to the C++ API crowd.

We may have to reference both the C++ and QML APIs in the BasemapGallery documentation, and make it clear which one is for whom. @ldanzinger, thoughts?

anmacdonald:
I think the documentation looks really good now, thank you for doing this.

Currently there is a QML API bias in the documentation - we might need to explicitly show how to set up a QML API / C++ API in the same documentation.


Also watch out, there has been a mass file rename going on in the `Example apps branch` - `demo_apps` is now stale as a branch.

uitools/examples/qml_quick/src/demos/CalloutDemo.qml
```diff
@@ -16,6 +16,7 @@ import Esri.ArcGISRuntime 100.13
 import Esri.ArcGISRuntime.Toolkit 100.13
 
 DemoPage {
+    //! [Set up Callout QML]
```
ldanzinger, line 19: ❌  I think we could reduce the scope of this snippet. Showing the `mapViewContents: Component` section is very app specific and could read confusing in the standalone doc. I'd just move the tag around MapView and its children instead.

uitools/examples/qml_quick/src/demos/CoordinateConversionDemo.qml
```diff
@@ -31,7 +31,7 @@ DemoPage {
             }
         }
     }
-
+//! [Set up Coordinate Conversion QML]
```
ldanzinger, line 34: ❌  Same comment here as Callout - better to leave out the `mapViewContents: Component` and only show the MapView and its children

uitools/examples/qml_quick/src/demos/BasemapGalleryDemo.qml
ldanzinger: Hmm, not only the Connections object, but the declaration of the Map object as well ties this to the QML API... This is an issue in pretty much every snippet here... Maybe we just need some standard text before each code snippet that says something like "This snippet shows how to XYZ using the QML API. The workflow for using the Toolkit with the C++ API is similar, but the ArcGIS Runtime objects will be created in C++ code instead."

Not sure what other options we have. I don't think I'm a fan of showing QML, C++ Quick, and Widget snippets in each class doc - that is likely going to be overkill...

mrwillyees: I see, so then it might be better to show the C++ API which is more complicated (at least I found it that way when writing the demos), but it might be a bit more messy since we have to set up different snippets also on the C++ side. A good point would be to redirect the users to the sample pages from our repository.

mrwillyees:
If it is fine how I set up the project by storing the resources under `/uitools/images/docs`, I will then open a PR for the toolkit build project because I need to add environment variables to set it up correctly. @ldanzinger

uitools/examples/qml_quick/src/demos/BasemapGalleryDemo.qml
anmacdonald: @ldanzinger It would be the Quick C++ API and the Quick QML API in the same class doc. There wouldn't be a widgets snippet. I don't know if that changes your feelings?

PR #469, opened 2021-11-22
Awilliams/qdocs code snippets vnext

mrwillyees:
cherrypicked commits from the  branch `qdocs_code_snippets` which was initially based on the branch `demo-apps`.
It is spawned from the current `v.next`, so all the newly introduced changes from `example-apps` are present, also the mass rename of files.
Doc project was tested and successfully created the docs.


uitools/examples/qml_quick/src/demos/AuthenticationViewDemoForm.qml
```diff
@@ -16,6 +16,7 @@ import Esri.ArcGISRuntime 100.13
 import Esri.ArcGISRuntime.Toolkit 100.13
 
 DemoPage {
+    //! [Set up Authentication View QML]
```
anmacdonald, line 19: ❌   I agree with Lucas we shouldn't expose any of the `mapViewContents: Component {` code. It confuses the story in how to set up an AuthenticationView.

Would it be fair to guess you've done this because you want to expose the URL? I don't think that is necessary for the url to be in the example docs. Perhaps a good middle ground is to rename `url` to something more descriptive like `exampleUrl`, then the snippet can be displayed like this:

```qml
//! [Set up Authentication View QML
MapView {
    Map {
        initUrl: exampleUrl
    }
    AuthenticationView {
        anchors.centerIn: parent
    }
}
//! [Set up Authentication View QML]
```

uitools/examples/qml_quick/src/demos/CalloutDemoForm.qml
```diff
@@ -51,5 +52,6 @@ DemoPage {
                 }
             }
         }
+//! [Set up Callout QML]
```
anmacdonald, line 55: 🙃 I'm seing a mix of indentaion in this PR. I don't mind which way round it goes, but please choose an indentation for the snippt lines which can be applied to all examples.

uitools/import/Esri/ArcGISRuntime/Toolkit/AuthenticationView.qml
```diff
@@ -47,6 +47,10 @@ import QtQuick.Controls 2.12
         \c{QtWebView::initialize()} immediately after the \c{QGuiApplication} 
         instance is created. See \l{https://doc.qt.io/qt-5/qtwebview-index.html}
         {Qt WebView}.
+
+  \image docs/authenticationview_qml.png authentication_view
+  Example code:
+  \snippet qml_quick/src/demos/AuthenticationViewDemoForm.qml Set up Authentication View QML
```
anmacdonald, line 53: One thing @ldanzinger  asked for was some more explicit text to draw attention to the fact that this example code is QML only. 

E.g.

```
Example of usage in th QML API (C++ API may differ):
```

anmacdonald (changes requested):
👍 Looking very good. Some minor cosmetic changes.

uitools/examples/qml_quick/src/demos/AuthenticationViewDemoForm.qml
mrwillyees: Yes I wanted to expose the url, but I like the idea to imply that they will need a url by using `exampleUrl`

ldanzinger:
@mrwillyees @anmacdonald not sure if it helps, but note there are several more advanced qdoc options for pulling in specific parts of files, leaving out others, adding dots, etc - https://doc.qt.io/qt-5/07-0-qdoc-commands-includingexternalcode.html

mrwillyees:
> @mrwillyees @anmacdonald not sure if it helps, but note there are several more advanced qdoc options for pulling in specific parts of files, leaving out others, adding dots, etc - https://doc.qt.io/qt-5/07-0-qdoc-commands-includingexternalcode.html

Thanks, and in fact that doc link is the one I used to set up the \snippet. I chose this functionality because it seemed the most flexible and less required maintenance over changes in the demo app.

anmacdonald (approved):
LGTM

PR #470, opened 2021-11-24
Example apps can now read an ENV variable to set the API key.

anmacdonald:
As discussed, added a configuration to read the API Key from the environment if not explicitly set.

Ran the formatter which edited the file moree than I expected. Sorry about that.


mrwillyees (commented):
Does it still request for the api in case the environment variable is not set?

anmacdonald:
Yeah in that case it will be a blank QByteArray converted to a blank QString.

https://doc.qt.io/qt-5/qtglobal.html#qgetenv

mrwillyees (approved):
LGTM

PR #471, opened 2021-11-25
Fixed the rowcount and columnspan of the popupview.

anmacdonald:
Fixes break caused by https://github.com/Esri/arcgis-runtime-toolkit-qt/pull/438/files

Effectively, the close button was removed from the content of the Popup, but the Label wasn't set to span two columns, this made everything go out of order with respect to the attachments.


mrwillyees (approved):
Ty for finding the bug

PR #472, opened 2021-11-25
Popup now favours label over fieldName if available.

anmacdonald:
While this is a behaviour change, this is probably a good quality of life improvement for the Popup.

| Old | New |
-----|------
| image | image |


ldanzinger (approved):
LGTM, thanks for this

PR #476, opened 2021-12-01
Change Qml Overview Map insetView property type

AndrewBladon:
This PR changes the insetView property type from `MapView` to `var`. This change was made in response to issues reported while testing the Qml version of the Display overview Map sample on Android.

***NOTE:*** Initial testing comprised the Qml sample viewer being built and deployed onto an android emulator - the issue was reproduced and these changes fixed the issue. However, while following these steps a second time during re-testing the original issue could not be reproduced.


anmacdonald (approved):
👍  LGTM. 

> NOTE: Initial testing comprised the Qml sample viewer being built and deployed onto an android emulator - the issue was reproduced and these changes fixed the issue. However, while following these steps a second time during re-testing the original issue could not be reproduced.

Just to clarify, are you saying the repro case stopped being a repro case at some point?

Please remember to merge `release/100.13.0` into `v.next` after this is committed.

PR #478, opened 2021-12-02
Awilliams/qdocwarnings fix

mrwillyees:
A few qdoc warning removed from the toolkit.

### DIFF
--- a/augmentedreality/Common/include/LocationDataSource.h
+++ b/augmentedreality/Common/include/LocationDataSource.h
@@ -34,7 +34,7 @@ class LocationDataSource : public QObject
   Q_PROPERTY(QGeoPositionInfoSource* geoPositionSource READ geoPositionSource
              WRITE setGeoPositionSource NOTIFY geoPositionSourceChanged)
   Q_PROPERTY(QCompass* compass READ compass WRITE setCompass NOTIFY compassChanged)
-  Q_PROPERTY(bool isStarted READ isStarted NOTIFY isStartedChanged)
+  Q_PROPERTY(bool started READ isStarted NOTIFY isStartedChanged)
   Q_PROPERTY(ArEnums::SensorStatus sensorStatus READ sensorStatus NOTIFY sensorStatusChanged)
   Q_PROPERTY(ArEnums::LocationTrackingMode locationTrackingMode READ locationTrackingMode
              WRITE setLocationTrackingMode NOTIFY locationTrackingModeChanged)
--- a/augmentedreality/Common/include/LocationDataSource.h
+++ b/augmentedreality/Common/include/LocationDataSource.h
@@ -34,7 +34,7 @@ class LocationDataSource : public QObject
   Q_PROPERTY(QGeoPositionInfoSource* geoPositionSource READ geoPositionSource
              WRITE setGeoPositionSource NOTIFY geoPositionSourceChanged)
   Q_PROPERTY(QCompass* compass READ compass WRITE setCompass NOTIFY compassChanged)
-  Q_PROPERTY(bool started READ isStarted NOTIFY isStartedChanged)
+  Q_PROPERTY(bool isStarted READ isStarted NOTIFY isStartedChanged)
   Q_PROPERTY(ArEnums::SensorStatus sensorStatus READ sensorStatus NOTIFY sensorStatusChanged)
   Q_PROPERTY(ArEnums::LocationTrackingMode locationTrackingMode READ locationTrackingMode
              WRITE setLocationTrackingMode NOTIFY locationTrackingModeChanged)
--- a/augmentedreality/Common/source/LocationDataSource.cpp
+++ b/augmentedreality/Common/source/LocationDataSource.cpp
@@ -40,7 +40,7 @@ using namespace Esri::ArcGISRuntime::Toolkit;
   are emitted every time the values change.
 
   Most of the time, it's not necessary to create an object of this class. A default LocationDataSource
-  is created automatically by \l ArcGISArSceneView when this is necessary. This class is public to
+  is created automatically by \l ArcGISArView when this is necessary. This class is public to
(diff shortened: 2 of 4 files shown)


augmentedreality/Common/include/LocationDataSource.h
```diff
@@ -34,7 +34,7 @@ class LocationDataSource : public QObject
   Q_PROPERTY(QGeoPositionInfoSource* geoPositionSource READ geoPositionSource
              WRITE setGeoPositionSource NOTIFY geoPositionSourceChanged)
   Q_PROPERTY(QCompass* compass READ compass WRITE setCompass NOTIFY compassChanged)
-  Q_PROPERTY(bool started READ isStarted NOTIFY isStartedChanged)
+  Q_PROPERTY(bool isStarted READ isStarted NOTIFY isStartedChanged)
```
mrwillyees, line 37: had to change the property because qdocs was expecting this name convention from the `isStarted()` function. I don't think there should be any problems, since the property `isStarted` is accessed via the getter. It doesn't matter the name of it

augmentedreality/Common/source/LocationDataSource.cpp
```diff
@@ -40,7 +40,7 @@ using namespace Esri::ArcGISRuntime::Toolkit;
   are emitted every time the values change.
 
   Most of the time, it's not necessary to create an object of this class. A default LocationDataSource
-  is created automatically by \l ArcGISArSceneView when this is necessary. This class is public to
+  is created automatically by \l ArcGISArView when this is necessary. This class is public to
```
mrwillyees, line 43: Not sure if this is the correct class. Had a look on it and it seems that does what the text is saying. 
`ArcGISSceneView` could not be found. Might have been deleted?

calcite-qml/Calcite/Calcite.qml
```diff
@@ -18,7 +18,8 @@ pragma Singleton
 import QtQuick 2.0
 
 /*!
- \qmltype Calctie
+ \qmltype Calcite
+ \inqmlmodule Esri.ArcGISRuntime.Toolkit
```
anmacdonald, line 22: ❌  This isn't a Toolkit module though! If anything, it's in its own Calcite module. If there's a way to blocklist the Calcite folder from being documented that would be better.

augmentedreality/Common/source/LocationDataSource.cpp
anmacdonald: ❓  I'm not too familiar with AR to say what is correct here. @GuillaumeBelz thoughts?

augmentedreality/Common/include/LocationDataSource.h
anmacdonald: ❌ This is a breaking change which changes the API interface of the QML API. It goes against the QML naming conventions.

I'm not sure what the problem is that we are trying to solve here, but since none of the properties are documented in QML, it might instead be resolved by documenting these propertites?

calcite-qml/Calcite/SliderHandle.qml
```diff
@@ -15,6 +13,9 @@ import QtQuick 2.12
  *  See the License for the specific language governing permissions and
  *  limitations under the License.
  ******************************************************************************/
+
+import QtQuick 2.12
```
anmacdonald, line 17: 👍 Good catch

anmacdonald (changes requested):

augmentedreality/Common/source/LocationDataSource.cpp
ldanzinger: The type should be `ArcGISArView` - https://github.com/Esri/arcgis-runtime-toolkit-qt/blob/main/augmentedreality/QmlApi/qml/ArcGISArView.qml

calcite-qml/Calcite/Calcite.qml
mrwillyees: Remove the qdocs commits from this file?

augmentedreality/Common/include/LocationDataSource.h
mrwillyees: ⁉"_It goes against the QML naming conventions._" The above properties have the same name as their getter.

💭I think qdoc is internally calling a linking (\l) based on the getter name. With the property started, there is an error about missing started property, which is not using isStarted (I guess qdoc is automatically document the property).

I will try to manually document the property.

anmacdonald: 💭  The second parameter of `Q_PROPERTY` is the name of the property in the QML API, all other arguments are to do with the behaviour of the property and are transparent to the QML interface (there is no way to call `isStarted()` from QML unless it is marked with `Q_INVOKABLE`).

calcite-qml/Calcite/Calcite.qml
anmacdonald: I wouldn't remove the comments, they are valuable, and may not remove the warnings.

In the `ArcGISQtToolkit.qdocconf` file which is used to build the toolkit there is an `excludedirs` property which looks like this:

```conf
# do not process platform-specific AR headers or examples
excludedirs += ../../toolkit/augmentedreality/Common/include/iOS \
               ../../toolkit/augmentedreality/Common/include/Android \
               ../../toolkit/augmentedreality/Examples
```

which can also have this added:

```conf
# do not process platform-specific AR headers or examples
excludedirs += ../../toolkit/augmentedreality/Common/include/iOS \
               ../../toolkit/augmentedreality/Common/include/Android \
               ../../toolkit/augmentedreality/Examples \
               ../../toolkit/calcite-qml
```

anmacdonald (approved):
LGTM

ldanzinger (approved):

anmacdonald:
@mrwillyees Quick remidner to put this in `v.next` also!

mrwillyees:
@anmacdonald Ah thanks! I will have to check also my other latest PR, not sure if I did to all the required ones!

PR #482, opened 2021-12-07
version update 100.13 -> 100.14

mfeigl:
Updating the version from 100.13 to 100.14
@anmacdonald @mrwillyees could you please review this PR?

### DIFF
--- a/augmentedreality/Examples/CppArExample/CppArExample.pro
+++ b/augmentedreality/Examples/CppArExample/CppArExample.pro
@@ -21,7 +21,7 @@ CONFIG += c++14
 
 TARGET = CppArExample
 
-ARCGIS_RUNTIME_VERSION = 100.13
+ARCGIS_RUNTIME_VERSION = 100.14
 include($$PWD/arcgisruntime.pri)
 
 HEADERS += \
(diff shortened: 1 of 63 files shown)


anmacdonald (changes requested):
How does the script you use find the dependencies?

The reason I ask is that some of the `Esri.ArcGISRuntime` qmlimport is fake in `cpp_examples:

https://github.com/Esri/arcgis-runtime-toolkit-qt/blob/40c58ae41cdd2ac68ccac27ab0762f47c1fd0a18/uitools/examples/cpp_quick/src/main.cpp#L64

And those registration lines also need updated.

If it's looking for a 

```c++
constexpr int VERSION_MINOR = 14;
```

Then I can update `main.cpp` to use that so it gets caught next time.

mfeigl:
The script is just a first pass to do the bulk of the conversions, and the process still has a considerable amount of manual labour to it. 

I will address the files you mentioned, are there many more different patterns which were missed?

mrwillyees (approved):
LGTM.
Does it even happen that some tools needs an older version of the Runtime?

anmacdonald:
> I will address the files you mentioned, are there many more different patterns which were missed?

@mfeigl Not that I can see.


I can't wait for Qt 6 when all the QML versioning can leave forever.

mfeigl:
> Does it even happen that some tools needs an older version of the Runtime?

I can't think where not importing the latest toolkit would be beneficial. 

Thanks for the reviews! @anmacdonald , ready for the second round.

anmacdonald (approved):
LGTM

PR #483, opened 2021-12-08
Awilliams/searchview match ios

mrwillyees:
Qt implementation missed a check before showing the "repeat search button". Implemented by keeping track of the last GeoView extent after every search commited, suggestion chosen, result chosen. 
This variable is compared with the current GeoView extent and if their diversity ratio > lastSearchArea => show repeat search button.


mrwillyees:
still a few qdebugs around. They have been useful for debugging because the signals I was working on are called a lot of times.

Additionally, found may be a bug/enhancement. If the search is performed programmatically, using the `.commitSearch` function, the repeat search button will never be shown because there is a state that is only set based on a click done on the UI [here](https://github.com/Esri/arcgis-runtime-toolkit-qt/blob/awilliams/searchview_match_ios/uitools/import/Esri/ArcGISRuntime/Toolkit/SearchView.qml#L213). If this is a possible enhancement I can have a look at it.

uitools/cpp/Esri/ArcGISRuntime/Toolkit/SearchViewController.cpp
```diff
@@ -188,7 +212,13 @@ namespace Toolkit {
 
           if (sceneView->isNavigating())
           {
-            setIsEligableForRequery(true);
+            auto widthDiff = queryArea().extent().width() - lastSearchArea().extent().width();
+            auto heightDiff = queryArea().extent().width() - lastSearchArea().extent().width();
+
+            auto widthThreshold = lastSearchArea().extent().width() * 0.25;
+            auto heightThreshold = lastSearchArea().extent().height() * 0.25;
+            if (widthDiff > widthThreshold || heightDiff > heightThreshold)
```
anmacdonald, line 220: ❓  The `sceneView` section is missing you `Check center difference` checks as seen in `MapView`. Is this intended?

🔧  There's enough code repetition here it would be worth centralising the logic instead of copy+pasting. Will help future maintainers.

uitools/cpp/Esri/ArcGISRuntime/Toolkit/SearchViewController.cpp
```diff
@@ -423,13 +463,28 @@ namespace Toolkit {
...
+      auto connection = std::make_shared<QMetaObject::Connection>();
       if (auto sceneView = qobject_cast<SceneViewToolkit*>(m_geoView))
       {
+        // When the geoview changes, update the lastsearcharea
+        *connection = connect(sceneView, &SceneViewToolkit::viewpointChanged, this, [sceneView, this, connection]()
+                              {
+                                disconnect(*connection);
+                                auto extent = sceneView->currentViewpoint(ViewpointType::BoundingGeometry).targetGeometry().extent();
+                                setLastSearchArea(extent);
+                              });
```
anmacdonald, line 475: 💭 🔧  As I understand it - these connections are here because we need to find the difference between the currently progmatically "set" viewpoint, and user navigation away from that set viewpoint.

I think there's some room for simplification here. 

In the `setViewpoint` lambda at line 161, you could test when  `isNavigating() == false`, and then update `m_lastSearchArea`. This means `m_lastSearchArea` is updated to whatever the viewpoint is changed ONLY when the user is NOT navigating.

If this works, it would remove the 4 singleshot connections that are currently used.

uitools/cpp/Esri/ArcGISRuntime/Toolkit/SearchViewController.cpp
```diff
@@ -168,7 +172,24 @@ namespace Toolkit {
...
+            double heightThreshold = lastSearchArea().extent().height() * m_thresholdRatioRepeatSearch;
+            if (widthDiff > widthThreshold || heightDiff > heightThreshold)
+            {
+              setIsEligableForRequery(true);
+            }
+            // Check center difference.
+            double centerDiff = ArcGISRuntime::GeometryEngine::distance(lastSearchArea().extent().center(), queryArea().extent().center());
+            double currentExtentAvg = (lastSearchArea().extent().width() + lastSearchArea().extent().height()) / 2;
+            double threshold = currentExtentAvg * m_thresholdRatioRepeatSearch;
+            if (centerDiff > threshold)
```
anmacdonald, line 189: ❓ Do we need this check along with the extent test? Seems redundant after the extent calculations have already been done.

anmacdonald (changes requested):
This is some good work. Happy we are closer to iOS. 

Some fixes in the comments.

uitools/cpp/Esri/ArcGISRuntime/Toolkit/SearchViewController.h
```diff
@@ -88,6 +89,9 @@ namespace Toolkit {
     Geometry queryArea() const;
     void setQueryArea(Geometry queryArea);
 
+    Geometry lastSearchArea();
+    void setLastSearchArea(Geometry);
```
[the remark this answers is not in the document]
mrwillyees: For future reference: all the public API should exactly match the YAML tool design. This is not there, so is private, or do we have a bit of flexibility on our part?
🔧I agree in making it private!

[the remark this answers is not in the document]
anmacdonald: I'd gun for flexibility on our part!

uitools/cpp/Esri/ArcGISRuntime/Toolkit/SearchViewController.cpp
mrwillyees: it seems so, because the extent one will only check how wide or high (in km for example) is currently shown on the screen. Will trigger an update only for zooming out and in (changes the km being shown on the screen).
The center check is to check if the user pans around but keeping the same zoom

uitools/cpp/Esri/ArcGISRuntime/Toolkit/SearchViewController.cpp
```diff
@@ -168,7 +172,24 @@ namespace Toolkit {
...
+            // Check extent difference.
+            double widthDiff = abs(queryArea().extent().width() - lastSearchArea().extent().width());
+            double heightDiff = abs(queryArea().extent().width() - lastSearchArea().extent().width());
+
+            double widthThreshold = lastSearchArea().extent().width() * m_thresholdRatioRepeatSearch;
+            double heightThreshold = lastSearchArea().extent().height() * m_thresholdRatioRepeatSearch;
+            if (widthDiff > widthThreshold || heightDiff > heightThreshold)
+            {
+              setIsEligableForRequery(true);
+            }
```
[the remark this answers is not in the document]
mrwillyees: I tested and it seems to be working differently from the IOS version. I might have given the variable `m_thresholdRatioRepeatSearch` a misleading name. In practice, they check if either the height or the width has been increased/reduced by a 25% by comparing the current and old height/width. The `m_thresholdRatioRepeatSearch` is just a percentage of increase.

uitools/cpp/Esri/ArcGISRuntime/Toolkit/SearchViewController.cpp
anmacdonald: Ah that makes sense. If you could put in seperate functions or give them comments denoting that one is a pan test and one is a zoom test that'd be a great help in readibility.

uitools/cpp/Esri/ArcGISRuntime/Toolkit/SearchViewController.cpp
mrwillyees: Using your idea to only set the lastsearcharea after the viewpoint has changed (waiting for geoview signal) and that the geoview->isnotnavigating.
It does seem working, but after some debugging I think there might be some cases in which it will not work. For example by setting a qdebug() every time that the lastsearcharea is set, I can see that it is set after a search, suggestion selected (viewpoint moved), but also when the user pans or zooms. If the user will do a small pan, the lastsearcharea is set, invalidating the viewpoint that should only be set AFTER a search, not a pan or zoom. demo showing it:
https://user-images.githubusercontent.com/91129623/146052257-f49e7bdd-5865-4296-bdff-12715a852ec1.mp4

anmacdonald: Good investigation. I can also see this bug. The zoom behaviour seems particularly egregious. This is not how I'd expect the API to behave. Let's open a ticket in the Qt repo, we might also be able to bugfix ourselves also.

anmacdonald: 🔧 

Overall it sounds like we need an API improvement in this area. As written `isNavigating` gives too many false positives.

Your initial solution is good enough, but what I would like to see is some boilerplate removed.

Within the `internal` folder can you make a new C++ header file, (something like `singleShotConnection.h`)

You can make a helper for singleshot by doing something like this:

```c++
namespace Esri {
namespace ArcGISRuntime {
namespace Toolkit {

template <typename Sender, typename Signal, typename Receiver, typename Slot>
QMetaObject::Connection singleShotConnection(Sender* sender, Signal&& signal, Receiver* reciever, Slot&& slot)
{
  auto connection = std::make_shared<QMetaObject::Connection>();
  *connection = QObject::connect(sender, std::forward<Signal>(signal),
                                 reciever, [connection, slot = std::forward<Slot>(slot)](auto&&...args)
                                 {
                                   QObject::disconnect(*connection);
                                   slot(args...);
                                 });
  return *connection;
}
```

And update all single-shot connections in the codebase. E.g.:

```cpp
singleShotConnection(sceneView, &SceneViewToolkit::viewpointChanged, this, [sceneView, this]()
{
  auto extent = sceneView->currentViewpoint(ViewpointType::BoundingGeometry).targetGeometry().extent();
  setLastSearchArea(extent);
});
```

I think that will go a long way towards maintainability and readability.

uitools/cpp/Esri/ArcGISRuntime/Toolkit/SearchViewController.h
```diff
@@ -126,12 +130,16 @@ namespace Toolkit {
...
     void isAutomaticConfigurationEnabledChanged();
+    void thresholdRatioRepeatSearchChanged();
 
   private:
     void onQueryChanged();
     void onSourcesAdded(const QModelIndex& parent, int firstSource, int lastSource);
     void onSourcesRemoved(const QModelIndex& parent, int firstSource, int lastSource);
 
+    bool checkPanningDifferenceLastSearch(Geometry& geom);
+    bool checkZoomingDifferenceLastSearch(Geometry& geom);
```
anmacdonald, line 141: 🔧  Good stuff making these! Couple of points.

* Arguments should be `const Geometry&` not `Geometry&`.
* AFAIK these don't require the scope of `this` so could exist entirely within the anonymous namespace of the C++ file.
* OR if you do need to keep these as class methods, please mark the methods as `const`.

mrwillyees: The way I structured is that they depend on two internal variables (`threshold` and the `lastsearcharea`), so I prefer to keep them in the class scope. Adding the missing const.

uitools/cpp/Esri/ArcGISRuntime/Toolkit/SearchViewController.cpp
```diff
@@ -510,6 +517,8 @@ namespace Toolkit {
       auto source = m_sources->element<SearchSourceInterface>(m_sources->index((i)));
       if (source)
       {
+        auto connection = std::make_shared<QMetaObject::Connection>();
+        //set the lastSearchArea after the geoView viewpoint has changed and only once.
```
anmacdonald, line 521: ❌  Delete this.

uitools/cpp/Esri/ArcGISRuntime/Toolkit/SearchViewController.cpp
```diff
@@ -618,6 +641,26 @@ namespace Toolkit {
...
+                  }
+                  else if (auto sceneView = qobject_cast<SceneViewToolkit*>(m_geoView))
+                  {
+                    *connection = connect(sceneView, &SceneViewToolkit::viewpointChanged, this, [sceneView, connection, this]()
+                                          {
+                                            disconnect(*connection);
+                                            auto extent = sceneView->currentViewpoint(ViewpointType::BoundingGeometry).targetGeometry().extent();
+                                            m_lastSearchArea = extent;
+                                          });
+                  }
```
anmacdonald, line 663: 🔧  Make these `singleShot`

uitools/examples/shared/qml/tools.js
```diff
@@ -9,6 +9,10 @@
 //
 // See the Sample code usage restrictions document for further information.
 const TOOLS = [
+                {
+                  "name": "Search View",
+                  "url": "SearchViewDemoForm.qml"
+                },
```
anmacdonald, line 15: 🙃 Indentation issue here.

anmacdonald (changes requested):
Looking very good. Just a couple of final cosmetic tweaks!

uitools/cpp/Esri/ArcGISRuntime/Toolkit/SearchViewController.cpp
mrwillyees: I already made them in my other PR. I think you are right in making them `singleshot` here!

anmacdonald (approved):
LGTM!

PR #487, opened 2021-12-18
Callout update

anmacdonald:
image

* This update changes the callout tail rendering from `Canvas` to `Shape`. 
* `Callout` is now a Pane, rather than an `Item`, and so now follows the sizing and styling behaviour of a Qt Control. This has simplified quite a bit of the sizing logic around the contents, and also has standardized the behaviours to be more in line with the QML ecosystem (for example, changing border width will no longer change Geometry size - it is a visual property only.)
* `leaderPosition` defaulted to `Bottom`, now defaults to `Automatic`.
* Button images have been replaced with `svgs` (was `pngs`). This has removed some of the pixelation around those buttons.
* I've ensured the public API is still compatible with the old callout. Several properties have been marked as `\obsolete` in this regard. But no breaking changes to the public API have been introduced. 
    * If/when a breaking change is allowed, I'd like to remove the obsolete properties, and to combine the `calloutWidth`/`maxWidth` properties into one property. In an ideal world `autoAdjustWidth: true` would cause the callout width to expand up to `maxWidth`, while `autoAdjustWidth: false` would cause the callout to be fixed to the size of `maxWidth`. Currently we have `calloutWidth` and `maxWidth` properties - one of these values is always ignored depending on the state of the `autoAdjustWidth` flag.
    *  We _could_ alias `calloutWidth` to match `maxWidth` right now, but this will require an update to our current test framework, and may therefore be too much of a behaviour break. @JamesMBallard thoughts?

Reviewing note: this is pretty much a rewrite - so a diff of the `Callout` file will not be very effective. It would be best to review as an entirely new file, but I leave this to your discretion. 

@ldanzinger please review the docs
@JamesMBallard @mrwillyees Please review the code.


uitools/import/Esri/ArcGISRuntime/Toolkit/Callout.qml
```diff
@@ -209,143 +161,152 @@ Item {
...
+    onCalloutDataChanged: {
+        if (calloutData === null) {
+            dismiss();
+        } else if (calloutData.visible) {
+            showCallout();
+        }
+    }
+
+    /*!
+      When \c true, the width of the callout content automatically resizes up
```
ldanzinger, line 173: ❓ Not certain, but I believe doc'ing in QML requires you to specify the \brief command (this was the case when we initially wrote the doc for this). Can you confirm please?

uitools/import/Esri/ArcGISRuntime/Toolkit/Callout.qml
```diff
@@ -209,143 +161,152 @@ Item {
...
-        target: calloutData
-        function onScreenPointChanged() {
-            anchorPointx = calloutData.screenPoint.x;
-            anchorPointy = calloutData.screenPoint.y;
+    /*!
+        \obsolete
+        Use \c{background.border.width} instead.
+    */
+    property int borderWidth
+    onBorderWidthChanged: background.border.width = borderWidth;
```
ldanzinger, line 234: ❓ can you use binding instead of assignment? There are several of these instances

uitools/import/Esri/ArcGISRuntime/Toolkit/Callout.qml
anmacdonald: I'm not seeing any difference in the documented output, nor am I seeing any warnings. Perhaps this requirement has been relaxed? 

Either way, this is a good catch and there's no harm in adding them back so I'll do so.

JamesMBallard:
>We could alias calloutWidth to match maxWidth right now, but this will require an update to our current test framework, and may therefore be too much of a behaviour break. @JamesMBallard thoughts?

Please elaborate. If it cleans up the code, and has no fallout for users/apps, I'm certainly open to this. I don't understand the implications on our test framework.

uitools/import/Esri/ArcGISRuntime/Toolkit/Callout.qml
anmacdonald: This reads as a one-way binding, where any changes to the obsolete property will change the non-obsolete property (i.e. changes to `borderWidth` will be reflected in `background.border.width`), but the same can not be said the other way round (i.e. changes to `background.border.width` will not be reflected in `borderWidth`).

For the sake of clarity: `property int borderWidth: background.border.width` will not work by itself, as the binding only exists for as long as the user doesn't set a custom value of `borderWidth`. As soon as the user gives `borderWidth` a custom value the binding disappears.

In an ideal world, `borderWidth` would be an alias of `background.border.width`, as alias acts like a two way binding, but `alias` is strict in what you can apply it to, and `background.border.width` is an example of a property which can _not_ be aliased.

Achieving a two-way binding is possible, just more verbose than I would like, effectivey every property unrolls into:

```qml
    /*!
        \obsolete
        Use \c{background.border.width} instead.
    */
    property int borderWidth: background.border.width
    Binding on borderWidth {
        value: background.border.width

    }
    Binding {
        target: background.border
        property: "width"
        value: borderWidth
    }

    /*!
        \obsolete
        Use \c{background.color} instead.
    */
    property color backgroundColor: background.color
    Binding on backgroundColor {
        value: background.color

    }
    Binding {
        target: background
        property: "color"
        value: backgroundColor
    }

```

I'm happy to update if two-way binding behaviour is desired. I had hoped that a user would use one property or the other, not both, but I can understand the desire to be foolproof with the obsolete properties.

ldanzinger: I'm ok with what you have.

JamesMBallard:
@anmacdonald can you (or did you already) investigate any implications this change will have when we eventually move towards Qt 6 and RHI?

https://doc.qt.io/qt-5/qml-qtquick-shapes-shape.html#details
> * The Direct 3D 12 backend is not currently supported.
> * When OpenGL is not used directly by the scene graph, for example because it is using the graphics abstraction layer (QRhi), only the generic shape renderer is available.
> * When the application is running with the generic, triangulation-based Shape implementation, the geometry generation happens entirely on the CPU. This is potentially expensive

After a quick glance, those all look about the same or somewhat more accommodating than the Canvas restrictions at Qt 5, but the Qt 6 canvas page drops all mentions of OpenGL
https://doc.qt.io/qt-6/qml-qtquick-canvas.html

While Shapes still seems tightly coupled with OpenGL at Qt 6:
https://doc.qt.io/qt-6/qml-qtquick-shapes-shape.html#rendererType-prop

This question should've come up sooner, I apologize for that. We need to make sure we not making our lives harder for the Qt 6 transition, and it looks like there are some things to consider here.

ldanzinger (approved):
doc looks good

anmacdonald:
@JamesMBallard

The test in question is the `"Show Callout 4"` test. In the test `maxWidth` is set to `1000` and `autoAdjustWidth` is set to `false`. So, in this test, the value of `maxWidth: 1000` is expected to be ignored, and the width will instead be `calloutWidth: 300`.

With the change I am proposing, `calloutWidth` would be an alias of `maxWidth` and so in the `Callout 4` test the width will be `1000`.

I doubt that in a "real world" application the developer would want to set a `maxWidth` value and ignore it, but that would be the change of behaviour. To summarise:

|                                                                                    | Old behaviour | Proposed behaviour |
|---------------------------------------------------|----------------|---------------------|
| `autoAdjustWidth: true` and `calloutWidth` updated   | no change. `maxWidth` is used for maximum width calculations. | Maximum width is `calloutWidth`. |
| `autoAdjustWidth: true` and `maxWidth` updated   | Maximum width is `maxWidth`. | Maximum width is `maxWidth`.  |
| `autoAdjustWidth: false` and `calloutWidth` updated   | Fixed width is `calloutWidth`. |  Fixed width is `calloutWidth`.   |
| `autoAdjustWidth: false` and `maxWidth` updated   | no change. `calloutWidth` is used for fixed width calculations. |  Fixed wWidth is `maxWidth`.   |

anmacdonald:
@JamesMBallard 

> can you (or did you already) investigate any implications this change will have when we eventually move towards Qt 6 and RHI?

Good question, I hadn't looked into it before!

| Qt 5 | Qt 6 |
|------|-----|
| image | image |

The RHI seciton reads as:

> When running with the OpenGL backend of Qt Quick, only the generic, triangulation-based approach is available. When OpenGL is not used directly by the scene graph, for example because it is using the graphics abstraction layer (QRhi), only the generic shape renderer is available.

So it  is supported, but it sounds like shape rendering is on the CPU. Considering we are only drawing one triangle which is cached, I'm not too worried about the performance implications of this.

uitools/import/Esri/ArcGISRuntime/Toolkit/Callout.qml
```diff
@@ -404,692 +333,349 @@ Item {
...
+            onClicked: accessoryButtonClicked()
+            icon.source: {
+                if (accessoryButtonType === "Info")
+                    return "images/information.svg";
+                else if (accessoryButtonType === "Add")
+                    return "images/plus-circle.svg";
+                else if (accessoryButtonType === "Custom")
+                    return customImageUrl;
+                else
+                    return "";
```
JamesMBallard, line 426: 🙃 

Personal preference for readability.

```suggestion

                return "";
```

uitools/import/Esri/ArcGISRuntime/Toolkit/Callout.qml
```diff
@@ -404,692 +333,349 @@ Item {
...
-        // If max width has been specified and fits in MapView, just return that value
-        var styleMaxWidthExclExtras = calloutWidth - widthOfExtras;
-        if (styleMaxWidthExclExtras > 0 && styleMaxWidthExclExtras < maxWidthForMapView) {
-            return styleMaxWidthExclExtras;
+        // Calculate the portion of property `width` allocated to the label `this`.
+        // Called via `labelWidthFrom.bind(label)(width)`.
+        function labelWidthFrom(width) {
+            if (!this.text)
+                return 0;
+            return width - (image.visible ? image.width : 0 ) - (accessoryButton.visible ? accessoryButton.width : 0);
```
JamesMBallard, line 678: 🙃 
```suggestion
                return 0;

            return width - (image.visible ? image.width : 0 ) - (accessoryButton.visible ? accessoryButton.width : 0);
```

uitools/import/Esri/ArcGISRuntime/Toolkit/Callout.qml
```diff
@@ -404,692 +333,349 @@ Item {
...
-                        id: title
-                        text: calloutData ? calloutData.title : ""
-                        wrapMode: Text.NoWrap
-                        renderType: Text.NativeRendering
-                        color: titleTextColor
-                        font {
-                            pixelSize: 11
-                            family: "sanserif"
+                    relativeY: {
+                        switch(internal.leaderPosition) {
```
JamesMBallard, line 503: ❓ None of the switches have default cases. Should they?

uitools/import/Esri/ArcGISRuntime/Toolkit/Callout.qml
```diff
@@ -404,692 +333,349 @@ Item {
...
+            Layout.fillHeight: true
+            Layout.preferredWidth: autoAdjustWidth ? -1 : internal.labelWidthFrom.bind(this)(root.calloutWidth)
+            Layout.maximumWidth: autoAdjustWidth ? internal.labelWidthFrom.bind(this)(root.maxWidth) : -1
+            Layout.columnSpan: {
+                let span = 1;
+                if (!accessoryButton.visible)
+                    span++;
+                if (!image.visible)
+                    span++;
+                return span;
```
JamesMBallard, line 388: 🙃 
```suggestion
                    span++;

                return span;
```

uitools/import/Esri/ArcGISRuntime/Toolkit/Callout.qml
```diff
@@ -404,692 +333,349 @@ Item {
...
-            // handler to override for drawing
-            onPaint: {
-                screenCoordinates.x = borderWidth;
-                screenCoordinates.y = borderWidth;
-                drawCalloutFrame();
+            Layout.rowSpan: {
+                let span = 1;
+                if (!detail.visible)
+                    span++;
+                return span;
```
JamesMBallard, line 394: 🙃 
```suggestion
                    span++;

                return span;
```

There are a few of these. I won't comment on them all but you get the idea. Optional throughout.

JamesMBallard:
Thanks @anmacdonald, Qt6 and RHI story sounds good. Thanks for checking.

>but that would be the change of behaviour...

In this case, it sounds fine to me. I am assuming most developers will be using our toolkit code out of the box with no changes, so aren't concerned with the internals. As long as both properties are still there so no old code would outright break, I'd rather make that change now and update that toolkit test with the new expectations.

@ldanzinger any concerns with this? I don't think the AppStudio team would face any fallout from this, but I'm not 100% sure.

>this is pretty much a rewrite - so a diff of the Callout file will not be very effective. It would be best to review as an entirely new file

Good suggestion. Even the split view diff was difficult.

JamesMBallard (approved):
All good with the RHI and test explanations.

ldanzinger:
No concerns with me. Neil and I have discussed that it will look different out of the box, but have confirmed there are some easy properties you can set to make it look like the old callout. Thinking about it now, it would be good to add a note to the class level doc that says this tool went under major refactor for version 100.14 so that it can support QML styling, and as such the default look is different. Mentioning this, along with the properties that you need to set to make it look the same, should be sufficient for anyone concerned with the differing appearance

anmacdonald:
@ldanzinger Good point, I've updated the docs to reflect this!

uitools/import/Esri/ArcGISRuntime/Toolkit/Callout.qml
anmacdonald: Good point, I've jiggled things around so `default` always matches the `LeaderPosition.Bottom` case.

anmacdonald:
> In this case, it sounds fine to me. 

@JamesMBallard Great, I've made the tweaks and retested, everything seems kosher.

uitools/import/Esri/ArcGISRuntime/Toolkit/Callout.qml
```diff
@@ -42,111 +42,79 @@ import "LeaderPosition.js" as Enums
...
+
+     \note That the Callout has gone through a major revision as of ArcGISRuntime 100.14.
+     Part of this revision has been a change to the styling behaviour of the Callout, making the Callout
+     compliant with your currently applied theme. To revert to the classic Callout look, you can supply the
+     old style properties to the Callout as provided below.
+     \oldcode
+            Callout {
+              calloutData: myCalloutData
+            }
+     \newcode
```
ldanzinger, line 54: nice! I've seen this in the qdoc documentation and wanted to use it somewhere 👍

ldanzinger (approved):
Doc LGTM

JamesMBallard (approved):

mrwillyees (approved):

PR #491, opened 2021-12-23
Awilliams/oneshot internal

mrwillyees:
Using the new singleshot helper function to automatically keep track of connection created and delete it once it has been called. 
Missing docs at the moment. Shall I create a `.cpp` to add the docs?


anmacdonald (approved):
LGTM thanks for doing this I see this as a good quality of life improvement.

PR #494, opened 2022-01-04
Awilliams/spinbox

mrwillyees:
The commented out parts are hints to different ways it was implemented before. Will remove them once the first round of PR is done.
Other than that, I see a difference between the down arrow and up arrow. It must because of the transformation I apply, will investigate if we decide it looks good enough to keep it

### DIFF
--- a/calcite-qml/Calcite/SpinBox.qml
+++ b/calcite-qml/Calcite/SpinBox.qml
@@ -64,7 +64,6 @@ T.SpinBox {
         height: parent.height
         width: height
         Rectangle {
-            //y: parent.y + parent.height - height - control.bottomPadding / 2
             width: parent.width
             height: parent.width
             color: enabled ? (up.hovered ? Calcite.foreground2 : Calcite.foreground1) : Calcite.background
@@ -137,8 +136,7 @@ T.SpinBox {
             height: parent.height
             color: enabled ? Calcite.foreground1 : Calcite.background
             border {
-                //not sure when an error occurs in a spinbox, acceptable input cannot be used
-                //color: !control.acceptableInput ? (control.activeFocus ? Calcite.dangerHover : Calcite.danger) : (control.activeFocus ? Calcite.brand : Calcite.borderInput)
+                // error box not shown. No error shown in the calcite interactive design.
                 color: control.activeFocus ? Calcite.brand : Calcite.borderInput
             }
         }
--- a/calcite-qml/Calcite/SpinBox.qml
+++ b/calcite-qml/Calcite/SpinBox.qml
@@ -48,7 +48,8 @@ T.SpinBox {
         text: control.displayText
 
         font: control.font
-        color: enabled ? Calcite.text2 : Calcite.text2
+        //color: enabled ? Calcite.text3 : Calcite.text2
+        color : Calcite.text3
         selectionColor: Calcite.brandHover
         selectedTextColor: Calcite.foreground1
         horizontalAlignment: Qt.AlignHCenter
@@ -68,13 +69,15 @@ T.SpinBox {
         height: parent.height
         width: height
         Rectangle {
-            y: parent.y + parent.height - height - control.bottomPadding / 2
+            //y: parent.y + parent.height - height - control.bottomPadding / 2
             width: parent.width
-            height: parent.width
+            height: parent.width - bottomPadding * 2
+            y : bottomPadding
+            color: enabled ? Calcite.foreground1 : Calcite.background
             border {
-                color: "black"
+                color: Calcite.border1
             }
-        }
+
 //        Ripple {
 //            clipRadius: 2
 //            x: control.spacing
@@ -86,24 +89,25 @@ T.SpinBox {
 //            color: control.Material.rippleColor
 //        }
 
-        Rectangle {
-            x: (parent.width + width) / 2
-            y: (parent.height - height) / 2
-            width: Math.min(parent.width / 3, parent.height / 3)
-            height: 2
-            color: enabled ? Calcite.text1: Calcite.foreground2
-            transform: Rotation {origin.x: 0; origin.y: 1; angle: 135;}
+            Rectangle {
+                x: (parent.width + width) / 2
+                y: (parent.height - height) / 2
+                width: Math.min(parent.width / 3, parent.height / 3)
+                height: 2
+                //color: enabled ? Calcite.text3: Calcite.foreground2
+                color: Calcite.text3
+                transform: Rotation {origin.x: 0; origin.y: 1; angle: 135;}
+            }
+            Rectangle {
+                x: (parent.width + width) / 2
+                y: (parent.height - height) / 2
+                width: Math.min(parent.width / 3, parent.height / 3)
+                height: 2
+                color: Calcite.text3
+                transform: Rotation {origin.x: 0; origin.y: 1; angle: 225;}
+            }
         }
-        Rectangle {
-            x: (parent.width + width) / 2
-            y: (parent.height - height) / 2
-            width: Math.min(parent.width / 3, parent.height / 3)
-            height: 2
-            color: enabled ? Calcite.text1: Calcite.foreground2
-            transform: Rotation {origin.x: 0; origin.y: 1; angle: 225;}
-        }
-
-        }
+    }
 
 
     down.indicator: Item {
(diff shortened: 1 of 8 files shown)


calcite-qml/Calcite/SpinBox.qml
```diff
@@ -0,0 +1,188 @@
...
+                transform: Rotation {origin.x: 0; origin.y: 1; angle: 45;}
+            }
+            Rectangle {
+                x: (parent.width - width) / 2
+                y: (parent.height - height) / 2
+                width: Math.min(parent.width / 3, parent.height / 3)
+                height: 2
+                color: down.hovered ? Calcite.text1 : Calcite.text3
+                transform: Rotation {origin.x: 0; origin.y: 1; angle: -45;}
+            }
```
anmacdonald, line 146: 💭   Instead of building this from rectangles, take a look at importing the icons at:  https://esri.github.io/calcite-ui-icons/#? 

In particular `chevron-left.svg` and `chevron-right.svg`

There is one niggle with using images, though: currently, [ColorOverlay](https://doc.qt.io/qt-6/qml-qt5compat-graphicaleffects-coloroverlay.html) is used elsewhere in this module to set an icons color, and is being deprecated in Qt 6, leaving no clear solution to colorize an image for when we port to Qt 6.

So, I'm not outright suggesting moving to use the calcite icons, as it might end up as more work for us in the future (depending if Qt give us a descent alternative.) But the icons might provide us with more calcite-compliant chevrons in terms of look. 

You will need to make a judgement call.

calcite-qml/Calcite/SpinBox.qml
```diff
@@ -0,0 +1,188 @@
...
+        color : Calcite.text1
+        selectionColor: Calcite.brandHover
+        selectedTextColor: Calcite.foreground1
+        horizontalAlignment: Qt.AlignHCenter
+        verticalAlignment: Qt.AlignVCenter
+
+        //cursorDelegate: CursorDelegate { }
+
+        readOnly: false
+        //readOnly: !control.editable
```
anmacdonald, line 61: ❓  Why ignore the `editable` flag?

calcite-qml/Calcite/SpinBox.qml
```diff
@@ -0,0 +1,188 @@
...
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ ******************************************************************************/
+import QtQuick 2.15
+import QtQuick.Templates 2.15 as T
+
+T.SpinBox {
+    id: control
+
+    editable : true
```
anmacdonald, line 22: ❓ Is there a reason for this? I wouldn't attempt to override the business logic in the style.

calcite-qml/Calcite/SpinBox.qml
```diff
@@ -0,0 +1,188 @@
...
+        x: control.mirrored ? 0 : parent.width - width
+        implicitWidth: 40
+        implicitHeight: 40
+        height: parent.height
+        width: height
+        Rectangle {
+            //y: parent.y + parent.height - height - control.bottomPadding / 2
+            width: parent.width
+            height: parent.width - bottomPadding * 2
+            y : bottomPadding
```
anmacdonald, line 76: 🔧 By the looks of things, this only works if and when `topPadding == bottomPadding`, and will look off-center if the developer were to set different topPadding and bottomPadding independently on their SpinBox.

calcite-qml/Calcite/SpinBox.qml
```diff
@@ -0,0 +1,188 @@
...
+    down.indicator: Item {
+        x: control.mirrored ? parent.width - width : 0
+        implicitWidth: 40
+        implicitHeight: 40
+        height: parent.height
+        width: height
+
+        Rectangle {
+            width: parent.width
+            height: parent.height - bottomPadding * 2
```
anmacdonald, line 123: Same `topPadding == bottomPadding` issue.

calcite-qml/Calcite/SpinBox.qml
```diff
@@ -0,0 +1,188 @@
...
+                transform: Rotation {origin.x: 0; origin.y: 1; angle: 135;}
+            }
+            Rectangle {
+                x: (parent.width + width) / 2
+                y: (parent.height - height) / 2
+                width: Math.min(parent.width / 3, parent.height / 3)
+                height: 2
+                color: up.hovered ? Calcite.text1 : Calcite.text3
+                transform: Rotation {origin.x: 0; origin.y: 1; angle: 225;}
+            }
```
anmacdonald, line 109: 💭 I made a comment elsehwere about potentially using Calcite's icons.

calcite-qml/Calcite/SpinBox.qml
```diff
@@ -0,0 +1,188 @@
...
+//            }
+//        }
+
+        Rectangle {
+            id: backgroundRect
+            x: parent.width / 2 - width / 2
+            //y: parent.y + parent.height - height - control.bottomPadding / 2
+            y : bottomPadding
+            width: control.availableWidth
+            height: parent.height - control.bottomPadding * 2
```
anmacdonald, line 179: 🔧  More padding potential issues.

calcite-qml/demo/qml/main.qml
```diff
@@ -214,13 +214,17 @@ ApplicationWindow {
                     Layout.alignment: Qt.AlignHCenter
                 }
 
-                BusyIndicator {
-                    Layout.alignment: Qt.AlignHCenter
-                }
+//                BusyIndicator {
+//                    Layout.alignment: Qt.AlignHCenter
+//                }
```
anmacdonald, line 219: ❓ Is this being commented out because the form is now too big?

anmacdonald (changes requested):
👍 I'm pleased with how this has turned out. It's looking really good. 


Edit: Also: would you mind adding a line to the `Changelog.md` saying that `SpinBoxes` are now supported! Also in Calcite's `README.md` can you tick the `SpinBox` item.

I realize I forgot to mention updating the `Changelog.md` with your previous PR, so feel free to back-fill with your additions!

calcite-qml/Calcite/SpinBox.qml
mrwillyees: If as you say the `ColorOverlay` is being used in other parts, we will have to find a solution to it when adopting Qt 6, so it is not going to be a single instance problem here for the spinbox. I would adopt it and later on apply the common solution used for the coloring of images.

calcite-qml/demo/qml/main.qml
```diff
@@ -217,10 +217,15 @@ ApplicationWindow {
                 BusyIndicator {
                     Layout.alignment: Qt.AlignHCenter
                 }
+
                 Label {
                     text: "Loading..."
                     Layout.alignment: Qt.AlignHCenter
                 }
+                SpinBox {
+                    Layout.alignment: Qt.AlignHCenter
```
mrwillyees, line 226: The `height` property is ignored. To change the size, user will need to use the `implicitHeight`. Is that a wanted behavior? It can be fixed otherwise

calcite-qml/Calcite/SpinBox.qml
```diff
@@ -0,0 +1,146 @@
...
+
+        readOnly: !control.editable
+        validator: control.validator
+        inputMethodHints: control.inputMethodHints
+    }
+
+    up.indicator: Item {
+        x: control.mirrored ? 0 : parent.width - width
+        implicitWidth: 32
+        implicitHeight: 32
```
mrwillyees, line 63: Could there be a problem by hardcoding the 32x32? It reflects better the design sizes

calcite-qml/Calcite/SpinBox.qml
```diff
@@ -0,0 +1,146 @@
...
+
+            Image {
+                id: downIndicatorImage
+                fillMode: Image.PreserveAspectFit
+                anchors {
+                    fill: parent
+                }
+                source: "images/chevron-left.svg"
+                sourceSize.width: parent.width
+                sourceSize.height: parent.width
```
mrwillyees, line 118: I noticed a major difference in using the sourceSize to scale the `svg` image, compared to only using the `fill : parent`. Qt discourages to binding `sourceSize` for svg because it can be slow, but below is a screenshot of not using it (left arrow).
image

calcite-qml/demo/qml/main.qml
anmacdonald: `ColumnLayout` sizes and positions its content so I am not surprised `height` does not work in this context. It should work for a `SpinBox` which is anchored to its parent without a layout-based parent.

calcite-qml/Calcite/calcite.qrc
```diff
@@ -37,5 +37,8 @@
         <file>images/check.svg</file>
         <file>images/caret-double-vertical.svg</file>
         <file>images/line-solid.svg</file>
+        <file>SpinBox.qml</file>
+        <file>images/chevron-left.svg</file>
+        <file>images/chevron-right.svg</file>
```
anmacdonald, line 42: 🙃 alphabetical order.

calcite-qml/Calcite/SpinBox.qml
anmacdonald: Taking a look at the [Image.sourceSize](https://doc.qt.io/qt-5/qml-qtquick-image.html#sourceSize-prop) docs:

>  If the source is an intrinsically scalable image (eg. SVG), this property determines the size of the loaded image regardless of intrinsic size. Avoid changing this property dynamically; rendering an SVG is slow compared to an image.

I would interpret "dynamically" to mean binding sourceSize to a changing value. Here, you are binding to a value that does not change for the lifetime of the SpinBox, so I don't think the performance problems will be hit.

❓ Should it be `sourceSize.height: parent.width` or `sourceSize.height: parent.height`?

calcite-qml/demo/qml/main.qml
anmacdonald: `ColumnLayout` positions and sizes its contents, so `height` is controlled by the parent.

If you create a `SpinBox` outside of a layout I suspect you'll find `height` works then.

calcite-qml/Calcite/SpinBox.qml
anmacdonald: If it complies with the calcite design, then sounds good to me!

anmacdonald (changes requested):
👍 Good stuff. Very pedantic comments left.

Remember to remove the last of the commented code, and update `CHANGELOG.md` before merging.

calcite-qml/Calcite/SpinBox.qml
mrwillyees: should be `height`, you are correct

calcite-qml/demo/qml/main.qml
mrwillyees: It does

calcite-qml/Calcite/calcite.qrc
```diff
@@ -32,10 +32,13 @@
...
         <file>SliderHandle.qml</file>
         <file>SliderGroove.qml</file>
+        <file>SpinBox.qml</file>
         <file>Switch.qml</file>
         <file>TextField.qml</file>
         <file>images/check.svg</file>
         <file>images/caret-double-vertical.svg</file>
         <file>images/line-solid.svg</file>
+        <file>images/chevron-left.svg</file>
+        <file>images/chevron-right.svg</file>
```
mrwillyees, line 42: I saw from previous PR that you prefer these resources to be ordered alphabetically. Do you manually change the ordering? When I use QT creator to add new/existing resources, they are placed like this: at the end.

mrwillyees: Also, I renamed it by removing the size that is in the original names when pulled from calcite icons. Should we stick with doing so?

anmacdonald: I also remove the size.
I do reorder manually, it makes maintainability easier in the long run in terms of finding what goes where.

PR #498, opened 2022-01-17
Awilliams/floorfilter

mrwillyees:
floorfilter qml and cpp versions


uitools/cpp/Esri/ArcGISRuntime/Toolkit/FloorFilterController.cpp
```diff
@@ -0,0 +1,363 @@
...
+  }
+
+  void FloorFilterController::populateFacilitiesForSelectedSite()
+  {
+    m_facilities->clear();
+    auto manager = getFloorManager(m_geoView);
+    if (!manager)
+      return;
+
+    const auto allFacilites = manager->facilities();
```
mrwillyees, line 331: I think it would be slightly better to find the selected site using the ID iterating over `manager->sites`, then use the list `facilities` stored into it. This comes from the fact that there would be less sites than total facilities

uitools/import/Esri/ArcGISRuntime/Toolkit/FloorFilter.qml
```diff
@@ -0,0 +1,254 @@
...
+                        else if (internal.currentVisibileListView
+                                 === FloorFilter.VisibleListView.FACILITY) {
+                            controller.selectedFacilityId = model.modelId
+                            facilityFilterMenu.visible = false
+                            internal.isFloorFilterCollapsed = false
+                        }
+                    }
+                    background: Rectangle {
+                        border.color: "black"
+                    }
```
anmacdonald, line 200: ❌ Contradicts the style. I think this may also interfere with highlighting.

uitools/import/Esri/ArcGISRuntime/Toolkit/FloorFilter.qml
```diff
@@ -0,0 +1,254 @@
...
+                visible: true
+                Layout.columnSpan: 3
+                Layout.fillWidth: true
+                implicitHeight: contentHeight
+                implicitWidth: contentWidth
+                model: internal.currentVisibileListView
+                       === FloorFilter.VisibleListView.SITE ? controller.sites : controller.facilities
+                delegate: ItemDelegate {
+                    //width: parent.width
+                    text: '\u2022 ' + model.name
```
anmacdonald, line 184: 🔧 Consider removing `'\u2022 '

The dot only appears on the highlighted item.

Note that `ItemDelegate` has a `highlighted: bool` toggle that can be set, that puts a checkmark next to the selected item. This was based on the behaviour of selects: https://esri.github.io/calcite-web/documentation/components/#selects

uitools/import/Esri/ArcGISRuntime/Toolkit/FloorFilter.qml
```diff
@@ -0,0 +1,254 @@
...
+                // property var itemSelected: {text: ""}
+                property string itemSelected: ""
+
+                Action {
+                    id: close
+                    icon.source: "images/x.svg"
+                    onTriggered: {
+                        internal.isFloorFilterCollapsed = true
+                    }
+                }
```
anmacdonald, line 57: 🔧  Why does the `Action` exist outside of the layout?

uitools/import/Esri/ArcGISRuntime/Toolkit/FloorFilter.qml
```diff
@@ -0,0 +1,254 @@
...
+
+                ColumnLayout {
```
anmacdonald, line 59: ❓   Is this `ColumnLayout` required? I notice the other toolbar has no internal `ColumnLayout`, I assume because of the outer `ColumnLayout`.

uitools/import/Esri/ArcGISRuntime/Toolkit/FloorFilter.qml
```diff
@@ -0,0 +1,254 @@
...
+                    Repeater {
+                        id: repeater
+                        property int downItem
+                        model: controller.levels
+                        delegate: ToolButton {
+                            Layout.fillWidth: true
+                            text: model.shortName
+                            onClicked: {
+                                repeater.setItemHighlighted(
+                                            index) // _q can access Repeater without using the id?
```
anmacdonald, line 80: Consider making use of the `AbstractButton.autoExclusive` to stop multiple buttons from being pressed at once.

```qml
ToolButton {
    autoExclusive: true
    onClicked: { downItem = this; checked = true; }
    highlighted: checked // May not be required?
```

Theoretically this will remove the need for `setItemHighlighted`.

anmacdonald (changes requested):
Good progress! Not given a thorough once-over, but here's some suggestions.

uitools/import/Esri/ArcGISRuntime/Toolkit/FloorFilter.qml
anmacdonald: Edit: I think the parent layout is taking care of things. See question if `ColumnLayout` on lin 59 is required.

calcite-qml/demo/qml/main.qml
```diff
@@ -120,6 +120,7 @@ ApplicationWindow {
 
         Menu {
             id: contextMenu
+            visible: true
```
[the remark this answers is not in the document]
mrwillyees: You are right, this visible should not be default visible. Removing it

uitools/import/Esri/ArcGISRuntime/Toolkit/esri_arcgisruntime_toolkit_view.qrc
```diff
@@ -1,24 +1,24 @@
 <RCC>
-  <!-- This section contains all view components (shared by C++ and QML APIs) -->
```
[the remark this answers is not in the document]
mrwillyees: Not sure how it was removed, didn't mean to delete it.
I usually modify the `qrc` files using a different text editor because Qt Creator doesn't show the xml format

uitools/import/Esri/ArcGISRuntime/Toolkit/FloorFilter.qml
mrwillyees: Do you mean to entirely remove the dot and instead use the check mark? Or still keep it for an highlighted item?

uitools/import/Esri/ArcGISRuntime/Toolkit/esri_arcgisruntime_toolkit_view.qrc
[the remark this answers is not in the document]
anmacdonald: It's hidden but there! 
image

uitools/import/Esri/ArcGISRuntime/Toolkit/FloorFilter.qml
anmacdonald: I would suggest using the checkmark for now. And the dot can become part of the style later, or replace the checkmark in the style.

uitools/import/Esri/ArcGISRuntime/Toolkit/FloorFilter.qml
```diff
@@ -0,0 +1,254 @@
...
+                           === FloorFilter.VisibleListView.FACILITY ? sourceSize.width : 0
+                    height: internal.currentVisibileListView
+                            === FloorFilter.VisibleListView.FACILITY ? sourceSize.width : 0
+                    MouseArea {
+                        anchors.fill: parent
+                        onClicked: internal.currentVisibileListView
+                                   = FloorFilter.VisibleListView.SITE
+                    }
+                }
+            }
```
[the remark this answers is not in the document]
mrwillyees: Did that. The styling color is a bit different from JS demo, so setting the `color` property of the icon manually

uitools/import/Esri/ArcGISRuntime/Toolkit/FloorFilter.qml
mrwillyees: The line 59 `ColumnLayout` is used to contain all the `ToolsBars` and they will expand the Layout. 
The same behaviour could be acheived with an `Item` but  then is needed to be set a `width` and `height` on it so it actually takes space (xes: `width: childRect.width`). What do you prefer?

mrwillyees: The other `ToolBar` do not have any Layouts because they are only a single item, so they will not overlap.
From the docs:

_ToolBar does not provide a layout of its own, but requires you to position its contents, for instance by creating a RowLayout. If only a single item is used within the ToolBar, it will resize to fit the implicit size of its contained item. This makes it particularly suitable for use together with layouts._

Long story short: multiple item `ToolBar` needs a Layout to not overlap their items.

uitools/import/Esri/ArcGISRuntime/Toolkit/FloorFilter.qml
mrwillyees: `Action` outside of the `Layout` because since it is not a graphical component I felt having it not in a `Layout`. I can set it inside, should not change any layout positions

uitools/import/Esri/ArcGISRuntime/Toolkit/FloorFilter.qml
```diff
@@ -89,19 +90,40 @@ Control {
...
+
+                Column {
+                    anchors.left: parent.left
+                    anchors.right: parent.right
+                    Component.onCompleted: console.log("width col", width)
+                    Repeater {
+                        id: repeater
+
+                        property int downItem
+                        property int buttonHeight: 50
```
mrwillyees, line 111: using a property to decide the buttomHeight. Could let it handled by the delegates, setting this property by delegates `onCompleted` event (lots of useless calls since I believe they will have all same height).
Possible problem: if the buttons change their height (may be a screen resize will trigger it?), their `height` and `buttomHeight` will not match anymore

uitools/import/Esri/ArcGISRuntime/Toolkit/FloorFilter.qml
```diff
@@ -89,19 +90,40 @@ Control {
...
+
+                        model: controller.levels
+                        delegate: ToolButton {
+                            Component.onCompleted: this.height = repeater.buttonHeight
```
mrwillyees, line 115: not creating a binding `height : repeater.buttonHeight` because the property `buttonHeight` will not change, so it would be a waste of a binding creation

uitools/import/Esri/ArcGISRuntime/Toolkit/FloorFilter.qml
```diff
@@ -366,32 +366,47 @@ Control {
...
+                orientation: Qt.Horizontal
+            }
 
-            Button {
+            Label {
                 id: showAllFacilities
-                flat: true
-                text: "show all facilities"
+                text: '<a href=" "style="text-decoration: none">All sites</a>'
+                textFormat: Text.RichText
```
mrwillyees, line 378: using a `Label` with links to mimic the blue text shown on design

uitools/import/Esri/ArcGISRuntime/Toolkit/FloorFilter.qml
```diff
@@ -366,32 +366,47 @@ Control {
                     }
                 }
             }
+            ToolSeparator {
+                Layout.fillWidth: true
+                Layout.columnSpan: 3
+                orientation: Qt.Horizontal
+            }
```
mrwillyees, line 373: using toolseparator on a single row, to create a separator between components.  Could use a `Rectangle` height 1, but each style would have hard-coded the color I set to it. @anmacdonald is that a good idea?

anmacdonald: I don't fully understand the problem here?

uitools/import/Esri/ArcGISRuntime/Toolkit/FloorFilter.qml
mrwillyees: No the color is not wrong, but from internal chat with Mark seems that the decided look is the blue text `All sites`. To get a blue color text without the overwriting of color in the tool, I thought of using the link functionality of a `Text` which in Calcite will output a blue text.
After click, no purple color.

uitools/import/Esri/ArcGISRuntime/Toolkit/FloorFilter.qml
mrwillyees: Not really a problem, just checking that using a `ToolSeparator` in a ListView to separate the component seems a good solution. It looks like this in default Qt style and similar in calcite:
image

anmacdonald: Oh I see, no that makes sense to me!

uitools/import/Esri/ArcGISRuntime/Toolkit/FloorFilter.qml
anmacdonald: Gotcha!

uitools/cpp/Esri/ArcGISRuntime/Toolkit/FloorFilterController.cpp
```diff
@@ -0,0 +1,698 @@
...
+      // Assumption: if too zoomed out to see sites, also too zoomed out to see facilities
+      return;
+    }
+
+    if (!floorManager)
+      return;
+
+    // If the centerpoint is within a site's geometry, select that site.
+    // This code gracefully skips selection if there are no sites or no matching sites
+    auto sites = floorManager ? floorManager->sites() : QList<FloorSite*>{};
```
mrwillyees, line 596: Don't need to check the validity of floorManager because it was checked on the line before.
```suggestion
    auto sites = floorManager->sites()
```

uitools/cpp/Esri/ArcGISRuntime/Toolkit/FloorFilterController.cpp
```diff
@@ -0,0 +1,698 @@
...
+                                       });
+
+    if (facilityResult != std::cend(facilities))
+    {
+      setSelectedFacilityId((*facilityResult)->facilityId());
+    }
+    else if (m_automaticSelectionMode == AutomaticSelectionMode::Always)
+    {
+      setSelectedFacilityId({});
+    }
```
mrwillyees, line 661: By testing out mine and yours viewpoint implementation I found a difference. In yours, when the facility is not in view, the levels visibility are not getting cleared. 
If you agree, then you are missing the `setSelectedLevelId({})`

https://user-images.githubusercontent.com/91129623/156550392-de1f2a0e-631c-4a39-bb29-8cfbe91ddea4.mp4

uitools/widgets/Esri/ArcGISRuntime/Toolkit/FloorFilter.cpp
```diff
@@ -0,0 +1,311 @@
...
+          const auto i = indexForId<FloorFilterSiteItem>(m_ui->sitesView->model(), newId);
+          m_ui->sitesView->selectionModel()->setCurrentIndex(i, QItemSelectionModel::SelectCurrent);
+        });
+
+    connect(
+        m_ui->sitesView, &QListView::doubleClicked, this, [this](QModelIndex index)
+        {
+          if (index.isValid())
+            m_ui->toolBox->setCurrentIndex(1);
+        });
```
mrwillyees, line 176: 💭 This functionality could be documented because I was not expecting at all, until I read this code. Quite useful honestly

uitools/widgets/Esri/ArcGISRuntime/Toolkit/FloorFilter.cpp
```diff
@@ -0,0 +1,311 @@
...
+    connect(
+        m_ui->sitesView->selectionModel(), &QItemSelectionModel::currentChanged, this, [this](QModelIndex index, QModelIndex previous)
+        {
+          if (index == previous)
+            return;
+
+          if (index == QModelIndex{})
+          {
+            m_controller->setSelectedSiteId("");
+          }
```
mrwillyees, line 187: 💭 If here you are trying to implement a deselection behavior, I checked the doc and by testing, I can see that the signal [ItemSelectionModel::currentChanged](https://doc.qt.io/qt-5/qitemselectionmodel.html#selectionChanged), lets you get a list of selected + deselected items (ctrl + click). Could be used to implement the deselection behavior (should not be MVP)

uitools/widgets/Esri/ArcGISRuntime/Toolkit/FloorFilter.cpp
```diff
@@ -0,0 +1,311 @@
...
+        m_ui->sitesView, &QListView::doubleClicked, this, [this](QModelIndex index)
+        {
+          if (index.isValid())
+            m_ui->toolBox->setCurrentIndex(1);
+        });
+
+    connect(
+        m_ui->sitesView->selectionModel(), &QItemSelectionModel::currentChanged, this, [this](QModelIndex index, QModelIndex previous)
+        {
+          if (index == previous)
```
mrwillyees, line 181: 💭  I noticed that the model is hooked up to the controller and when a facility is selected by panning around, this slot is being run. The operations set in this function have been already run in the controller, so it is performing twice. I would recommend to early return in case the sites were updated from the controller
```suggestion
          if (index == previous || m_facilitiesUpdatedFromController)
```

mrwillyees (commented):
Lots of nitpick, really really clean code!
Some of my suggestion are small improvements, on code that already works (small speed up). They are not fully tested, so might break something that I didn't take into account, take as they are. If you feel that it could be easily changed, go for it, otherwise let's keep the working code

tests/uitools/qml_quick/unit_tests/FloorFilterControllerUnitTest/tst_floorfiltercontroller.qml
```diff
@@ -0,0 +1,307 @@
...
+
+
+
+
+
+
+
+
+
+}// sort functionality
```
anmacdonald, line 307: Bad comment?

tests/uitools/qml_quick/functional_tests/FloorFilterControllerFuncTest/tst_floorfiltercontroller.qml
```diff
@@ -0,0 +1,189 @@
...
+        spy.wait();
+        console.log(map.floorManager.loadStatus);
+        if(map.floorManager.loadStatus !== Enums.LoadStatusLoaded)
+            spy.wait();
+        verify(map.floorManager.levels.length > 0);
+        verify(map.floorManager.levels.length === control.floorManager.levels.length);
+        for (var i = 0; i < map.floorManager.levels; ++i) {
+            verify(map.floorManager.levels[i].visible === control.floorManager.levels[i].visible);
+        }
+    }
```
anmacdonald, line 187: I think some of this logic, could be simplified by injecting the SignalSpy into the component directly, and forcing the FloorManager to load after the map has loaded also:

```qml

   Component {
        id: mapComponentMulti
        Map {
            id: map
            item: PortalItem {
                itemId: "49520a67773842f1858602735ef538b5"
                portal: Portal {
                    Component.onCompleted: load()
                    onErrorChanged: {
                        if (loadStatus === Enums.LoadStatusFailedToLoad) {
                            console.error(error.message);
                        }
                    }
                    onLoadStatusChanged: {
                        if (loadStatus === Enums.LoadStatusLoaded) {
                            map.floorManager.load()
                        }
                    }
                    credential: viewerCredentialIndoors
                    url: "https://indoors.maps.arcgis.com/"
                }
            }
            property SignalSpy spy: SignalSpy {
              target: map.floorManager
              signalName: "onLoadStatusChanged"
            }
        }
    }
```

tests/uitools/qml_quick/functional_tests/FloorFilterControllerFuncTest/tst_floorfiltercontroller.qml
```diff
@@ -0,0 +1,189 @@
...
+        compare(control.selectedSite.siteId, facility.site.siteId);
+    }
+
+    // singlesite mode on. only 1 site and 1 facility, automatically select them on load.
+    function test_singleSiteAutoSelect() {
+        var control = loadFloorManagerSingle();
+        verify(1, control.floorManager.sites.count);
+        compare(1, control.sites.count);
+        compare(1, control.facilities.count);
+        verify(control.selectedSiteId) // not empty string
```
anmacdonald, line 135: I'd be more comfortable comparing to an empty string than relying on implicit conversion here. `false == "0"` in javascript, which is the sort of thing which makes me nervous.

uitools/import/Esri/ArcGISRuntime/Toolkit/Controller/FloorFilterController.qml
```diff
@@ -0,0 +1,703 @@
...
+    onSelectedSiteRespectedChanged: {
+        if (!selectedSiteRespected) {
+            populateAllFacilities();
+        }
+    }
+    
+    /*!
+      \qmlproperty string selectedLevelId
+      \brief Stores the current selected level Id.
+      Default is {""}.
```
anmacdonald, line 200: Missing `\c`?

uitools/import/Esri/ArcGISRuntime/Toolkit/Controller/FloorFilterController.qml
```diff
@@ -0,0 +1,703 @@
...
+      \sa selectedSiteId
+    */
+    property ListModel facilities: ListModel {}
+
+    /*!
+      \qmlproperty ListModel sites
+      \brief Stores all the sites in the current \l floorManager.
+      \sa selectedSiteId
+    */
+    property ListModel sites: ListModel {}
```
anmacdonald, line 292: I think these need to be `readonly`, we want to mutate the lists, but I don't think we want to replace the lists wholesale, do we?

uitools/import/Esri/ArcGISRuntime/Toolkit/Controller/FloorFilterController.qml
```diff
@@ -0,0 +1,703 @@
...
+
+    property alias selectedSite: internal.selectedSite
+    property alias selectedFacility: internal.selectedFacility
+    property alias selectedLevel: internal.selectedLevel
```
anmacdonald, line 296: Should these be `readonly`, are they allowed to be settable?

anmacdonald (changes requested):
Fantastic work, very impressed with some of the polish for FloorFIlter QML. Couple of thoughts.

tests/uitools/qml_quick/functional_tests/FloorFilterControllerFuncTest/tst_floorfiltercontroller.qml
```diff
@@ -0,0 +1,189 @@
...
+TestCase {
+    id: floorFilterControllerFuncTest
+    name: "FloorFilterControllerUnitTest"
+
+    Credential {
+        id: viewerCredentialIndoors
+        // need to add CREDENTIAL_USERNAME and CREDENTIAL_PASSWORD to the Environment variables.
+        username: credential_username
+        password: credential_password
+    }
```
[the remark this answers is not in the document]
mrwillyees: harmful in the sense of useless?
I could stick with the publicly available single site, single facility test. I liked running with the internal webmaps because are more complicated than the public one.
I could modify the tests to run for both the public and the private in case the credential environment are set, otherwise only run for the public. In this way, internally we can still run those more complicated ones! How does it sound?

tests/uitools/qml_quick/shared/utils.h
```diff
@@ -0,0 +1,26 @@
...
+#include <QObject>
+#include <QQmlEngine>
+
+class Utils;
+
+class Utils : public QObject
+{
+  Q_OBJECT
+public:
+  explicit Utils(QObject* parent = 0);
```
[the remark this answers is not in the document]
mrwillyees: `Q_INVOKABLE` so we can be able to call it from QML?

uitools/examples/qml_quick/src/demos/FloorFilterDemoForm.qml
```diff
@@ -0,0 +1,49 @@
...
+import QtQuick 2.12
+import QtQuick.Controls 2.12
+import Esri.ArcGISRuntime 100.14
+import Esri.ArcGISRuntime.Toolkit 100.14
+
+DemoPage {
+    sceneViewContents: Component {
+        SceneView {
+            Scene {
+                initUrl: "https://www.arcgis.com/home/item.html?id=ed300862738440de94d777061fa04742"
```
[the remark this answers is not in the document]
mrwillyees: No, checked and if the Api key is set, it requires the credentials. If no Api key is set, it doesn't load. So credentials are needed.
TODO: find 3d scene public, or change this url

uitools/import/Esri/ArcGISRuntime/Toolkit/FloorFilter.qml
```diff
@@ -0,0 +1,599 @@
...
+        // absolute position center point floorfilter
+        property var floorFilterOrigin: (mapToItem(floorFilter.parent,
+                                                   floorFilter.x,
+                                                   floorFilter.y))
+        property double centerFloorFilterX: (floorFilterOrigin.x + floorFilter.width) / 2
+        property double centerFloorFilterY: (floorFilterOrigin.y + floorFilter.height) / 2
+        property double centerParentX: (parentOrigin.x + floorFilter.parent.width) / 2
+        property double centerParentY: (parentOrigin.y + floorFilter.parent.height) / 2
+        // position of floorfilter related to its parent.
+        property int leaderPosition: {
```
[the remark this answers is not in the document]
mrwillyees: Yes adopted that name because it was very similar as the `Callout` (wanted to be consistent). I am changing it then.

uitools/import/Esri/ArcGISRuntime/Toolkit/FloorFilter.qml
```diff
@@ -0,0 +1,599 @@
...
+                                && listView.visible)
+                            controller.zoomToSite(controller.selectedSiteId);
+                        else if (internal.currentVisibileListView
+                                 === FloorFilter.VisibleListView.Facility
+                                 || !listView.visible)
+                            controller.zoomToFacility(
+                                        controller.selectedFacilityId);
+                        else
+                            console.error("extra enum not accounted for.");
+                    }
```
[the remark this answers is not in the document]
mrwillyees: Removing the checks, the UI takes care of stopping the click. If user _really_ wants to click it, if I remove the checks it will be able to do it by finding the zoom `Button` object and trigger the signal `.clicked(null)`.

uitools/import/Esri/ArcGISRuntime/Toolkit/FloorFilter.qml
```diff
@@ -0,0 +1,599 @@
...
+                    // dont need to use the id of the column
+                    contentHeight: contentItem.childrenRect.height
+                    Layout.maximumHeight: repeater.buttonHeight * maxNumberLevels
+                    Layout.preferredHeight: repeater.buttonHeight * repeater.count
+                    Layout.fillWidth: true
+                    clip: true
+                    // scroll to the bottom of the flickable (so verticalorder 0 default selected level is visible)
+                    contentY : contentHeight-height
+                    ScrollBar.vertical: ScrollBar {
+                        policy: parent.contentHeight > parent.height ? ScrollBar.AlwaysOn : ScrollBar.AlwaysOff
```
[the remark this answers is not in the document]
mrwillyees: cheers

uitools/import/Esri/ArcGISRuntime/Toolkit/FloorFilter.qml
```diff
@@ -0,0 +1,599 @@
...
+                rowSpacing: 0
+                Button {
+                    objectName: "backToSite"
+                    Layout.margins: 0
+                    Layout.fillHeight: true
+                    Layout.rowSpan: 2
+                    Layout.alignment: Qt.AlignHCenter
+                    // setting width to 0 and not setting visibility to false, otherwise it would break the rows layout by having a missing row.
+                    Layout.preferredWidth: internal.currentVisibileListView
+                                           === FloorFilter.VisibleListView.Facility && controller.sites.count > 0 ? 32 : 0
```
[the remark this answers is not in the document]
mrwillyees: 32 is the default value I find looks good in case the currentView is facility and there are sites. Otherwise, hide the button by setting its height to 0

uitools/import/Esri/ArcGISRuntime/Toolkit/FloorFilter.qml
```diff
@@ -0,0 +1,599 @@
...
+                        anchors.fill: searchImg
+                        source: searchImg
+                        color: noResultsFoundLabel.color
+                    }
+                }
+
+                Label {
+                    id: noResultsFoundLabel
+                    objectName: "noResultsFoundLabel"
+                    text: "No results found"
```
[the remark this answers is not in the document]
mrwillyees: I see, this is for the translation. I guess, I should change all the string set in the UI to use this function `qsTr`?

uitools/import/Esri/ArcGISRuntime/Toolkit/FloorFilter.qml
```diff
@@ -0,0 +1,599 @@
...
+                                for (var i = 0; i < dm.items.count; ++i) {
+                                    var item = dm.items.get(i);
+                                    // empty string clears the filtering. All elements are visible (set into the `filtered` group).
+                                    if (searchTextField.text === "") {
+                                        item.inFiltered = true;
+                                    } else {
+                                        if (item.model.name.toLowerCase(
+                                                    ).includes(
+                                                    searchTextField.text.toLowerCase(
+                                                        )))
```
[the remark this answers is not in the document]
mrwillyees: I am positive it was, but won't do it again now. Setting it to a more sensible formatting

uitools/import/Esri/ArcGISRuntime/Toolkit/FloorFilter.qml
```diff
@@ -0,0 +1,599 @@
...
+
+    /*!
+      \qmlproperty enumeration updateLevelsMode
+      \brief The mode to use for updating levels visibility.
+      Default is \c {FloorFilterController.UpdateLevelsMode.AllLevelsMatchingVerticalOrder}.
+      \sa {FloorFilterController} {FloorFilterController.updateLevelsMode}
+    */
+    property int updateLevelsMode: controller.updateLevelsMode
+
+    property bool hideSiteFacilityButton: false
```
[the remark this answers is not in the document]
mrwillyees: I will keep in mind the rule of thumb. 
Regarding this property, I had a look at the design and could not find anything about it. I am planning to remove it because I don't see the usage: if the property is set to false, it will hide the `buildingMenuButton` which is the one that opens up the listView.

uitools/import/Esri/ArcGISRuntime/Toolkit/FloorFilter.qml
```diff
@@ -0,0 +1,599 @@
...
+                    ScrollBar.vertical: ScrollBar {}
+                    spacing: 5
+                    clip: true
+                    // scroll list view to selected item
+                    onCurrentIndexChanged: positionViewAtIndex(currentIndex, ListView.Visible)
+
+                    // delegate model used to implement the filtering functionality, by setting visible elements in the `filtered` group.
+                    model: DelegateModel {
+                        id: dm
+                        property Connections conn: Connections {
```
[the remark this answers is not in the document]
mrwillyees: Yes works as intended without an explicit property. 
For my understanding, why did I need to set into a property inside the `DelegateModel`?

uitools/import/Esri/ArcGISRuntime/Toolkit/FloorFilter.qml
```diff
@@ -0,0 +1,511 @@
...
+
+    property var geoView
+
+    property var controller: FloorFilterController {}
+
+    property int updateLevelsMode: controller.updateLevelsMode
+
+    property bool hideSiteFacilityButton: false
+
+    property int automaticSelectionMode: FloorFilterController.AutomaticSelectionMode.Always
```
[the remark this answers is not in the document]
anmacdonald: As the user has to access the controller anyway to get the enum, it makes sense to me to keep this on the controller only.

tests/uitools/qml_quick/shared/utils.cpp
```diff
@@ -0,0 +1,23 @@
+#include "utils.h"
+
+Utils::Utils(QObject* parent) :
+  QObject(parent)
+{
+}
+
+QObject* Utils::findChild(QObject* parent, QString name)
```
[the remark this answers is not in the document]
anmacdonald: Good find! Yes, let's use that instead.

uitools/cpp/Esri/ArcGISRuntime/Toolkit/FloorFilterController.cpp
```diff
@@ -0,0 +1,968 @@
...
+                           });
+      mapView->setViewpoint(b.toEnvelope());
+    }
+    else if (auto sceneView = qobject_cast<SceneViewToolkit*>(m_geoView))
+    {
+      m_settingViewpoint = true;
+      singleShotConnection(mapView, &MapViewToolkit::setViewpointCompleted, this, [this](bool /*success*/)
+                           {
+                             m_settingViewpoint = false;
+                           });
```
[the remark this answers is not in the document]
anmacdonald: Good catch! The sort of thing that should disappear in C++17 with if-statement initializes.

uitools/widgets/Esri/ArcGISRuntime/Toolkit/FloorFilter.cpp
```diff
@@ -0,0 +1,311 @@
...
+     */
+    template <typename T>
+    T* itemForIndex(QAbstractItemModel* model, QModelIndex modelIndex)
+    {
+      if (!model)
+        return nullptr;
+
+      if (modelIndex.isValid())
+      {
+        return qvariant_cast<T*>(model->data(modelIndex, Qt::UserRole));
```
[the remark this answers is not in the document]
anmacdonald: It's similar -- here we are not working on a`GenericListModel`, which is why the method is not available to us.

tests/uitools/qml_quick/shared/utils.h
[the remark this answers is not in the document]
anmacdonald: Beyond QML, this also registers the type with Qt's meta-object system.

uitools/import/Esri/ArcGISRuntime/Toolkit/FloorFilter.qml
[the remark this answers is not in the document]
anmacdonald: If possible!

uitools/import/Esri/ArcGISRuntime/Toolkit/FloorFilter.qml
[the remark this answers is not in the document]
anmacdonald: When left unassigned, it will be assigned to the "default" property: https://doc.qt.io/qt-5/qtqml-syntax-objectattributes.html#default-properties

For example, the default property on a QML `Item` is [children](https://doc.qt.io/qt-5/qml-qtquick-item.html#children-prop). I don't believe `DelegateModel` inherits from anything that allows is to have a default property.

uitools/import/Esri/ArcGISRuntime/Toolkit/Controller/FloorFilterController.qml
mrwillyees: You mean like this:
`\brief Stores the current \c string selected level Id.`

uitools/import/Esri/ArcGISRuntime/Toolkit/Controller/FloorFilterController.qml
mrwillyees: Yes correct. Even if the geoview/floormanager gets changed, the lists are the same, just with different content

uitools/import/Esri/ArcGISRuntime/Toolkit/Controller/FloorFilterController.qml
mrwillyees: No, they should not be settable directly from user. Just reflect what is in the `internal.selected<>`

uitools/import/Esri/ArcGISRuntime/Toolkit/Controller/FloorFilterController.qml
```diff
@@ -0,0 +1,703 @@
...
+      \brief Setting the levels visible that match the \a verticalOrder
+    */
+    function resetLevelsVisibility(verticalOrder){
+        for(var i = 0; i < floorManager.levels.length; ++i) {
+            var level = floorManager.levels[i];
+            level.visible = level.verticalOrder === verticalOrder;
+        }
+    }
+
+    function onSelectedChanged() {}
```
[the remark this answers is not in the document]
mrwillyees: May be might be better to transform to signal. This method is present in the design:
```
    //listener that is broadcasted when the selection of a site/facility/level is changed
    //allows the developer to add any logic when a site/facility/level is changed
    void onSelectedChanged()
```

uitools/import/Esri/ArcGISRuntime/Toolkit/Controller/FloorFilterController.qml
anmacdonald: Sorry I meant:

```
Default is \c{""}.
```

uitools/cpp/Esri/ArcGISRuntime/Toolkit/FloorFilterController.cpp
```diff
@@ -0,0 +1,968 @@
...
+    // Expectation: viewpoint is center and scale
+    if (m_automaticSelectionMode == AutomaticSelectionMode::Never ||
+        observedViewpoint.isEmpty() ||
+        isnan(observedViewpoint.targetScale()))
+      return;
+
+    auto floorManager = getFloorManager(m_geoView);
+
+    // Only take action if viewpoint is within minimum scale. Default minscale is 4300 or less (~zoom level 17 or greater)
+    double targetScale = 0.0;
```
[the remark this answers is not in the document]
mrwillyees: Oh yes, wonder why I would not see that, so obvious!

uitools/widgets/Esri/ArcGISRuntime/Toolkit/FloorFilter.cpp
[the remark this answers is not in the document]
mrwillyees: The question was more like:
should be beneficial to add this static_assert before casting?
`static_assert(std::is_base_of<QObject, T>::value, "Must inherit QObject");`

I was referring as the [similar method](https://github.com/Esri/arcgis-runtime-toolkit-qt/blob/main/uitools/cpp/Esri/ArcGISRuntime/Toolkit/Internal/GenericListModel.h#L91) implemented in the `GenericListModel`.

uitools/cpp/Esri/ArcGISRuntime/Toolkit/FloorFilterController.cpp
```diff
@@ -0,0 +1,968 @@
...
+    m_sites(new GenericListModel(&FloorFilterSiteItem::staticMetaObject, this))
+  {
+    m_sites->setDisplayPropertyName("name");
+    m_facilities->setDisplayPropertyName("name");
+    m_facilities->setTooltipPropertyName("parentSiteName");
+    m_levels->setDisplayPropertyName("longName");
+
+    connect(this, &FloorFilterController::selectedFacilityIdChanged, this, &FloorFilterController::selectedChanged);
+    connect(this, &FloorFilterController::selectedLevelIdChanged, this, &FloorFilterController::selectedChanged);
+    connect(this, &FloorFilterController::selectedSiteIdChanged, this, &FloorFilterController::selectedChanged);
```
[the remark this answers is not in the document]
anmacdonald: Good question - this signal was part of the original YAML design. In the transition to Qt signals - likely should be removed IMO.j

uitools/cpp/Esri/ArcGISRuntime/Toolkit/FloorFilterController.cpp
```diff
@@ -0,0 +1,968 @@
...
+                {
+                  if (level)
+                  {
+                    if (m_updatelevelMode == UpdateLevelsMode::AllLevelsMatchingVerticalOrder)
+                    {
+                      level->setVisible(newLevel ? level->verticalOrder() == newLevel->verticalOrder() : false);
+                    }
+                    else if (m_updatelevelMode == UpdateLevelsMode::SingleLevel)
+                    {
+                      level->setVisible(level->verticalOrder() == 0 || level == newLevel);
```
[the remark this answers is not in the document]
anmacdonald: Good point, didn't consider scenes.

I'll make the enum have one value for now, and we can add additional values later.

uitools/cpp/Esri/ArcGISRuntime/Toolkit/FloorFilterController.cpp
```diff
@@ -0,0 +1,968 @@
...
+    if (allLevels.isEmpty())
+    {
+      setSelectedLevelId({});
+      return;
+    }
+
+    // Sort levels by vertical order.
+    std::sort(std::begin(allLevels), std::end(allLevels), [](FloorLevel* a, FloorLevel* b)
+              {
+                return a && b ? a->verticalOrder() < b->verticalOrder() : static_cast<bool>(a);
```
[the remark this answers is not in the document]
anmacdonald: Is your thinking that these would be re-reversed in the UI?

uitools/import/Esri/ArcGISRuntime/Toolkit/Controller/FloorFilterController.qml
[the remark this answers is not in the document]
anmacdonald: All properties in QML come with an `on<PROPERTY>Changed` signal, so this should be given for free, I think?

uitools/cpp/Esri/ArcGISRuntime/Toolkit/FloorFilterController.cpp
```diff
@@ -0,0 +1,968 @@
...
+        setSelectedFacilityId({});
+        setSelectedSiteId({});
+      }
+      // Assumption: if too zoomed out to see sites, also too zoomed out to see facilities
+      return;
+    }
+
+    // If the centerpoint is within a site's geometry, select that site.
+    // This code gracefully skips selection if there are no sites or no matching sites
+    auto sites = floorManager ? floorManager->sites() : QList<FloorSite*>{};
```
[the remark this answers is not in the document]
anmacdonald: This is a change of behavior, as currently the behavior as written allows the siteId/facilityId/levelId to be cleared.

uitools/cpp/Esri/ArcGISRuntime/Toolkit/FloorFilterController.cpp
```diff
@@ -0,0 +1,968 @@
...
+      targetScale = 1500;
+    }
+
+    // If out of scale, stop here
+    if (observedViewpoint.targetScale() > targetScale)
+    {
+      return;
+    }
+
+    auto facilities = floorManager ? floorManager->facilities() : QList<FloorFacility*>{};
```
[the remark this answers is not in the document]
anmacdonald: This is a change of behavior, as currently the behavior as written allows the siteId/facilityId/levelId to be cleared.

uitools/widgets/Esri/ArcGISRuntime/Toolkit/FloorFilter.cpp
anmacdonald: Good find!

uitools/import/Esri/ArcGISRuntime/Toolkit/Controller/FloorFilterController.qml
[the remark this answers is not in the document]
mrwillyees: Yea, but I think the design wanted to have a general signal that triggers when a property is changed in the `floorfilter`.
Could I switch this function `onSelectedChanged()` to signal and emit it when `selectedSite/Facility/Level` is changed? I think you did something similar in your cpp (cannot find it anymore tho)
```
    connect(this, &FloorFilterController::selectedFacilityIdChanged, this, &FloorFilterController::selectedChanged);
    connect(this, &FloorFilterController::selectedLevelIdChanged, this, &FloorFilterController::selectedChanged);
    connect(this, &FloorFilterController::selectedSiteIdChanged, this, &FloorFilterController::selectedChanged);
```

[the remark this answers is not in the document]
anmacdonald: I did remove. As far as I can see, any change to site/facility/level will trigger `selectedLevelIdChanged`, making it the same as `selectedChanged`, so removed that.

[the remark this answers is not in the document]
mrwillyees: Doing the same then

uitools/cpp/Esri/ArcGISRuntime/Toolkit/FloorFilterController.cpp
```diff
@@ -0,0 +1,968 @@
...
+   level with verticalOrder == 0.
+   */
+  void FloorFilterController::populateLevelsForSelectedFacility()
+  {
+    m_levels->clear();
+    auto manager = getFloorManager(m_geoView);
+    if (!manager)
+      return;
+
+    auto allLevels = manager->levels();
```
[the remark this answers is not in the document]
anmacdonald: I think I see where you are going with this. It is a nice optimization, but not necessary at the moment. Given time constraints I'm going to maybe investigate this later.

PR #500, opened 2022-01-28
calcite actionbar implementation

mrwillyees:
implementation of the calcite actionbar using toolbar, toolbutton and toolseparator.
atm using the toolseparator as divisor between groups of actions as it can be seen here:
image
Quick demo showing difference between material implementation and calcite. to be noted is the not automatic fillwidth of a button to the parent (or parent layout), they need to be set manually for every button with `Layout.fillWidth : true`. Also, removed the backgorund `Rectangle` from the `ToolBar`
https://user-images.githubusercontent.com/91129623/151554530-8f9a1273-1f96-4a71-9fb3-8ab48173f04a.mp4


calcite-qml/Calcite/ToolBar.qml
```diff
@@ -0,0 +1,34 @@
...
+    implicitWidth: Math.max(implicitBackgroundWidth + leftInset + rightInset,
+                            contentWidth + leftPadding + rightPadding)
+    implicitHeight: Math.max(implicitBackgroundHeight + topInset + bottomInset,
+                             contentHeight + topPadding + bottomPadding)
+    spacing: 16
+    background: Rectangle {
+        implicitHeight: 48
+        color: "transparent"
+    }
+}
```
anmacdonald, line 34: 🔧 One thing to note is that in your current ToolBar implementation, there may be visible gaps between buttons unless they are squished tightly together explicitly with `spacing: 0`:

image

Based on JS FloorFilter / Figma, / Material Style, I  think you should have a filled background, as opposed to being transparent.

calcite-qml/Calcite/Calcite.qml
```diff
@@ -43,18 +43,28 @@ QtObject {
...
+    readonly property color foreground5: theme === Calcite.Theme.Light ? "#EAEAEA" : "#2B2B2B"
+
     readonly property color text1: theme === Calcite.Theme.Light ? "#151515" : "#FFFFFF"
 
     readonly property color text2: theme === Calcite.Theme.Light ? "#4A4A4A" : "#BFBFBF"
 
     readonly property color text3: theme === Calcite.Theme.Light ? "#6A6A6A" : "#9F9F9F"
 
+    readonly property color text4: theme === Calcite.Theme.Light ? "#2B2B2B" : "#FFFFFF"
+
```
anmacdonald, line 67: 🔧  We had a chat on restricting this list of var to: `https://github.com/Esri/calcite-colors/blob/master/src/colors.ts#L832`.

I am not against updating the existing color profiles to be conformant with figma/react.  But it is worth noting that as it currently stands,  that the ToolButtons look different in terms of highlighting behavior and text color to the rest of the available controls.

image

Looping in @nCastle1 as he might have an idea of what the most "correct" color scheme is.

calcite-qml/Calcite/ToolSeparator.qml
```diff
@@ -0,0 +1,36 @@
...
+ ******************************************************************************/
+import QtQuick 2.15
+import QtQuick.Templates 2.15 as T
+import QtQuick.Controls 2.15
+import QtQuick.Controls.impl 2.15
+
+T.ToolSeparator {
+    id: control
+
+    padding: 0
```
anmacdonald, line 25: ❓  Is this needed?

anmacdonald (changes requested):
👍 ToolButtons are looking great. There's a general question on conformity to the rest of the app that we'll need to think about.

Please remember to update **CHANGELOG.md` as part of this commit, and the Calcite README.

calcite-qml/Calcite/ToolBar.qml
mrwillyees: Fixing this by using the same color used in the `ToolBar` background.
An additional change could be to use by default `spacing : 0` since I cannot see any spacing in the specifications.

calcite-qml/Calcite/ToolSeparator.qml
mrwillyees: not needed, removed

mrwillyees:
reminder status: to be decided the style to use on the `ToolButton`, either the blue background used in the JS demos or the basic calcite (figma) `Action` component.

mrwillyees:
Implemented the blue background to be shown when `ToolBar.flat === true`, removed the new colors not Calcite standard and storing them into the tool directly

calcite-qml/Calcite/ToolButton.qml
```diff
@@ -0,0 +1,66 @@
...
+        spacing: control.spacing
+        mirrored: control.mirrored
+        display: control.display
+
+        icon: control.icon
+        text: control.text
+        font: control.font
+
+        color: control.icon.color
+        alignment: control.display === AbstractButton.TextBesideIcon ? Qt.AlignLeft : Qt.AlignCenter
```
anmacdonald, line 54: ❌  Looking at this in retrospect - the text no longer aligns properly.

image

Left alignment is possible the wrong way to go about this. Either we keep everything centered or we try something like:

```qml
alignment: control.display === AbstractButton.TextBesideIcon && icon.source.toString() !== "" && text != "" ? Qt.AlignLeft : Qt.AlignCenter
```

Thoughts?

anmacdonald (changes requested):
👍 Getting there. More to say on color scheme.

calcite-qml/Calcite/ToolBar.qml
```diff
@@ -0,0 +1,36 @@
...
+    implicitWidth: Math.max(implicitBackgroundWidth + leftInset + rightInset,
+                            contentWidth + leftPadding + rightPadding)
+    implicitHeight: Math.max(implicitBackgroundHeight + topInset + bottomInset,
+                             contentHeight + topPadding + bottomPadding)
+
+    spacing: 16
+
+    background: Rectangle {
+        implicitHeight: 48
+        color: Calcite.foreground1
```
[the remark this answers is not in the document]
mrwillyees: The `ToolBar` is the container of the buttons. The Calcite respective should be the `Action Bar`. From the Figma, I can see that it follows the idle colors of the buttons. From the screenshot, I believe that long spacing is not a button, but the below `ActionBar`, so I m incline to follow idle colors used for calcite (not flat). Does it answer the query?
image
image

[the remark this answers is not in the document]
anmacdonald: I'm afraid I'm not clear on this one. Are you advocating for using a foreground color as a background color?

anmacdonald (approved):
👍 Thanks for the changes and talking it through with me.

PR #501, opened 2022-02-02
using groupfilters to hide and show no result found delegate.

mrwillyees:
the item "no results found" is part of the listview and it is shown or hidden using the group filtering functionality provided by the `DelegateModel`. This item is not part of the `controller.model`, but it is stored in the `DelegateModelGroup.name == noResultsFound`.

clicking on show all facilities needs fix because the temporary no resuilts found is still shown.

Any better solution in your opinion?

### DIFF
--- a/uitools/import/Esri/ArcGISRuntime/Toolkit/Controller/FloorFilterController.qml
+++ b/uitools/import/Esri/ArcGISRuntime/Toolkit/Controller/FloorFilterController.qml
@@ -48,9 +48,19 @@ QtObject {
 
     property ListModel levels: ListModel {}
 
-    property ListModel facilities: ListModel {}
+    property ListModel facilities: ListModel {
+        ListElement {
+            name: ""
+            modelId: ""
+        }
+    }
 
-    property ListModel sites: ListModel {}
+    property ListModel sites: ListModel {
+        ListElement {
+            name: ""
+            modelId: ""
+        }
+    }
 
     property alias selectedSite: internal.selectedSite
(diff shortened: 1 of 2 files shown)


uitools/import/Esri/ArcGISRuntime/Toolkit/Controller/FloorFilterController.qml
```diff
@@ -48,9 +48,19 @@ QtObject {
 
     property ListModel levels: ListModel {}
 
-    property ListModel facilities: ListModel {}
+    property ListModel facilities: ListModel {
+        ListElement {
+            name: ""
+            modelId: ""
```
mrwillyees, line 54: ListModels needs to be initiated with roles, otherwise at line 289 in `floorfilter.qml`, a new `ListElement` is created and will  set the roles as no names. Meaning that when populating the `sites`, the roles set are null and no matching will be found, showing a delegate with no text. (was not expecting this from the docs: https://doc.qt.io/qt-5/qml-qtqml-models-delegatemodelgroup.html#insert-method-1
Not wokring Demo below:
https://user-images.githubusercontent.com/91129623/152151573-866a3437-481f-4dd3-a405-8b685f20c88a.mp4

Working demo below:
https://user-images.githubusercontent.com/91129623/152151941-8fc4416f-30c8-448c-b3f4-626326a3674c.mp4

mrwillyees:
closing. using different approach: swap list view with label with text "no results found", having visible prop opposite as listview
20e23058550dfc0da87af1e38cdef71f699c62e7

PR #503, opened 2022-02-14
Floorfilter/viewpoint

mrwillyees:
autoselection functionality, based on current viewpoint .


uitools/import/Esri/ArcGISRuntime/Toolkit/Controller/FloorFilterController.qml
```diff
@@ -34,22 +110,42 @@ QtObject {
...
+      if levelId == "", disable visibility previous level
+      */
+    function setSelectedLevelId(levelId) {
+        let idx = findElementIdxById(levelId,
+                                     FloorFilterController.TypeElement.Level)
+        if (levelId === "" && idx != null) {
+            // reset the previous level visibility
+            if (internal.selectedLevel) {
+                internal.selectedLevel.visible = false
+            }
```
mrwillyees, line 137: the visiblity disable might be extracted in a separate function.
But if we agree that passing an empty string means to disable visibility of current level, can remain as it is.

uitools/import/Esri/ArcGISRuntime/Toolkit/Controller/FloorFilterController.qml
```diff
@@ -23,6 +23,82 @@ QtObject {
...
+        }
+
+        let selectSite
+        for (var i = 0; i < floorManager.sites.length; ++i) {
+            if (GeometryEngine.intersects(floorManager.sites[i].geometry,
+                                          viewpointCenter.center)) {
+                selectSite = floorManager.sites[i]
+                break
+            }
+        }
```
anmacdonald, line 71: Consider using higher-order functions when `Array` is available.
```js
const selectedSite = floorManager.sites.find(s => GeometryEngine.intersects(s.geometry, viewpointCenter.center))
```

uitools/import/Esri/ArcGISRuntime/Toolkit/Controller/FloorFilterController.qml
```diff
@@ -34,22 +110,42 @@ QtObject {
         return null
     }
 
-    property int updateLevelsMode: FloorFilterController.UpdateLevelsMode.AllLevelsMatchingVerticalOrder
+    property int updateLevelsMode: FloorFilterController.UpdateLevelsMode.SingleLevel
```
anmacdonald, line 113: Remember to keep this as `AllLevelsMatchingVerticalOrder` for the final commit.

anmacdonald (changes requested):
Most of the changes makes sense so far! In terms of coding style I am seeing a lot of missing semicolons.

uitools/import/Esri/ArcGISRuntime/Toolkit/Controller/FloorFilterController.qml
mrwillyees: @anmacdonald what do you think about this? Is it ok to integrate the `disableVisiblityLastLevel` into the `setSelectedLevelId` if passing an empty string levelId?
I dont think there can be Ids with empty strings, so should be safe

uitools/import/Esri/ArcGISRuntime/Toolkit/Controller/FloorFilterController.qml
anmacdonald: Sorry, that's entirely my fault. This isn't a JS `Array`,  it's a [QQmlListPoperty](https://doc.qt.io/qt-5/qqmllistproperty.html), so it doesn't have the same effective abilities. 😭

JamesMBallard: Would https://bugreports.qt.io/browse/QTBUG-49613 solve this? This is an unfortunate pitfall of the API.

mrwillyees: @JamesMBallard does it mean that the functionality will come some time in the Qt 6 release?
At the moment I am transforming to an Array, but I have noticed that the JS implementation is a deep copy, so I think quite expensive to do at every viewpoint changed. I will roll it back to the traditional for loop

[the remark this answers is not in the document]
anmacdonald: I prefer the for loop, personally!

JamesMBallard: @mrwillyees unlikely since both the issues are still open and categorized as merely suggestions.

uitools/import/Esri/ArcGISRuntime/Toolkit/Controller/FloorFilterController.qml
```diff
@@ -38,18 +122,38 @@ QtObject {
 
     property bool autoselectSingleFacilitySite
 
+    property var automaticSelectionMode
```
anmacdonald, line 125: 🙃 Use `int` instead of `var`.

uitools/import/Esri/ArcGISRuntime/Toolkit/Controller/FloorFilterController.qml
```diff
@@ -23,6 +23,90 @@ QtObject {
...
     property GeoView geoView
 
+    property Connections geoViewConnections: Connections {
+        target: geoView
+        function onViewpointChanged() {
+            tryUpdateSelection()
+        }
+        // manually set true when floorManager is loaded
+        enabled: false
+    }
```
anmacdonald, line 33: 🔧 Move `geoViewConnections` into the `internal` object, otherwise it will be public in the controller interface..

uitools/import/Esri/ArcGISRuntime/Toolkit/Controller/FloorFilterController.qml
```diff
@@ -38,18 +122,38 @@ QtObject {
...
     property bool autoselectSingleFacilitySite
 
+    property var automaticSelectionMode
+
     property FloorManager floorManager
 
     //refresh()?
     property string selectedLevelId
 
     property string selectedFacilityId
```
anmacdonald, line 132: 🔧  Consider making these 

```qml
property readonly alias selectedLevelId: internal.selectedLevelId
``` 

This way they can only be set using your custom functions in the public interface.

uitools/import/Esri/ArcGISRuntime/Toolkit/Controller/FloorFilterController.qml
```diff
@@ -159,6 +266,8 @@ QtObject {
                                      FloorFilterController.TypeElement.Site)
         if (idx == null) {
             console.error("site id not found")
+            // todo: set site as null?
```
anmacdonald, line 269: 🙃 This `TODO` still relevant?

uitools/import/Esri/ArcGISRuntime/Toolkit/FloorFilter.qml
```diff
@@ -44,6 +44,14 @@ Control {
     //debug property: should always be true, autoselecting singlefacilitiesistes in case they are single
     property bool autoselectSingleFacilitySite: false
 
+    property var automaticSelectionMode: FloorFilterController.AutomaticSelectionMode.Always
```
anmacdonald, line 47: 🙃 `property int`

anmacdonald (changes requested):
👍 LGTM, Some minor comments but no blockers.

There's still a number of missing semicolons, and `console.log` statements which I have ignored for now. Ensure these are cleaned up before merging into `v.next`!

PR #506, opened 2022-03-08
Floorfilter/tooltip

mrwillyees:
showing tooltip in widget FloorFilter. This change also fixes the `allSites` for the cpp version, which uses the model `tooltip()` and shows it in the newline when clicking `allSites`.
image


anmacdonald (approved):
👍 LGTM. Good finds/edits!

PR #509, opened 2022-03-23
Adding helpers

anmacdonald:
This PR includes some additional helper methods added to the Toolkit and some formatting tweaks/updates to the existing tools.

### DIFF
--- a/uitools/cpp/Esri/ArcGISRuntime/Toolkit/Internal/DoOnLoad.h
+++ b/uitools/cpp/Esri/ArcGISRuntime/Toolkit/Internal/DoOnLoad.h
@@ -1,0 +1,84 @@
+/*******************************************************************************
+ *  Copyright 2012-2020 Esri
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ ******************************************************************************/
+#ifndef ESRI_ARCGISRUNTIME_TOOLKIT_INTERNAL_DOONLOAD_H
+#define ESRI_ARCGISRUNTIME_TOOLKIT_INTERNAL_DOONLOAD_H
+
+// ArcGISRuntime headers
+#include <Loadable.h>
+
+// Qt headers
+#include <QObject>
+
+// C++ headers
+#include <type_traits>
+
+namespace Esri {
+namespace ArcGISRuntime {
+namespace Toolkit {
+
+  /*
+    \internal
+    \brief Executes method \a f immediately if \a sender is loaded, otherwise
+    attempts to load \a sender and then executes \a f after the load occurs.
+    Returns a connection object which may be default-constructed.
+
+    Example:
+    \begincode
+    doOnLoad(floorManager, this, []
+             {
+                qDebug() << "FloorManager is guaranteed loaded."
+             });
+    \endcode
+   */
+  template <typename Target, typename Func>
+  QMetaObject::Connection doOnLoad(Target* sender, QObject* reveiver, Func&& f)
+  {
+    static_assert(
+        std::is_convertible<Target*, ::Esri::ArcGISRuntime::Loadable*>::value,
+        "Target type must use Loadable interface");
+    static_assert(
+        std::is_convertible<Target*, QObject*>::value,
+        "Target type must be a QObject");
+
+    if (sender->loadStatus() == LoadStatus::Loaded)
+    {
+      f();
+      return QMetaObject::Connection{};
+    }
+    else
+    {
+      auto connection = QObject::connect(
+          sender, &Target::loadStatusChanged, reveiver,
+          [f = std::forward<Func>(f)](LoadStatus loadStatus)
+          {
+            if (loadStatus == LoadStatus::Loaded)
+              f();
+          });
+
+      if (sender->loadStatus() == LoadStatus::NotLoaded)
+        sender->load();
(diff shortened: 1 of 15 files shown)


uitools/cpp/Esri/ArcGISRuntime/Toolkit/Internal/DoOnLoad.h
```diff
@@ -0,0 +1,84 @@
...
+
+  /*
+    \internal
+    \brief Executes method \a f immediately if \a sender is loaded, otherwise
+    attempts to load \a sender and then executes \a f after the load occurs.
+    Returns a connection object which may be default-constructed.
+
+    Example:
+    \begincode
+    doOnLoad(floorManager, this, []
```
lsmallwood, line 40: maybe this should be `doOnLoaded` since the action takes place when load completes (e.g. status is loaded) and not when you call `load()`?

uitools/cpp/Esri/ArcGISRuntime/Toolkit/Internal/DoOnLoad.h
```diff
@@ -0,0 +1,84 @@
...
+          sender, &Target::loadStatusChanged, reveiver,
+          [f = std::forward<Func>(f)](LoadStatus loadStatus)
+          {
+            if (loadStatus == LoadStatus::Loaded)
+              f();
+          });
+
+      if (sender->loadStatus() == LoadStatus::NotLoaded)
+        sender->load();
+      else if (sender->loadStatus() != LoadStatus::Loading)
```
lsmallwood, line 73: this looks wrong: I think, if it's loading you can just wait for load status changed. You should only call `retryLoad()` if it's `FailedToLoad`

uitools/cpp/Esri/ArcGISRuntime/Toolkit/Internal/GeoViews.h
```diff
@@ -52,4 +46,30 @@ namespace Toolkit
 
 #endif
 
+namespace Esri {
+namespace ArcGISRuntime {
+namespace Toolkit {
+  inline Map* getGeoModel(MapViewToolkit* mapView)
```
lsmallwood, line 52: if we're not actually returning the base `GeoModel` should these just be named `getMap`?

uitools/cpp/Esri/ArcGISRuntime/Toolkit/Internal/GeoViews.h
```diff
@@ -52,4 +46,30 @@ namespace Toolkit
...
+  {
+    return mapView ? mapView->map() : nullptr;
+  }
+
+  inline Scene* getGeoModel(SceneViewToolkit* sceneView)
```
lsmallwood, line 57: any reason we don't implement these in a cpp?

uitools/cpp/Esri/ArcGISRuntime/Toolkit/Internal/GeoViews.h
```diff
@@ -52,4 +46,30 @@ namespace Toolkit
...
+  {
+    return sceneView ? sceneView->arcGISScene() : nullptr;
+  }
+
+  inline auto getGeoModelChangedSignal(MapViewToolkit* /*mapView*/)
```
lsmallwood, line 62: 💭  why do we need an arg here? seems like this could be templated rather than passing an arg we never use...

uitools/cpp/Esri/ArcGISRuntime/Toolkit/Internal/DoOnLoad.h
anmacdonald: This covers two cases:

* LoadStatusUnknown
* LoadStatusFailedToLoad

My question is: what's does one do in face of a `LoadStatusUnknown`?

uitools/cpp/Esri/ArcGISRuntime/Toolkit/Internal/DoOnLoad.h
anmacdonald: this is a good comment! As this is used across a couple of these PRs I'll make this change upstream.

uitools/cpp/Esri/ArcGISRuntime/Toolkit/Internal/DoOnLoad.h
lsmallwood: Not sure. I don't think it is ever used in RTC so may be designed for QML declarative type scenarios?

ldanzinger (approved):
LGTM

uitools/cpp/Esri/ArcGISRuntime/Toolkit/Internal/GeoViews.h
anmacdonald: These are part of a way of giving a `getGeoModel` method to GeoViews as used here:

https://github.com/Esri/arcgis-runtime-toolkit-qt/blob/b6c0d18982a4faef43f7b368cbc5a329f93924dc/uitools/cpp/Esri/ArcGISRuntime/Toolkit/FloorFilterController.cpp#L95

So `getGeoModel` allows for one to be agnostic with the passed type.

uitools/cpp/Esri/ArcGISRuntime/Toolkit/Internal/GeoViews.h
anmacdonald: This is for a template trick that allows for one to agnostically get the right signal based on some arbitrary object:

https://github.com/Esri/arcgis-runtime-toolkit-qt/blob/b6c0d18982a4faef43f7b368cbc5a329f93924dc/uitools/cpp/Esri/ArcGISRuntime/Toolkit/FloorFilterController.cpp#L120

uitools/cpp/Esri/ArcGISRuntime/Toolkit/Internal/GeoViews.h
anmacdonald: Currently this is  a header-only file at the moment. These two methods could be added to a C++, but no other file can be moved ATM.

I'd like to keep them inlined (for now), as these are a cut-and-dry-case of when inlining is appropriate, IMO. But the long-term goal is to get rid of these entirely and rely on a `getGeoModel` coming from the API.

PR #513, opened 2022-03-28
fix racecondition/early delete of connection

mrwillyees:
Problem: when selecting a facility from the list `allSites`, it would zoom in and out immediately.
Figured out that the `singleShotConnection` in `zoomToEnvelope`, would run and then delete the connection without taking into account of the success of the setting of viewpoint. We are quickly calling `setViewpoint` for a facility and its parent site, but the second call is most likely called while the  first `setViewpoint` is still in progress. This triggers the `setViewpointCompleted` with `success == false`. In the previous code, we would then set the `m_settingViewpoint == false`, letting the `tryUpdate` to work. Problematically, this is done before the second `setViewpointCompleted` is done, so the `tryUpdate` starts evaluating the viewpoint caused from the `zoomToEnvelope`. Big nono, fixed with a traditional singleshotconnection but taking into account of the success!
Have a look if we can clean it please!
TODO: do the same for the sceneView

### DIFF
--- a/uitools/cpp/Esri/ArcGISRuntime/Toolkit/FloorFilterController.cpp
+++ b/uitools/cpp/Esri/ArcGISRuntime/Toolkit/FloorFilterController.cpp
@@ -828,10 +828,10 @@ namespace Toolkit {
     EnvelopeBuilder b{envelope};
     b.expandByFactor(ZOOM_PADDING);
 
+    auto c = std::make_shared<QMetaObject::Connection>();
     if (auto mapView = qobject_cast<MapViewToolkit*>(m_geoView))
     {
       m_settingViewpoint = true;
-      auto c = std::make_shared<QMetaObject::Connection>();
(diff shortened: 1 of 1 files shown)


uitools/cpp/Esri/ArcGISRuntime/Toolkit/FloorFilterController.cpp
```diff
@@ -829,22 +828,31 @@ namespace Toolkit {
     EnvelopeBuilder b{envelope};
     b.expandByFactor(ZOOM_PADDING);
 
+    auto c = std::make_shared<QMetaObject::Connection>();
```
anmacdonald, line 831: 🙃  Minor tweak: it's better to have two `c` declarations as opposed to one. Overall, keep declarations in the tightest scope possible.

anmacdonald (approved):
LGTM

Minor tweak, but no need to re-review!

PR #514, opened 2022-03-29
bubble sort impl on the GUI. removed sort from controller

mrwillyees:
cpp and qml are sorting facilities in different places.
cpp using the proxysort model on the gui.
qml sorts in the controller.
Doing so, qml GUI and cpp controller has no facilities sorted (controller expects that facilities sorted).

here, I implemented a non optimized bubble sort as a proof of concept and I use it to sort on the GUI (kinda proxysortdelegate model). Any thoughts?
A second approach would be to sort facilities in the cpp controller so they are sorted for the case qml GUI and cpp controller. This means that in the case of cpp GUI and cpp controller, the facilities are sorted twice


anmacdonald (approved):
LGTM.

Sprint X I think we might need/want to inject some C++ ListModels into QML.

Possibly GenericListModel + some sort of easy to use SortFilterProxyModel.

We're hitting a lot of funky model-based workarounds as QML isn't completely suitable for our needs.

PR #518, opened 2022-03-31
Awilliams/timeslieder enhancements

mrwillyees:
implemented the enhancement requested for timeslider:
-  Disable forward and rewind buttons when in play
-  Allow press and hold on forward and rewind buttons to skip until released

removed onClick trigger and replaced with the timer running immediately the first time and only after 200ms keep running as it is hold. I tested and it feels all good, no weird behavior

### DIFF
--- a/uitools/import/Esri/ArcGISRuntime/Toolkit/TimeSlider.qml
+++ b/uitools/import/Esri/ArcGISRuntime/Toolkit/TimeSlider.qml
@@ -201,9 +201,16 @@ Pane {
         Button {
             id: stepBackButton
             icon.source: "images/step_back.png"
-            enabled: (!startTimePinned || !endTimePinned) && playAnimation.running
+            enabled: (!startTimePinned || !endTimePinned) && !playAnimation.running
             palette: timeSlider.palette
-            onClicked: timeSlider.incrementFrame(-1);
+            Timer {
+                id: pressedHoldBack
+                repeat: true
+                running: parent.pressed
+                interval: 200
+                onTriggered: timeSlider.incrementFrame(-1);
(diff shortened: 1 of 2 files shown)


mrwillyees:
shortFormat date, small window. 2 lines, clips around the `Labels`
image
longFormat date, small window. 3 lines, clips around the `Labels`
image

uitools/import/Esri/ArcGISRuntime/Toolkit/TimeSlider.qml
```diff
@@ -245,6 +262,9 @@ Pane {
             Layout.alignment: Qt.AlignRight  | Qt.AlignBottom
             Layout.fillWidth: true
             Layout.margins: 5
+            wrapMode: Text.WordWrap
+            clip: true
+            maximumLineCount: 3
```
anmacdonald, line 267: Could maybe add `elide: Text.ElideLeft?`
 https://doc.qt.io/qt-5/qml-qtquick-text.html#elide-prop

anmacdonald (approved):
This is a good quality of life improvement.

uitools/import/Esri/ArcGISRuntime/Toolkit/TimeSlider.qml
mrwillyees: I tried using it earlier, but only the last string will be elided. And Only in case of ElideRight, so in the demo I m using both ElideRight

https://user-images.githubusercontent.com/91129623/161112398-842e8a09-da2c-41ab-a7a0-4d77d8976ba3.mp4

.

anmacdonald:
LGTM

PR #519, opened 2022-04-15
added missing svg

mrwillyees:
added svg module for IOS only, otherwise all the svg icons are not rendered in IOS


mrwillyees:
fyi @anmacdonald

anmacdonald (approved):
LGTM good find

PR #521, opened 2022-04-18
Fixing callout width/height calculations.

anmacdonald:
Making changes to `Callout` with a custom component. 

The height is no longer `fillHeight: true`, which means the callout sets the component width, and the component sets the callout height. This seems closer to the flow of width/height calculations used previously.


anmacdonald:
Related: https://github.com/Esri/arcgis-runtime-samples-qt/pull/1371

mrwillyees:
I get it, thanks for the heads up; will test in certification.
LGTM

mrwillyees (approved):

PR #522, opened 2022-04-18
Adding required  lines.

anmacdonald:
Toolkit, like ArcGIS, includes it's own required `QT +=` dependencies. The most important of which are `webview` for AuthenticationView, and `svg` for icon rendering.

### DIFF
--- a/uitools/toolkitwidgets.pri
+++ b/uitools/toolkitwidgets.pri
@@ -13,6 +13,8 @@
 # limitations under the License.
 include($$PWD/common.pri)
 
+QT += widgets webenginewidgets svg
+
 RESOURCES += $$PWD/images/esri_arcgisruntime_toolkit_images.qrc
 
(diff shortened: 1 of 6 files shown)


uitools/toolkitwidgets.pri
```diff
@@ -13,6 +13,8 @@
 # limitations under the License.
 include($$PWD/common.pri)
 
+QT += widgets webenginewidgets svg
```
JamesMBallard, line 16: ❓ 
Is `webenginewidgets` new? Seems right we'd need it.

JamesMBallard (approved):

uitools/toolkitwidgets.pri
anmacdonald: It's been around since Qt 5.4 AFAIK. The inclusion of "OAuthView" in Widgets forces it as a requirement unless we can make it a conditional compilation in the future.

PR #525, opened 2022-04-22
Fixing height issues.

anmacdonald:
Fixed issues where a a custom callout component was not rendering if it did not produce its own height.


JamesMBallard (approved):
LGTM.

PR #531, opened 2022-05-16
Tweaking DialogButtonBox.

anmacdonald:
Some tweaks to DialogButtonBox. If you compare to Materials style DialogButtonBox this is more in line with how they have written theirs out.

1. This is to suppress the "implicitWidth biding loop detected" warning
2. Added some padding around the background.
3. ButtonLayout is no longer forced to Mac layout as default.


mfeigl (approved):
LGTM

PR #537, opened 2022-06-28
Ldanzinger/10015 patch

ldanzinger:
@mfeigl please review. Qt team is updating version to always include patch to make things easier when a patch release is created. Also, renamed calcite project file from calite to calcite


mfeigl (changes requested):
LGTM, I have a few comments only:

❓ Should the following files update `ARCGIS_RUNTIME_VERSION = 100.15` as well:
* BasemapGalleryItemUnitTest.pro
* FloorFilterControllerFuncTest.pro
* FloorFilterFuncTest.pro
* FloorFilterControllerUnitTest.pro

❌ The PR diff gives the impression that `calite_test.pro` was renamed without using `git mv`; if so, we would lose the git history and I'd recommend renaming it back to the incorrect spelling and then renaming again, now with `git mv`

ldanzinger:
@mfeigl good call, updated those tests too. Strange, I did use git mv...

PR #538, opened 2022-07-07
Update ScalebarController.qml

ldanzinger:
Per a user request, marking as internal for the time being. We only have C++ Qt Quick Scale Bar, and the QML controller is there only as a stub. Currently it shows up on the developer site even though it isn't implemented for QML API which is causing confusion

https://developers.arcgis.com/qt/toolkit/api-reference/qml-scalebarcontroller.html

@mfeigl please review.


mfeigl (approved):
LGTM!

PR #541, opened 2022-09-01
Version update from 100.15 to 200.0.0

mfeigl:
Updating the version from 100.15 to 200.0.0

Note, qml `import` no longer requires the version number to be specified.


mrwillyees (approved):
LGTM
we also have removed all the import versioning from qt modules

PR #542, opened 2022-09-13
Marton/qt6 upgrade

mfeigl:
Upgrading the toolkit from Qt 5 into Qt 6.2.4 as well as removed deprecated code.

### DIFF
--- a/calcite-qml/Calcite/CheckIndicator.qml
+++ b/calcite-qml/Calcite/CheckIndicator.qml
@@ -14,7 +14,7 @@
  *  limitations under the License.
  ******************************************************************************/
 import QtQuick 2.15
-import QtGraphicalEffects 1.12
+import Qt5Compat.GraphicalEffects
 
 Rectangle {
     property var control;
--- a/uitools/examples/cpp_quick/src/main.cpp
+++ b/uitools/examples/cpp_quick/src/main.cpp
@@ -36,7 +36,7 @@
 int main(int argc, char* argv[])
 {
   QtWebView::initialize();
-  QGuiApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
+  //QGuiApplication::setAttribute(Qt::AA_EnableHighDpiScaling); --- this flag is not needed for Qt6
   QGuiApplication app(argc, argv);
 
   // Use of Esri location services, including basemaps and geocoding, requires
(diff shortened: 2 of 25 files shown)


lsmallwood (approved):
LGTM @mfeigl

calcite-qml/Calcite/CheckIndicator.qml
```diff
@@ -14,7 +14,7 @@
  *  limitations under the License.
  ******************************************************************************/
 import QtQuick 2.15
-import QtGraphicalEffects 1.12
+import Qt5Compat.GraphicalEffects
```
ldanzinger, line 17: Please log a follow up issue to investigate alternatives, and possibly remove usage before 200.0

uitools/examples/cpp_quick/src/main.cpp
```diff
@@ -61,21 +60,19 @@ int main(int argc, char* argv[])
     Esri::ArcGISRuntime::ArcGISRuntimeEnvironment::setApiKey(apiKey);
   }
 
-  QQuickStyle::addStylePath("qrc:///esri.com/imports/");
```
ldanzinger, line 64: does calcite styling still work with this removed?

uitools/import/Esri/ArcGISRuntime/Toolkit/SearchView.qml
```diff
@@ -19,7 +19,7 @@ import Esri.ArcGISRuntime.Toolkit.Controller
 import QtQuick 2.12
 import QtQuick.Controls 2.12
```
ldanzinger, line 20: please log a follow up issue to remove all QML module versions from the imports

ldanzinger (approved):
Changes look good to me. I commented with a couple follow up issues we should make. 

I also have a few other questions:
- You mentioned some tools weren't appearing - are they working now? If not, can you log an issue for that?
- I think we need an issue logged to investigate our C++ registration since we specify all of these versions 100.x versions, and presumably don't need to now, or minimally need to change the current major minor version - https://github.com/Esri/arcgis-runtime-toolkit-qt/blob/marton/qt6_upgrade/uitools/register/Esri/ArcGISRuntime/Toolkit/internal/register_cpp.cpp
- Was there no usage of Runtime deprecated code in the Toolkit? Don't see any removals here

You can get started on these issues next

uitools/import/Esri/ArcGISRuntime/Toolkit/Controller/qmldir
```diff
@@ -15,7 +15,7 @@
 module Esri.ArcGISRuntime.Toolkit.Controller
 
 # Current version increment
-CurrentVersion 200.0.0 CurrentVersion.qml
+CurrentVersion 200.0 CurrentVersion.qml
```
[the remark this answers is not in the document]
mfeigl: Nope, good spot!

calcite-qml/Calcite/CheckIndicator.qml
mfeigl: Done

uitools/import/Esri/ArcGISRuntime/Toolkit/SearchView.qml
mfeigl: Done.

mfeigl:
> You mentioned some tools weren't appearing - are they working now? If not, can you log an issue for that?

Done.


> I think we need an issue logged to investigate our C++ registration since we specify all of these versions 100.x versions, and presumably don't need to now, or minimally need to change the current major minor version - https://github.com/Esri/arcgis-runtime-toolkit-qt/blob/marton/qt6_upgrade/uitools/register/Esri/ArcGISRuntime/Toolkit/internal/register_cpp.cpp


I changed this to be version 200 - will create an issue to investigate further. 

> Was there no usage of Runtime deprecated code in the Toolkit? Don't see any removals here

I think there was only one deprecation: [Time Slider Controller](uitools/cpp/Esri/ArcGISRuntime/Toolkit/TimeSliderController.cpp) changed `LayerListModel::layerAdded` into `LayerListModel::itemAdded`.

uitools/examples/cpp_quick/src/main.cpp
mfeigl: I think it is still fine - here's some details on why it needed removed: https://doc.qt.io/qt-6/qtquickcontrols-changes-qt6.html#custom-styles-are-now-proper-qml-modules

PR #544, opened 2022-09-20
changed c++ version to c++17

annika-oeth:
changed from c++14 to c++17, didn't test because this is a necessary update


ldanzinger (approved):
LGTM

PR #557, opened 2022-09-28
added param to func

mrwillyees:
in qt6 when using js functions that pass named parameters, it is needed to explicitly name them. I recall this is the chosen style when defining the handler function:
`onEvent: parameter => {
}`

The warning outputted is similar as this: `Injection of parameters into signal handlers is deprecated. Use JavaScript functions with formal parameters instead.`


mfeigl (approved):
Yes, correct - I don't know how I missed that earlier; nicely found!

PR #559, opened 2022-09-28
Removing ColorOverlay as it's not supported in Qt6

mfeigl:
Removing `ColorOverlay` as it is not supported out of the box in Qt 6. These changes are minimal, aesthetic, and sub-perceptual - although it's sad to see them go, we can live without them.

Tested both Cpp and Qml example apps. 

@ldanzinger please review.

### DIFF
--- a/calcite-qml/Calcite/CheckIndicator.qml
+++ b/calcite-qml/Calcite/CheckIndicator.qml
@@ -14,7 +14,6 @@
  *  limitations under the License.
  ******************************************************************************/
 import QtQuick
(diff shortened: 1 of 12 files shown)


mrwillyees (approved):
lgtm. If we want to get the coloring without the ColorOverlay, for simple cases we could modify the `svg` directly and create a dark/bright svgs. For example if you open the pause.svg and add inside the first `path` tag -> `fill="#ff0000"`, you will see it coloring the pause bars of that red color. Just an idea!

PR #564, opened 2022-09-30
Awilliams/register 200

mrwillyees:
Modified the registering template functions to register qml components and controllers both 100.10 and 200.0 ArcGISRuntime.Toolkit.
In this way, a user can import:
Esri.ArcGISRuntime.Toolkit 100 / some of the 100.10 till 100.15 / 200 / 200.0
Esri.ArcGISRuntime.Toolkit.Controller 100 / 100.10 till 100.15 / 200 / 200.0
I had to hardcode some things since now we know that 100.10->100.15 are the only 100 minor versions supported from Toolkit.

I followed similar process as we did in the qt home team: register everything for first 100 version (100.10) and for first 200 version (200.0)


uitools/register/Esri/ArcGISRuntime/Toolkit/internal/register_cpp.cpp
```diff
@@ -153,9 +163,17 @@ namespace Esri::ArcGISRuntime::Toolkit {
...
+      constexpr int START_MINOR_VERSION = 10;
+      constexpr int END_MINOR_VERSION = 15;
+      constexpr int START_MAJOR_VERSION = 100;
+      // resiter version 100
+      for (int i = START_MINOR_VERSION; i <= END_MINOR_VERSION; ++i) {
+        qmlRegisterModule(NAMESPACE, START_MAJOR_VERSION, i);
+      }
+      // register version 200 onwards
+      for(int i = 0; i <= VERSION_MINOR; ++i) {
         qmlRegisterModule(NAMESPACE, VERSION_MAJOR, i);
```
JamesMBallard, line 175:
```suggestion
      for (int i = 0; i <= VERSION_MINOR; ++i) 
      {
        qmlRegisterModule(NAMESPACE, VERSION_MAJOR, i);
```

uitools/register/Esri/ArcGISRuntime/Toolkit/internal/register_cpp.cpp
```diff
@@ -153,9 +163,17 @@ namespace Esri::ArcGISRuntime::Toolkit {
...
     void registerModuleRevisions()
     {
-      constexpr int START_VERSION = 10;
-      for (int i = START_VERSION; i <= VERSION_MINOR; ++i)
+      constexpr int START_MINOR_VERSION = 10;
+      constexpr int END_MINOR_VERSION = 15;
+      constexpr int START_MAJOR_VERSION = 100;
+      // resiter version 100
+      for (int i = START_MINOR_VERSION; i <= END_MINOR_VERSION; ++i) {
+        qmlRegisterModule(NAMESPACE, START_MAJOR_VERSION, i);
```
JamesMBallard, line 171:
```suggestion
      for (int i = START_MINOR_VERSION; i <= END_MINOR_VERSION; ++i) 
      {
        qmlRegisterModule(NAMESPACE, START_MAJOR_VERSION, i);
```

JamesMBallard (commented):
Looks pretty good. Have you tested all the import combinations?

uitools/register/Esri/ArcGISRuntime/Toolkit/internal/register_cpp.cpp
```diff
@@ -124,26 +124,28 @@ namespace Esri::ArcGISRuntime::Toolkit {
     /*
      \internal
      \brief Function for registration. Registers the C++ type Foo as
-     Foo in QML with the appropriate version and namespace information.
+     Foo in QML with the 100.10 and 200.0 version.
```
mfeigl, line 127: ❓ Shouldn't this line keep the _"and namespace information."_?

mfeigl (approved):
Right, makes sense to me! Two further questions:
❓ Does this mean that every new tool will need to be registered at both `100.10` and `200.0` in the future?
❓ Does this mean that from now on, while within the 200 series, in this file only the `constexpr int VERSION_MINOR = 0;` will need to be updated at every release?

mrwillyees:
> Looks pretty good. Have you tested all the import combinations?
Yes, for the `Esri.ArcGISRuntime.Toolkit.Controller` I tested all the mentioned combinations.
For `Esri.ArcGISRuntime.Toolkit` all work except a 100.(!=10)
This is because that module is defined by a `qmldir` file which only create a module versioning if there are components for that minor version. If I use the old `qmldir`, they all work because there is at least 1 component that was created in every version of 100. I would keep it that way and add only 200.0, otherwise we have to manually add all the 100 versionings.
From the docs:
```
A version cannot be imported if no types have been explicitly exported for that version. If a module provides a MyButton type in version 1.0 and a MyWindow type in version 1.1, you cannot import version 1.2 or version 2.0 of that module.
```

mfeigl:
> Yes, do you prefer to rename it so it is easier to find when searching for the versioning upgrade?
No, the current name is fine for me

JamesMBallard (approved):

PR #565, opened 2022-10-04
mfeigl/port ar v200

mfeigl:
This is still in progress, but also in a good position to begin the code review, please see checkbox for current status.

- [x] C++ AR builds on desktop
- [x] C++ AR builds and runs on iOS
- [ ] C++ AR builds on Android
- [x] Qml AR builds on desktop
- [x] Qml AR builds and runs on iOS]
- [ ] Qml AR builds on Android


augmentedreality/Examples/CppArExample/CppArExample.pro
```diff
@@ -46,7 +59,7 @@ OTHER_FILES += \
 # The path to the ArcGIS runtime toolkit for Qt sources, corresponding to the files downloaded
 # from the GitHub repo: https://github.com/Esri/arcgis-runtime-toolkit-qt
 
-ARCGIS_TOOLKIT_PATH = # must be set to the root path of the toolkit folder
+ARCGIS_TOOLKIT_PATH = $$PWD/../../../ # must be set to the root path of the toolkit folder
```
JamesMBallard, line 62: Was this intentional?

augmentedreality/Examples/CppArExample/CppArExample.pro
```diff
@@ -21,6 +21,19 @@ CONFIG += c++17
...
+}
+
+equals(QT_MAJOR_VERSION, 6) {
+    lessThan(QT_MINOR_VERSION, 2) {
+        error("This version of the ArcGIS Runtime SDK for Qt requires Qt 6.2.4")
+    }
+  equals(QT_MINOR_VERSION, 2) : lessThan(QT_PATCH_VERSION, 4) {
+    error("This version of the ArcGIS Runtime SDK for Qt requires Qt 6.2.4")
+  }
+}
```
JamesMBallard, line 35:
```suggestion
lessThan(QT_MAJOR_VERSION, 6) {
    error("This version of the ArcGIS Runtime SDK for Qt requires Qt 6")
}
```

🔧 
This is fine but it will create maintenance in the future as we move up Qt versions. For now I think we should keep it simple and stick with one Qt 6 check.

augmentedreality/Common/source/ArcGISArViewInterface.cpp
```diff
@@ -355,16 +355,9 @@ void ArcGISArViewInterface::setPlaneColor(const QColor& planeColor)
...
  */
-void ArcGISArViewInterface::geometryChanged(const QRectF& newGeometry, const QRectF& oldGeometry)
+void ArcGISArViewInterface::geometryChange(const QRectF& newGeometry, const QRectF& oldGeometry)
 {
-  // enable detection of orientation changes.
-  if (window() && window()->screen())
-  {
-    window()->screen()->setOrientationUpdateMask(Qt::LandscapeOrientation	| Qt::PortraitOrientation |
-                                                 Qt::InvertedLandscapeOrientation | Qt::InvertedPortraitOrientation);
-  }
```
JamesMBallard, line 365: Is there any equivalent orientation logic for Qt 6? @GuillaumeBelz any thoughts here?

augmentedreality/Examples/CppArExample/CppArExample.pro
mfeigl: Yes but only for the time being - will revert for the PR and keep it while in draft.

augmentedreality/Examples/CppArExample/CppArExample.pro
mfeigl: I think this will be fine when we move up Qt versions, it is checking 6.2.4 as a minimum. This would only cause problems if we were to support Qt versions lower than 6.2.4

augmentedreality/Common/source/ArcGISArViewInterface.cpp
GuillaumeBelz: `setOrientationUpdateMask` was used to limit the signal `orientationChanged` to a list of selected orientations. By default, the mask was 0, then no signal was emitted. We set the mask to receive the signal when the orientation was changed.

Now, the mask is removed and signal is always emitted.

GuillaumeBelz (approved):

mfeigl:
The desktop implementation builds, the iOS builds and runs, and I have updated the obvious Android changes but kept the ambiguous ones. @JamesMBallard, I re-requested your review.

JamesMBallard (approved):

augmentedreality/Examples/CppArExample/CppArExample.cpp
```diff
@@ -341,7 +341,7 @@ void CppArExample::onMouseClicked(QMouseEvent& event)
 {
   Q_CHECK_PTR(m_arcGISArView);
 
-  const QPoint screenPoint = event.localPos().toPoint();
+  const QPoint screenPoint = event.position().toPoint();
```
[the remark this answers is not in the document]
mfeigl: Sounds good, made the change.

PR #566, opened 2022-10-07
Awilliams/callout deprecations removal

mrwillyees:
Removed deprecated properties from Callout

### DIFF
--- a/uitools/import/Esri/ArcGISRuntime/Toolkit/Callout.qml
+++ b/uitools/import/Esri/ArcGISRuntime/Toolkit/Callout.qml
@@ -289,6 +289,8 @@ Pane {
         background.children.push(shapeTail.createObject())
     }
 
+
+
     contentItem: GridLayout {
         id: calloutLayout
         columns: 3
@@ -317,6 +319,7 @@ Pane {
         Label {
             id: title
             text: calloutData ? calloutData.title : ""
+            color: palette.windowText
             wrapMode: Text.Wrap
             clip: true
             elide: Text.ElideRight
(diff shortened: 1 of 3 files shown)


uitools/import/Esri/ArcGISRuntime/Toolkit/Callout.qml
```diff
@@ -54,11 +54,13 @@ import QtQuick.Shapes
      \newcode
             Callout {
               calloutData: myCalloutData
-              titleTextColor: "#000000"
-              backgroundColor: "#ffffff"
-              borderColor: "#000000"
-              borderWidth: 2
-              cornerRadius: 5
+              palette.windowText: "#000000"
+              background: Rectangle {
```
mfeigl, line 58: ❓ I was just wondering whether a `pane` would make more sense to be used here than `rectangle`?

mfeigl (approved):
LGTM, one question but not a show stopper.

uitools/import/Esri/ArcGISRuntime/Toolkit/Callout.qml
mrwillyees: It would give more functionalities, but I am just overriding the template Callout in calcite which has a Rectangle. I don't think we need anything complicated for the background, and from the Pane docs I cannot find the radius property.

mrwillyees:
I am deferring to you here: in the demo example I have purposely added many properties to show the user when reading the docs (the docs pick up the qml code and show it). They are actually a bit redundant because I might be setting them as the default value, just for the sake of showing off. If you don't like it, remove and mold as you prefer

PR #573, opened 2022-11-29
Using alias to differentiate between widgets and quick map types

mfeigl:


uitools/cpp/Esri/ArcGISRuntime/Toolkit/UtilityNetworkTraceController.cpp
```diff
@@ -30,6 +30,7 @@
 #include <GraphicsOverlayListModel.h>
 #include <IdentifyLayerResult.h>
 #include <Map.h>
+#include <MapQuickView.h>
```
mrwillyees, line 33: I could build the cpp demo app without adding this include

mrwillyees (approved):
lgtm, the include is not needed. Approving

uitools/cpp/Esri/ArcGISRuntime/Toolkit/UtilityNetworkTraceController.cpp
mfeigl: Yes, but if the tool is used for widgets, it would complain about missing `QMouseEvent`

mfeigl:
> lgtm, the include is not needed. Approving

The include is needed for a widgets build.

PR #575, opened 2022-12-05
Mfeigl/changing utility network trace button order

mfeigl:
Changed the Utility Network Trace tool's button order. One (which is now on the left) is permanently displayed, while the other (now on the right) is conditionally displayed, so it made more sense not to have the permanent button jumping based on the condition of the other button.


mrwillyees (approved):
I think you are right, rather unnecessary the left and right movement of the second button

PR #579, opened 2022-12-08
Provide convenience hooks when logging in via UserCredentialsView

PaulAllanSturm:
When using AuthenticationView to login with UserCredentialsView, provide convenience hooks to username and password fields

### DIFF
--- a/uitools/import/Esri/ArcGISRuntime/Toolkit/AuthenticationView.qml
+++ b/uitools/import/Esri/ArcGISRuntime/Toolkit/AuthenticationView.qml
@@ -65,10 +65,6 @@ Item {
     */
     property var controller: AuthenticationController { }
 
-    property var activeLoginView
-
-    signal activeLoginViewReady
-
     Connections {
         target: controller
         function onCurrentChallengeTypeChanged() {
@@ -80,10 +76,7 @@ Item {
                         this.object.open();
 
                         if (component === userCredentialsView)
-                        {
-                            activeLoginView = this.object;
-                            activeLoginViewReady();
-                        }
+                            activeLoginViewReady_(this.object);
                     }
                 }
             }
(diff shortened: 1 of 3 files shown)


GuillaumeBelz (approved):

JamesMBallard (approved):
I think this approach is good, but we need to wait to merge until we finalize the branch for 200.0 release.

Also, we need to get buy-in from @ldanzinger on the general approach.

ldanzinger:
I like it, it is minimally invasive and makes our lives easier for testing. With that said, there are a couple things to consider trying:

- The aliases, though marked with trailing underscore and commented, will still show up as public properties in the IDE intellisense. Not the end of the world, but best to keep this as tight as possible. Instead of putting the property aliases directly at the root of the UserCredentialsView, we could possibly make a QtObject with an id called internal, and we could add internal properties there. We do something somewhat similar in Callout https://github.com/Esri/arcgis-runtime-toolkit-qt/blob/main/uitools/import/Esri/ArcGISRuntime/Toolkit/Callout.qml. Client code would then look more like `authenticationView.internal.prefillUsername`
- We should mark as internal with qdoc command

PaulAllanSturm:
@ldanzinger - I updated it to be as minimal as possible and marked with qdoc commands. Unfortunately with QML, you can't refer to objects by id from a different qml file, so we can't do `authenticationView.internal.prefillUsername` from client code in another file. At best we can have a single function marked qdoc internal.

ldanzinger (approved):
LGTM - thanks for looking into the other idea

PaulAllanSturm:
I added a new commit that added hooks to the OAuth webview.

ldanzinger (approved):
LGTM. I'm ok with the solution for now, as it opens up a lot of potential for better testing. If it becomes brittle, we can rethink it

PR #594, opened 2023-07-19
Bri12415/ar core

bri12415:
I believe we still need to do this synchronously because often when trying this asynch we are met with camera is null type errors, and the app must be reset to use the camera. This was observed on release.

### DIFF
--- a/augmentedreality/Common/source/Android/ArCoreWrapper.cpp
+++ b/augmentedreality/Common/source/Android/ArCoreWrapper.cpp
@@ -304,7 +304,7 @@ void ArCoreWrapper::createArSession()
     loop.exec();
   }
 
-  //If the time is active we timed out
+  //If remaining time is 0 the timer ran out
   if(timeout.remainingTime() == 0)
   {
     emit m_arcGISArView->errorOccurred("ARCore failure", "Failed to access to the camera.");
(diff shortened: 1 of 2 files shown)


augmentedreality/Common/source/Android/ArCoreWrapper.cpp
```diff
@@ -285,20 +284,41 @@ void ArCoreWrapper::createArSession()
...
     return;
 
-  // request camera permission
-  auto checkPermissionResultFuture = QtAndroidPrivate::checkPermission(QtAndroidPrivate::PermissionType::Camera);
-  checkPermissionResultFuture.waitForFinished();
-  const auto checkPermissionResult = checkPermissionResultFuture.result();
-  if (checkPermissionResult != QtAndroidPrivate::PermissionResult::Authorized)
+  QCameraPermission cameraPermission;
+  QEventLoop loop;
+  QTimer timeout;
```
bri12415, line 289: The timer is there to prevent a permanent hang in the loop.exec() due to some sort of hiccup or something. The tradeoff is that if the user returns and selects a permission after the timer has ended, the app will crash after. IMO, we should not use a timer, but I wanted to see others thoughts on this. If for some reason the loop.exec() bugs out on a forever loop, the user will just close and re-open the app.

JamesMBallard (commented):
I would like you to pursue an alternate fix. Putting in timers to block and make async things behave synchronously is the wrong approach. Maybe we need to shuffle some logic around, and probably add an internal permissions manager helper class somewhere. @GuillaumeBelz may have some ideas as well.

PR #595, opened 2023-07-24
block until permissions

bri12415:
After researching other solutions, I believe this is the one we want to use. Ive listed a couple other ideas below and why we cant use them.

**QFuture w/ QFutureWatcher**
Can’t use QFuture, because of this error `W qt.permissions: : Permissions can only be requested from the GUI (main) thread`. The call to `Qt::Concurrent::run` spawns a new thread which is where our permission check would happen

**Semaphore w/ QFutureWatcher**
Same as above

### DIFF
--- a/augmentedreality/Common/include/Android/ArCoreWrapper.h
+++ b/augmentedreality/Common/include/Android/ArCoreWrapper.h
@@ -42,9 +42,8 @@ class ArCoreFrameRenderer;
 class ArCorePointCloudRenderer;
 class ArCorePlaneRenderer;
 
-class ArCoreWrapper : public QObject
+class ArCoreWrapper
 {
-    Q_OBJECT
 
 public:
   ArCoreWrapper(ArcGISArViewInterface* arcGISArView);
--- a/augmentedreality/Common/include/Android/ArCoreWrapper.h
+++ b/augmentedreality/Common/include/Android/ArCoreWrapper.h
@@ -42,8 +42,10 @@ class ArCoreFrameRenderer;
 class ArCorePointCloudRenderer;
 class ArCorePlaneRenderer;
 
-class ArCoreWrapper
+class ArCoreWrapper : public QObject
 {
+    Q_OBJECT
+
 public:
   ArCoreWrapper(ArcGISArViewInterface* arcGISArView);
   ~ArCoreWrapper();
--- a/augmentedreality/Common/source/Android/ArCoreWrapper.cpp
+++ b/augmentedreality/Common/source/Android/ArCoreWrapper.cpp
@@ -289,7 +289,7 @@ void ArCoreWrapper::createArSession()
 
   if (qApp->checkPermission(cameraPermission) == Qt::PermissionStatus::Undetermined)
   {
-    qApp->requestPermission(cameraPermission, [&loop]() {
+    qApp->requestPermission(cameraPermission, &loop, [&loop]() {
       //Connect finishing requestPermission to quitting the blocking loop
       loop.quit();
     });
--- a/augmentedreality/Common/source/Android/ArCoreWrapper.cpp
+++ b/augmentedreality/Common/source/Android/ArCoreWrapper.cpp
@@ -289,7 +289,7 @@ void ArCoreWrapper::createArSession()
 
   if (qApp->checkPermission(cameraPermission) == Qt::PermissionStatus::Undetermined)
   {
-    qApp->requestPermission(cameraPermission, this, [this, &loop]() {
+    qApp->requestPermission(cameraPermission, [&loop]() {
       //Connect finishing requestPermission to quitting the blocking loop
       loop.quit();
     });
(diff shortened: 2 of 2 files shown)


augmentedreality/Common/source/Android/ArCoreWrapper.cpp
```diff
@@ -285,20 +284,27 @@ void ArCoreWrapper::createArSession()
...
+
+  if (qApp->checkPermission(cameraPermission) == Qt::PermissionStatus::Undetermined)
   {
-    auto requestPermissionResultFuture = QtAndroidPrivate::requestPermission(QtAndroidPrivate::PermissionType::Camera);
-    requestPermissionResultFuture.waitForFinished();
-    const auto requestPermissionResult = requestPermissionResultFuture.result();
-    if (requestPermissionResult != QtAndroidPrivate::PermissionResult::Authorized)
-    {
+    qApp->requestPermission(cameraPermission, this, [this, &loop]() {
+      //Connect finishing requestPermission to quitting the blocking loop
```
JamesMBallard, line 293:
```suggestion
    qApp->requestPermission(cameraPermission, this, [this, &loop]() 
    {
      //Connect finishing requestPermission to quitting the blocking loop
```

JamesMBallard (approved):

augmentedreality/Common/include/Android/ArCoreWrapper.h
```diff
@@ -42,8 +42,10 @@ class ArCoreFrameRenderer;
 class ArCorePointCloudRenderer;
 class ArCorePlaneRenderer;
 
-class ArCoreWrapper
+class ArCoreWrapper : public QObject
```
GuillaumeBelz, line 45: Why did you add inheritance from QObject?

GuillaumeBelz: Ok, maybe it's for `requestPermission`?

In this case, as `this` is not used and context is used only for object lifetime, maybe you can use `&loop` as context and don't derive from `QObject`

augmentedreality/Common/source/Android/ArCoreWrapper.cpp
```diff
@@ -285,20 +284,29 @@ void ArCoreWrapper::createArSession()
...
+  QCameraPermission cameraPermission;
+  QEventLoop loop;
+
+  if (qApp->checkPermission(cameraPermission) == Qt::PermissionStatus::Undetermined)
   {
-    auto requestPermissionResultFuture = QtAndroidPrivate::requestPermission(QtAndroidPrivate::PermissionType::Camera);
-    requestPermissionResultFuture.waitForFinished();
-    const auto requestPermissionResult = requestPermissionResultFuture.result();
-    if (requestPermissionResult != QtAndroidPrivate::PermissionResult::Authorized)
+    qApp->requestPermission(cameraPermission, this, [this, &loop]()
```
GuillaumeBelz, line 292:
```suggestion
    qApp->requestPermission(cameraPermission, &loop, [&loop]() 
```

augmentedreality/Common/include/Android/ArCoreWrapper.h
bri12415: Ah I thought I needed that for the call into qApp but it looks like I dont

augmentedreality/Common/source/Android/ArCoreWrapper.cpp
```diff
@@ -289,8 +289,7 @@ void ArCoreWrapper::createArSession()
 
   if (qApp->checkPermission(cameraPermission) == Qt::PermissionStatus::Undetermined)
   {
-    qApp->requestPermission(cameraPermission, this, [this, &loop]() 
-    {
+    qApp->requestPermission(cameraPermission, [&loop]() {
```
GuillaumeBelz, line 292: I guess it's necessary to have a context object in second position https://doc.qt.io/qt-6/qcoreapplication.html#requestPermission-1

`QEventLoop` is a `QObject`, it makes sens to use it as context.

```cpp
qApp->requestPermission(cameraPermission, &loop, [&loop]() {
```

bri12415: It looks like the call without a context worked as well

qcoreapplication.h:165
```
    // requestPermission to a functor or function pointer (without context)
    template <typename Func, std::enable_if_t<
        !QtPrivate::FunctionPointer<Func>::IsPointerToMemberFunction
        && !std::is_same<const char *, Func>::value, bool> = true>
    void requestPermission(const QPermission &permission, Func func)
    {
        requestPermission(permission, nullptr, std::move(func));
    }
```

bri12415: Nevermind, I was looking at the dev branch my mistake

GuillaumeBelz: It was a trap :)

GuillaumeBelz (approved):

PR #600, opened 2023-09-08
James/async updates

JamesMBallard:
Update all toolkit code to new Async pattern.

A few things to point out:
 * Locator tool changes were nontrivial. There was some logic happening with signals that was not trivially replicated with QFuture pattern.
 * UtilityNetwork tool changes similarly nontrivial to port. 

Neither was a rewrite, but it was not a simple copy/paste.


JamesMBallard:
I tested all these tools locally. Testing the UtilityNetwork tool is difficult, but it was performing traces.

aji13187 (approved):
LGTM

PR #603, opened 2023-11-22
removing modeldata usage

ldanzinger:
resolves basemap gallery issues with Qt 6.6


bri12415:
@ldanzinger Can we close this one?

PR #609, opened 2024-03-08
Aji13187/test regression fetch data mmpk

aji13187:
Looks like Qt5 to Qt6 there is a difference in the default style settings, causing the problem here.
Fixed as per Ref:- https://forum.qt.io/topic/143627/problems-with-qml-style-change-from-qt5-to-qt6/3

Not reproducible locally.


ldanzinger:
Should we close this out @aji13187?

aji13187:
> Should we close this out @aji13187?

Yes. I'll close this.

PR #610, opened 2024-03-27
modelData not listModelData

bri12415:
This was mistakenly done by me from https://github.com/Esri/arcgis-maps-sdk-toolkit-qt/pull/605. I had thought that the data structure being referred to by `controller.clientCertificateInfos` was one of our ListModels, but it is not. It is just a QStringList, so we can use the `modelData` keyword.

**Before:**
image

**After:**
image

Thanks @ldanzinger for pointing this out.


JamesMBallard (approved):

ldanzinger:
@bri12415 if you haven't done so already, it might be worth a scan to see if there's other areas where we made the same error?

bri12415:
@ldanzinger will do

PR #614, opened 2024-04-26
Bri12415/cmake toolkit

bri12415:


uitools/toolkitcpp.cmake
```diff
@@ -0,0 +1,118 @@
...
+set(TOOLKITCPP_RESOURCES
+    ${CMAKE_CURRENT_LIST_DIR}/images/esri_arcgisruntime_toolkit_images.qrc
+    ${CMAKE_CURRENT_LIST_DIR}/import/Esri/ArcGISRuntime/Toolkit/esri_arcgisruntime_toolkit_view.qrc)
+
+target_sources(${PROJECT_NAME} PRIVATE
+    ${TOOLKITCPP_SOURCES}
+    ${TOOLKITCPP_HEADERS}
+    ${TOOLKITCPP_RESOURCES})
+
+find_package(Qt6 COMPONENTS REQUIRED quickcontrols2 webview svg quicklayouts)
```
bri12415, line 114: This relies on the `include()` call for the outer `CMakeLists.txt` file being after we call `qt_add_executable()`. We can either try and make this more resilient against where the user puts this `include()`, or we could update our templates to have a comment showing them where to put the `include()`

bri12415: I think ideally the solution we come up with would just be able to be included anywhere.

JamesMBallard: I agree it should be able to be inserted anywhere, within the parameters of idiomatic CMake. I'm not strong with CMake, so there may be normal CMake restrictions about this that are well-known to CMake folks that we don't need to worry about.

bri12415: Looking at it again, it might be best to have this available through a package. I had experimented with that before going with `include()`, but had hit a snag somewhere. But it seems like it would make more sense to do something like this
```
find_package(ArcGISRuntime 200.5.0 COMPONENTS REQUIRED Cpp)
find_package(ArcGISToolkit 1.0.0 something something)
```

rather than this
```
find_package(ArcGISRuntime 200.5.0 COMPONENTS REQUIRED Cpp)
include(<path to cmake>)
```

and then the placement of the `find_package` seems to come naturally.

uitools/toolkitcpp.cmake
```diff
@@ -0,0 +1,78 @@
...
+    ${CPPPATH}/CoordinateOptionDefaults.cpp
+    ${CPPPATH}/FloorFilterController.cpp
+    ${CPPPATH}/FloorFilterFacilityItem.cpp
+    ${CPPPATH}/FloorFilterLevelItem.cpp
+    ${CPPPATH}/FloorFilterSiteItem.cpp
+    ${CPPPATH}/LocatorSearchSource.cpp
+    ${CPPPATH}/Internal/BasemapGalleryImageProvider.cpp
+    ${CPPPATH}/Internal/GenericListModel.cpp
+    ${CPPPATH}/Internal/GenericTableProxyModel.cpp
+    ${CPPPATH}/Internal/MetaElement.cpp
```
ldanzinger, line 42: Is there a way in CMake to basically just add anything in that directly with a wildcard *.cpp?

bri12415: There is this: https://cmake.org/cmake/help/latest/command/file.html#glob
But here is a note from the CMake documentation on its usage

> Note: We do not recommend using GLOB to collect a list of source files from your source tree. If no CMakeLists.txt file changes when a source is added or removed then the generated build system cannot know when to ask CMake to regenerate. The CONFIGURE_DEPENDS flag may not work reliably on all generators, or if a new generator is added in the future that cannot support it, projects using it will be stuck. Even if CONFIGURE_DEPENDS works reliably, there is still a cost to perform the check on every rebuild.

ldanzinger: fair enough, we shouldn't go against best practices

JamesMBallard: In that case we should follow recommend practices and just list all files manually. Unfortunate, but we're not adding/removing lots of files to the toolkit right now.

uitools/toolkitcpp.cmake
```diff
@@ -0,0 +1,88 @@
...
+
+find_package(Qt6 COMPONENTS REQUIRED QuickControls2 WebView Svg QuickLayouts)
+
+target_link_libraries(${PROJECT_NAME} PRIVATE
+    Qt6::QuickControls2
+    Qt6::WebView
+    Qt6::Svg)
+
+if(IOS)
+    target_link_libraries(${PROJECT_NAME} PRIVATE Qt6::qquicklayoutsplugin)
```
GuillaumeBelz, line 79: `qquicklayoutsplugin` is in lower case, the rest of Qt6 libs use upper cases. Is it correct?

bri12415: Yeah I don't know why this deviates from the norm, but this is the correct name.

uitools/toolkitcpp.cmake
```diff
@@ -0,0 +1,88 @@
...
+else()
+    target_link_libraries(${PROJECT_NAME} PRIVATE Qt6::QuickLayouts)
+endif()
```
JamesMBallard, line 82: Can you help me understand why this is needed at all? If we are using QuickLayouts, shouldn't it be added to `target_link_libraries` above?

bri12415: Good question, I haven't been able to get to the bottom of this yet.

This is what the robot tells me:

> The Qt6::qquicklayoutsplugin is a static version of the Qt6::QuickLayouts library. It's used on iOS because of the static linking requirement. On other platforms, the dynamic Qt6::QuickLayouts library is used instead.

And this is what Qt has to say (https://doc.qt.io/qt-6/porting-to-ios.html)
> In Qt for iOS, everything is compiled statically and placed into the application bundle. The applications are "sandboxed" inside their bundles and cannot make use of shared object files. Because of this, also the plugins used by the Qt modules need to be statically linked.

But that doesn't really answer why this is the only library that needs to be handled this way.

JamesMBallard: I think it's ok that we keep it for now, but it feels like the kinda thing that will come back to haunt us at some point with future Qt versions.

JamesMBallard:
Between the two approaches you outlined, is one the more preferred CMake approach for what we're doing?

This sounds reasonable to me. 
```cmake
# Uncomment this line to include the ArcGIS Toolkit for Qt
include(<path to toolkitcpp.cmake>)
```

bri12415:
> Between the two approaches you outlined, is one the more preferred CMake approach for what we're doing?
> 
> This sounds reasonable to me.
> 
> ```cmake
> # Uncomment this line to include the ArcGIS Toolkit for Qt
> include(<path to toolkitcpp.cmake>)
> ```

Based on this discussion here, it seems like the `include` is the way to go imo, but we may need some include guards
https://discourse.cmake.org/t/include-vs-find-package-rules-of-thumb/6713

JamesMBallard:
Ok, so is anything else needed in this PR?

bri12415:
Just verification on linux which @aji13187 is helping me with (thanks again). Ill also deploy to mobile from windows just to double check

JamesMBallard:
Please retarget this PR to the v.next branch. We can discuss backporting it to main later after all our internal testing is good.

bri12415:
Right, will do. My mistake.

JamesMBallard (approved):
LGTM as long as all platform testing looks good.

bri12415:
Sorry for the re-request @JamesMBallard , I forgot about widgets.

uitools/toolkitwidgets.cmake
```diff
@@ -0,0 +1,66 @@
...
+
+target_include_directories(${PROJECT_NAME} PRIVATE 
+    ${CMAKE_CURRENT_LIST_DIR}/widgets
+    ${WIDGETPATH})
+
+target_sources(${PROJECT_NAME} PRIVATE
+    ${TOOLKITCOMMON_SOURCES}
+    ${TOOLKITWIDGET_SOURCES}
+    ${TOOLKITWIDGET_RESOURCES}
+    ${TOOLKITWIDGET_FORMS})
```
JamesMBallard, line 57: It's been a while since I worked with Widgets, but I am surprised that the .ui files would appear in the `target_sources` list. 
This is a separate list in qmake via `FORMS`. I assume it's tested, but it seems different.

JamesMBallard (commented):
Changes all look good so far. There are some readme updates needed as well for the new CMake workflows. Were you planning those in this PR or a follow up?

uitools/toolkitwidgets.cmake
bri12415: I agree it is strange, but this is how Qt has it on their page: https://doc.qt.io/qt-6/designer-using-a-ui-file.html#compile-time-form-processing

specifically for our widgets:
> The form is listed among the C++ source files in qt_add_executable(). The option CMAKE_AUTOUIC tells CMake to run the uic tool to create a ui_calculatorform.h file that can be used by the source files.

JamesMBallard: Good to know, thanks.

PR #617, opened 2024-05-07
Better CMake practices

bri12415:
Turn toolkitcpp.cmake and toolkitwidgets.cmake into function includes.


uitools/common.cmake
```diff
@@ -13,6 +13,43 @@
 # limitations under the License.
 set(CPPPATH ${CMAKE_CURRENT_LIST_DIR}/cpp/Esri/ArcGISRuntime/Toolkit)
 
+set(TOOLKITCOMMON_HEADERS
```
GuillaumeBelz, line 16: I think a good practice with cmake is to have a `CMakeLists.txt` for each directory and `add_subdirectory` to add them to the root `CMakeLists.txt`.

bri12415: So does this mean we should replace
```
# Uncomment this line to add the toolkit to your project
# See https://github.com/Esri/arcgis-maps-sdk-toolkit-qt for details
# include(path/to/toolkitcpp.cmake)
```

with

```
# Uncomment this line to add the toolkit to your project
# See https://github.com/Esri/arcgis-maps-sdk-toolkit-qt for details
# add_subdirectory(path/to/CMakeLists.txt /path/to/buildoutput)
```

JamesMBallard: If this is the better workflow then we should use it, however we cannot have two different CMakeLists.txt in the same directory for Qt Quick and Widgets, so we'll have to change things up for that.

JamesMBallard: >So does this mean we should also replace

Yes if that's what we're doing.

bri12415:
After talking to Frank Miller, he pointed out a few things:

* `add_subdirectory` is intended for use with multiple targets/libraries. That's not really what we are trying to achieve here.

* If we just define a function inside of `toolkitcpp.cmake`, the user can include `toolkitcpp.cmake` anywhere, and call it after they set up their target.

* This would also allow users to specify which target they want to add the toolkit functionality to. Previously it was just assumed this was ${PROJECT_NAME}, but this could differ based on user needs. 

* According to Frank, this is a common CMake workflow and would be familiar to CMake users.

GuillaumeBelz:
Is there a doc for users about how to use these cmake files? Or an issue for that?

GuillaumeBelz (approved):
looks ok

bri12415:
Yeah, its here https://github.com/Esri/arcgis-maps-sdk-toolkit-qt/blob/v.next/uitools/README.md. I'll need to update this and the templated CMake file for our template apps, I just wanted those changes to go in at the same time but I can have the changes to the toolkit doc on this one.

uitools/common.cmake
```diff
@@ -11,7 +11,44 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-set(CPPPATH ${CMAKE_CURRENT_LIST_DIR}/cpp/Esri/ArcGISRuntime/Toolkit)
+set(CPPPATH ${TOOLKITUI_DIR}/cpp/Esri/ArcGISRuntime/Toolkit)
```
cfrankmiller, line 14: Both places where this file is included has this variable already set and this line is setting it again in the parent context. Consider doing the following instead to prevent surprises.

```
if(NOT CPPPATH)
  message(FATAL_ERROR "CPPPATH is expected to be defined")
endif()
```

Alternatively, instead of including inline, you could consider defining a function that takes out variables instead. Something like this

```
function(set_common_vars headers sources)
  set(CPPPATH ${TOOLKITUI_DIR}/cpp/Esri/ArcGISRuntime/Toolkit)
  set(${headers}
    ${CPPPATH}/AuthenticationController.h
    ${CPPPATH}/BasemapGalleryController.h
    # ...
    PARENT_SCOPE
  )
  set(${sources}
    # ...
    PARENT_SCOPE
  )
function()
```

bri12415: Thanks for pointing that out, I ended up going with your second suggestion.

uitools/toolkitwidgets.cmake
```diff
@@ -11,56 +11,165 @@
...
-target_sources(${PROJECT_NAME} PRIVATE
-    ${TOOLKITCOMMON_SOURCES}
-    ${TOOLKITWIDGET_SOURCES}
-    ${TOOLKITWIDGET_RESOURCES}
-    ${TOOLKITWIDGET_FORMS})
+    target_compile_definitions(${TARGET_NAME} PRIVATE WIDGETS_ARCGISRUNTIME_TOOLKIT)
+endfunction()
 
-find_package(Qt6 COMPONENTS REQUIRED Widgets WebEngineWidgets Svg)
+function(setup_common_vars HEADERS SOURCES)
```
bri12415, line 47: I know this is repeated in both `toolkitwidgets.cmake` and `toolkitcpp.cmake`, I wanted to put this in `common.cmake`, but then the `setup_common_vars` function would either need to be passed the CPP_PATH as an argument, or rely on CPP_PATH being defined by the times its called in its parent scope.

uitools/toolkitcpp.cmake
```diff
@@ -11,42 +11,134 @@
...
-    Qt6::WebView
-    Qt6::Svg)
+    list(APPEND QML_IMPORT_PATH ${TOOLKITUI_DIR}/import)
 
-if(IOS)
-    target_link_libraries(${PROJECT_NAME} PRIVATE Qt6::qquicklayoutsplugin)
-else()
-    target_link_libraries(${PROJECT_NAME} PRIVATE Qt6::QuickLayouts)
-endif()
+    set(QML_IMPORT_PATH ${QML_IMPORT_PATH} CACHE STRING "" FORCE)
```
cfrankmiller, line 51: There may be a problem here. I believe that each time cmake is run the path will be appended again and this cache variable and will continue to grow.

It's also possible that the user could stomp on what we set if they follow [the documentation](https://doc.qt.io/qtcreator/creator-qml-modules-with-plugins.html#importing-qml-modules).

I'm curious to know more about this QML_IMPORT_PATH cache variable. The documentation makes it sound like its only for Qt Creator. Is that correct?

bri12415: Good point, originally i had thought this was needed but I think I was incorrect.  It looks like this may be handled in our `Esri::ArcGISRuntime::Toolkit::registerComponents` function that we require users to call. So it may be the case that I can just get rid of these two lines.

I am not sure if the QML_IMPORT_PATH variable is exclusively used by the Qt Creator IDE, but I know that it is just another avenue to tell the QML engine where to look for modules you are importing, and where it can find the relevant `qmldir` folder for that module.

uitools/toolkitwidgets.cmake
bri12415: Ok I see what you are saying. I think it is simplest to just list the files necessary, no reason to overcomplicate.

frankmiller (approved):

bri12415:
@JamesMBallard Whenever you have time, there were some changes to the cmake file and the docs.

JamesMBallard (approved):

PR #619, opened 2024-05-22
Bri12415/toolkit libary cmake

bri12415:
Qt recommends CMake's `add_subdirectory` in place of QMake's `include(*.pri)` (see [here](https://wiki.qt.io/CMake_Port/Porting_Guide), at the bottom of the page.). `add_subdirectory` works by locating a `CMakeLists.txt` file within the provided subdirectory ([doc](https://cmake.org/cmake/help/latest/command/add_subdirectory.html)).

Because of this, we have split the cpp and the widgets toolkits into two separate subdirectories of `uitools`: `toolkitcpp` and `toolkitwidgets`. Each of these has their own `CMakeLists.txt` file at their root directory. The overall directory structure looks like this now (excluding README.md and the like for readability)
```
arcgis-toolkit-qt
├── uitools/
│   ├── examples/
│   ├── toolkitcpp/
|   |   ├── CMakeLists.txt
|   |   ├── cpp/
|   |   ├── images/
|   |   ├── import/
|   |   ├── register/
│   ├── toolkitwidgets/
|   |   ├── CMakeLists.txt
|   |   ├── cpp/
|   |   ├── images/
|   |   ├── widgets/

```

There is duplication of the `cpp` and `images` directory as both `toolkitcpp` and `toolkitwidgets` relies on these files. This was something I couldn't figure out how to avoid while also supporting two separate copyable subdirectories. 

The workflow for Cmake would look like this
```
cp -r /path/to/toolkitcpp /path/to/project/ // <--- copy the directory to your project's directory
```

```
qt_add_executable(${PROJECT_NAME} ${SOURCE_FILES})

add_subdirectory(toolkitcpp) // <--- add the subdirectory to your project

target_link_libraries(${PROJECT_NAME} PRIVATE
  Qt6::Core
  Qt6::Quick
  Qt6::Multimedia
  Qt6::Positioning
  Qt6::Sensors
  Qt6::WebSockets
  ArcGISRuntime::Cpp
  libtoolkitcpp) // <--- link to the generated toolkit library

```

The workflow for qmake does not change.
```
ARCGIS_RUNTIME_VERSION = 200.5.0
include($$PWD/arcgisruntime.pri)
include(path/to/toolkitcpp.pri)

```

Both workflows still require the user calling `Esri::ArcGISRuntime::Toolkit::registerComponents()` from their `main.cpp` as before.

Once this is merged, I think a good follow up would be to CMake-ify the toolkit examples we provide.

Note: I intentionally did not update the README.md yet to avoid further bloat of this PR.


JamesMBallard:
>This means that users will need to copy uitools into their projects directory

I did not realize this was also required. Is there any way to reference this from a relative path? For example an out-of-source directory. There seems to be [mixed success](https://www.google.com/search?q=cmake+add_subdirectory+out+of+tree) with this.

bri12415:
@GuillaumeBelz @JamesMBallard This is ready now whenever you all have time.

uitools/toolkitwidgets/images/docs/bookmarksView.gif
JamesMBallard, line 1: I assume these types of things are just for doc and won't be embedded into any projects. Can we at least centralize these so we don't need to duplicate them? If that would be another big piece of work, we can live with the duplication.

bri12415:
After scrum today, we decided to try and pull the `toolkitcpp.pri` and `toolkitwidgets.pri` into their original places. This way, users wont be met with an error if they pull these changes and are referencing `toolkit*.pri` in their applications at its original location.

JamesMBallard (approved):

GuillaumeBelz (approved):

PR #623, opened 2024-06-10
Toolkit examples minimum Qt version is set

aji13187:
Minimum Qt version required for toolkit examples are set.


JamesMBallard (approved):
lgtm, please coordinate merging with dev ops.

PR #641, opened 2024-08-30
New authentication OAuth prototype

JamesMBallard:
## High level design goals

1. Any code using `AuthenticationView` has no changes. All existing code still works. Even source file names and location are unchanged.
2. New abstraction added called `ArcGISAuthenticationView`. If users declare this in QML, it will determine which authentication system is being used and dispatch to either the "new" system, or the "existing" system.
3. Allows for iterative changes and improvements without breaking any existing client code.

## Details of new logic

1. New ArcGISAuthenticationController class that handles the "new" authentication workflows. This has internal logic that will no-op certain things if the legacy system is active.
2. ArcGISAuthenticationView has a Loader pattern to determine which authentication system is in place and loads the proper one.
3. ArcGISAuthenticationView_internal.qml contains all the new authentication dialog logic. Nothing is shared with the existing classes.
4. The toolkit handles and processes OAuth
5. In QML the `ArcGISAuthenticationController` is a singleton. There should only be one, so creating multiple would be wrong. This is the component that is registered as the 'handler' of authentication challenges.


JamesMBallard:
@GuillaumeBelz @ldanzinger please take a look. This is for early review to see if we are generally happy with the pattern.

JamesMBallard:
We have more naming concerns. I went with new classes prefixed with "ArcGIS", but there is some concern that this will cause confusion when we speak more broadly about "ArcGIS" authentication vs "Network-tier" authentication.

ldanzinger:
@JamesMBallard only scanned the code and your notes. I like the approach in that you don't break the existing workflows and you have a clear way to add the new pattern (without having to make the new code jump through special hoops). 

> We have more naming concerns. I went with new classes prefixed with "ArcGIS", but there is some concern that this will cause confusion when we speak more broadly about "ArcGIS" authentication vs "Network-tier" authentication.

This is the only thing that jumped out to me. I reviewed the code before reading that comment, and I initially assumed that there was a base AuthenticationView, and then there would be an ArcGISAuthenticationView and a NetworkTierAuthenticationView, or something along those lines. I think it's good for now, and we can fine tune the name later

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/ArcGISAuthenticationView_internal.qml
```diff
@@ -0,0 +1,115 @@
...
+    //             target: controller
+    //             function onCurrentChallengeTypeChanged() {
+    //                 reject();
+    //             }
+    //         }
+    //         onClosed: {
+    //             this.destroy();
+    //         }
+    //     }
+    // }
```
JamesMBallard, line 93: This is still pending in our new implementation. Coming soon, but I wanted to leave it in.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/ArcGISAuthenticationController.cpp
```diff
@@ -0,0 +1,263 @@
...
+  });
+}
+
+void ArcGISAuthenticationController::respond(const QUrl& url)
+{
+  if (!m_currentOAuthUserLoginPrompt)
+    return;
+
+  m_currentOAuthUserLoginPrompt->respond(url);
+  m_currentOAuthUserLoginPrompt->deleteLater();
```
GuillaumeBelz, line 166: Is it possible to use `unique_ptr` or it's required to use `deleteLater`? If no, is it documented?

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/ArcGISAuthenticationController.cpp
```diff
@@ -0,0 +1,263 @@
...
+/*!
+  \brief Destructor.
+ */
+ArcGISAuthenticationController::~ArcGISAuthenticationController() = default;
+
+ArcGISAuthenticationController* ArcGISAuthenticationController::create(QQmlEngine* qmlEngine, QJSEngine* /*jsEngine*/)
+{
+  static QPointer<ArcGISAuthenticationController> instance = new ArcGISAuthenticationController(qmlEngine);
+
+  if (!instance)
```
GuillaumeBelz, line 83: How this condition can be true? Is `instance` can be deleted somewhere?

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/ArcGISAuthenticationController.cpp
```diff
@@ -0,0 +1,263 @@
...
+}
+
+ArcGISAuthenticationController* ArcGISAuthenticationController::instance()
+{
+  return create(nullptr, nullptr);
+}
+
+void ArcGISAuthenticationController::handleArcGISAuthenticationChallenge(ArcGISAuthenticationChallenge* challenge)
+{
+  std::lock_guard<std::mutex> lock(m_mutex);
```
GuillaumeBelz, line 96: Why a mutex for `handleArcGISAuthenticationChallenge`? Is it required by this specific implementation in toolkit or this should be done all the time when implementing `handleArcGISAuthenticationChallenge`? Is it documented somewhere?

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/ArcGISAuthenticationController.cpp
```diff
@@ -0,0 +1,263 @@
...
+        m_currentChallenge->continueAndFailWithError(e.error());
+        m_currentChallenge->deleteLater();
+        m_currentChallenge = nullptr;
+      });
+
+      return;
+    }
+  }
+
+  emit displayUsernamePasswordSignInView();
```
GuillaumeBelz, line 128: What is the role of this signal? It's not emitted if one user configuration is used for request URL. Is it expected?

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/ArcGISAuthenticationController.cpp
```diff
@@ -0,0 +1,263 @@
...
+  {
+    if (userConfiguration->canBeUsedForUrl(requestUrl))
+    {
+      OAuthUserCredential::createAsync(userConfiguration, this).then(this, [this](OAuthUserCredential* credential)
+      {
+        if (!m_currentChallenge)
+          return;
+
+        m_currentChallenge->continueWithCredential(credential);
+        m_currentChallenge->deleteLater();
```
GuillaumeBelz, line 112: Same question about `unique_ptr` here.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/ArcGISAuthenticationController.cpp
JamesMBallard: Either one works. Do you think std::unique_ptr would be preferable? We can remove any parent that's set on this and just use unique_ptr.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/ArcGISAuthenticationController.cpp
JamesMBallard: This is defensive code, but in years past we ran into situations where our QML API singletons would eventually get deleted. This is just in case the same thing happens, but I have no indication it will in the toolkit and I've not run into it yet.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/ArcGISAuthenticationController.cpp
JamesMBallard: It is because users can set `m_userConfigurations` (indirectly via setter). It is highly unlikely both could happen at the same time, but I wanted the code to be safe and prevent any data races.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/ArcGISAuthenticationController.cpp
GuillaumeBelz: Here's my concerns. 

`this` is set as context parameter in `then` and `onFailed`. According to the documentation https://doc.qt.io/qt-6/qfuture.html#then-3:

> If the context is destroyed before the chain has finished, the future is canceled.

That means the future can be canceled immediately when the controller is deleted. And in this case, `then` and `onFailed` may not be called and `m_currentChallenge` is not deleted. It looks like it's a memory leak.

To avoid that, I guess it's necessary to do:
- set controller as parent of challenge
- or add `onCancelled` to clean up challenge pointer (and other raw pointers in controller)
- clean up pointers in controller's destructor
- use `unique_ptr`

It's not clear if the controller can be deleted or not. It's implemented as a singleton. But there is a check for the pointer validity in `instance()` so I assume it can be deleted.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/ArcGISAuthenticationController.cpp
JamesMBallard: Currently there are two supported workflows:
1. Token (username/password dialog).
2. OAuth.

Users have to opt-in to OAuth as you know, but we prefer OAuth if users have provided any `OAuthUserConfiguration` so that is checked first. If there are no matching `OAuthUserConfiguration`, then we fall through to username/password dialog - and that's what this signal is for.

>It's not emitted if one user configuration is used for request URL. Is it expected?

Yes, it is expected. `ArcGISAuthenticationManager::oAuthUserLoginPromptIssued` will emit from the API, and the toolkit code above will do `emit displayOAuthSignInView();`.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/ArcGISAuthenticationController.cpp
JamesMBallard: I see your point and I think it would be better as a std::unique_ptr. I will make that change and retest.

uitools/toolkitcpp/CMakeLists.txt
```diff
@@ -73,6 +73,7 @@ target_include_directories(libtoolkitcpp PUBLIC
 
 target_compile_definitions(libtoolkitcpp PRIVATE
     CPP_ARCGISRUNTIME_TOOLKIT
+    MAPS_SDK_QT_TOOLKIT
```
bri12415, line 76: What is the purpose of this define (here and the pri file)? I don't see it referenced in this PR or in the toolkit, maybe I'm just not seeing it somewhere.

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/ArcGISAuthenticationView_internal.qml
```diff
@@ -0,0 +1,115 @@
...
+            displayView(userCredentialsView)
+        }
+
+        function onDisplayOAuthSignInView() {
+            displayView(oAuth2View)
+        }
+    }
+
+    Component {
+        id: userCredentialsView
```
GuillaumeBelz, line 54: Maybe add `Component` suffix in `id`, as it's a component, not a real view?

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/ArcGISAuthenticationView_internal.qml
```diff
@@ -0,0 +1,115 @@
...
+            onClosed: {
+                this.destroy();
+            }
+            Component.onCompleted: activeLoginViewReady_(this)
+        }
+    }
+
+    Component {
+        id: oAuth2View
+        ArcGISOAuth2View {
```
GuillaumeBelz, line 67: Same about `Component` suffix

uitools/toolkitcpp/CMakeLists.txt
JamesMBallard: I am attempting to indicate that `OAuthUserLoginPrompt` is an internal implementation detail of the Maps SDK and shouldn't be used directly in client code. In order to do that, I have this at the top of the file:
```c++
#if !defined MAPS_SDK_QT_TOOLKIT && !defined MAPS_SDK_API_BUILD
#pragma message("OAuthUserLoginPrompt.h is not intended to be used directly. API may change at any time.")
#endif
```

So the toolkit and our API build will identify themselves as known consumers, exempt from the ominous build warning.

JamesMBallard:
@bri12415 @GuillaumeBelz all comments are addressed. Can you please take another look?

bri12415:
> @bri12415 @GuillaumeBelz all comments are addressed. Can you please take another look?

yes sorry, I missed the re-request.

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/ArcGISAuthenticationView_internal.qml
```diff
@@ -0,0 +1,102 @@
...
+    //                 reject();
+    //             }
+    //         }
+    //         onClosed: {
+    //             this.destroy();
+    //         }
+    //     }
+    // }
+
+    signal activeLoginViewReady_(var activeLoginView)
```
bri12415, line 86: Do you think it would improve readability at all if we moved the signal declaration towards the top of this QML object (and others)? It just seems off to me that we reference the signal in some of the Component parts before we actually declare it.

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/ArcGISUserCredentialsView.qml
```diff
@@ -0,0 +1,143 @@
...
+        }
+
+        Button {
+            text: qsTr("Continue")
+            enabled: hasValidInput()
+            DialogButtonBox.buttonRole: DialogButtonBox.AcceptRole
+        }
+
+        onAccepted: {
+            acceptWithCurrentUsernameAndPassword();
```
bri12415, line 51: Which `accept` signal would this be responding to? An implicit signal emitted by the `DialogButtonBox`? I'm just confused because I see that this function also emits the `accept` signal.

bri12415 (approved):
Just some questions I am slightly confused on.

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/ArcGISUserCredentialsView.qml
```diff
@@ -0,0 +1,143 @@
...
+    function prefillUsernameAndPassword(username, password) {
+        usernameTextField.text = username;
+        passwordTextField.text = password;
+    }
+
+    /*!
+      \internal
+      \brief Fill and accept the username and password
+     */
+    function applyUsernameAndPassword(username, password) {
```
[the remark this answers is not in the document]
JamesMBallard: You are right, this is wrong. The accept logic I ported from the existing UI control doesn't work the same with the new workflow, so I had to make some small tweaks. You are right that it isn't needed here, but we need to call accept in at least one exit point, otherwise the dialogs keep piling up on top of each other and aren't dismissed.

I will fix this and rerun all tests. Good catch.

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/ArcGISAuthenticationView_internal.qml
JamesMBallard: We sure can. I mimicked what AuthenticationView.qml does, but it's at the bottom of that file to attempt to separate it from the public part of the API. This entire file is documented as internal, so I will apply your suggestion. I think that will help readability.

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/ArcGISUserCredentialsView.qml
JamesMBallard: It's reacting to `DialogButtonBox` since that's the component it's within, and there is no `onAccepted` in the outer scope for the `Dialog`. It does call acceptWithCurrentUsernameAndPassword, which will call `accept` on the Dialog (coming soon in a commit), which is how it dismisses itself once credentials are supplied.

Also, this is identical to the existing [UserCredentialsView](https://github.com/Esri/arcgis-maps-sdk-toolkit-qt/blob/main/uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/UserCredentialsView.qml#L50-L52).

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/ArcGISAuthenticationView.qml
```diff
@@ -0,0 +1,65 @@
...
+
+    Connections {
+        target: viewLoader.item
+        ignoreUnknownSignals: true
+        function onActiveLoginViewReady_(activeLoginView) {
+            activeLoginViewReady_(activeLoginView);
+        }
+    }
+
+    signal activeLoginViewReady_(var activeLoginView)
```
bri12415, line 64: Do you mind pulling this one up as well like it is in `ArcGISAuthenticationView_internal.qml`?

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/ArcGISUserCredentialsView.qml
bri12415: Okay I understand now, thanks for the explanation. So the eventual `accept` that comes from `acceptWithCurrentUsernameAndPassword` closes the `userCredentialsView` Dialog.

bri12415 (approved):

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/ArcGISAuthenticationView.qml
JamesMBallard: Yeah, missed that one sorry. Done.

GuillaumeBelz (approved):

PR #643, opened 2024-09-03
Cherry picking v.next readme fix to main

jared-2016:
Any objection to cherry picking this fix over to main?


JamesMBallard (approved):
No objection

PR #644, opened 2024-10-01
Fix Ar example build errors

bri12415:
Add `!` operator for throwing error when include of `ArApi.pri` fails.


augmented_reality/Examples/main.cpp
```diff
@@ -1,3 +1,4 @@
+
```
GuillaumeBelz, line 1: Remove empty line

PR #645, opened 2024-10-03
Jare8800/popup elements reuse popup view

jared-2016:
I have created a branch off the latest from v.next called `popupElementsWIP` as a place to stage changes for the PopopView based on PopupElements. We discussed and agreed this is probably the better route rather than merging WIP into v.next then having to pull it out before the release as we want this to be complete before merging into v.next offically.

This PR introduces some new properties to PopupView and the backing functions to the PopupViewController.
Originally I just started off with a new PopupView2 and controller as a sandbox environment for getting the intial prototype changes in. Once things started to iron out I came up with a proposal to support both the Existing PopupView with PopupManager as well as with Popup/PopupElements.

Highlevel of what these changes do:

The order of precedence for which PopupView experience you will get is as follows:
If you set only `popup` on PopupView, the PopupView supporting the new PopupElements will be used
If you set only `popupManager` on PopupView, the exisiting PopupView suppporting popupManagers will be used.
if you set both, a warning will be logged in the console but will default to the newer approach of PopupView supporting PopupElements.

The PopupViewController takes the popup and calles `evaluateExpressionsAsync` on the popup then goes through the list of popups via popup->evaluatedElements() and determines which `PopupElementType` each is and constructs a `TextPopupElementViewController` and adds it to our GenericListModel.

Note: In doing my final pass/cleanup right before creating this PR I noticed a significant oversight with the use of GenericListModel as I instantiate it as `new GenericListModel(&TextPopupElementViewController::staticMetaObject, this)` which would not make it compatible with any additional Controllers for the new PopupViews. I am currently trying to find a solution to this but I didn't want to hold up this PR just for that as I could use some feedback on what might be the best course of action in this situation.

Once the ListModel is Populated with the Controllers that model is used to create a list view which depending on the `PopupElementType` load the correct component and pass that controller to the view and the view is created.

In the case of this PR the only view is the TextElementPopupView which contains a Text component in RIchText format which is populated from the `PopupElements::text()` property. If a click is register, it checks `Text. hoveredLink` to see if there was a link in embedded in the text at the location of the click and if there is, it will open it on the system default browser.

There were some other additional changes to PopupView such as switching the NOTIFY signal on text to it's own signal and adding in some logic to determine which title to use if both Popup and PopupManager were provided.

https://github.com/user-attachments/assets/1c0e5ec7-3252-4d79-8a77-1ff7f2f87d0a


augmented_reality/Examples/ArExample.cpp
```diff
@@ -1,3 +1,4 @@
+
```
jared-2016, line 1: I think this is our git hooks doing this, as well as some other funky stuff with calling our Headers STL headers since in toolkit it looks like we use `< ... >` instead of ` "..." `

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/PopupViewController.cpp
```diff
@@ -43,7 +53,8 @@ namespace Esri::ArcGISRuntime::Toolkit {
  \endlist
  */
 PopupViewController::PopupViewController(QObject* parent):
-  QObject(parent)
+  QObject(parent),
+  m_popupElementsModel(new GenericListModel(&TextPopupElementViewController::staticMetaObject, this))
```
jared-2016, line 57: The problem I mentioned in the PR description 😞

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/TextPopupElementViewController.cpp
```diff
@@ -0,0 +1,123 @@
...
+      \li \c 0
+  \row
+      \li \c FieldsPopupElement
+      \li \c 1
+  \row
+      \li \c MediaPopupElement
+      \li \c 2
+  \row
+      \li \c AttachmentsPopupElement
+      \li \c 3
```
jared-2016, line 78: There are 8 total including unknown in PopupTypes but I think we are only creating popupviews for 5 types being:
AttachmentsPopupElement, FieldsPopupElement, MediaPopupElement, TextPopupElement, ExpressionPopupElement

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/TextPopupElementViewController.cpp
```diff
@@ -0,0 +1,123 @@
...
+ */
+TextPopupElementViewController::TextPopupElementViewController(QObject *parent)
+    : QObject{parent}
+{
+}
+
+  /*!
+  \brief Constructor. Takes a \a PopupElement and \a parent object.
+ */
+TextPopupElementViewController::TextPopupElementViewController(QPointer<PopupElement> textPopupElement, QObject *parent) :
```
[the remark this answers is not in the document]
jared-2016: Agreed it's expecting `TextPopupElement` so we should enforce that on creation.

JamesMBallard:
>The order of precedence for which PopupView experience you will get is as follows:

Presumably this came up in early design discussions, but why try to make both systems work with the same PopupView and PopupViewController? Would it make sense to split out a new PopupElementViewController and PopupElementView to promote the new workflows? That way we can more easily inform users about the newer type(s), and deprecate the existing PopupView classes when the time comes.

ldanzinger:
You can make arguments either way, but for me, putting it in the same type has a few benefits:

* User code barely changes

Old way (pseudocode):
```
var popup = identifyLayerResult.popups[0];
var manager = new PopupManager(popup);
popupView.popupManager = manager;
```

New way (pseudocode):
```
var popup = identifyLayerResult.popups[0];
popupView.popup = popup;
```

* while the new popup system/spec is completely different, they shouldn't need to be heavily invested in the details of how it all works internally and what the differences are between the old way and the new way. Just give us the Popup and we will figure it out from there
* We didn't have a good name for the new type (PopupView is what we would want)
* We can deprecate the popupManager property along with PopupManager in the API, and remove it when the time comes

JamesMBallard:
Ok, sounds good.

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/PopupView.qml
```diff
@@ -131,91 +139,140 @@ Page {
...
+            model: controller.popupElements
+            spacing: 10
+            clip: true
+
+            delegate: Item {
+
+                height: loader.item ? loader.item.height : 200
+
+                Component.onCompleted: {
+                    if (model.popupElementType === 0) {
```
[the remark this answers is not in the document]
jared-2016: We do this elsewhere in the toolkit, e.g. https://github.com/Esri/arcgis-maps-sdk-toolkit-qt/blob/76097f9b632744e1eec87e2138367b6fb2a82aa1/uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/AuthenticationController.cpp#L150-L156

So my choice in this approach was continuity as I was following a similar approach that already existed and it was the simpler route. I could be persuaded to change it though.

We had this before for the QML API so I imagine this wouldn't require much effort to do.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/PopupElementViewItem.h
```diff
@@ -0,0 +1,17 @@
...
+#include <QObject>
+
+class PopupElementViewItem : public QObject
+{
+    Q_OBJECT
+public:
+    explicit PopupElementViewItem(QObject *parent = nullptr);
+
+signals:
+
```
JamesMBallard, line 14:
```suggestion
```

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/PopupView.qml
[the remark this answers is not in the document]
JamesMBallard: I'll second this request. I think we can do better by making the enum available in QML.

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/TextPopupElementView.qml
```diff
@@ -0,0 +1,62 @@
+/*******************************************************************************
+ *  Copyright 2012-2020 Esri
```
JamesMBallard, line 2: Current dates on new files.
```suggestion
 *  Copyright 2012-2024 Esri
```

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/TextPopupElementView.qml
```diff
@@ -0,0 +1,62 @@
...
+ ******************************************************************************/
+
+import Esri.ArcGISRuntime.Toolkit.Controller
+
+import QtQuick
+import QtQuick.Controls
+import QtQuick.Dialogs
+import QtQuick.Window
+import QtQuick.Layouts
+import QtWebView
```
JamesMBallard, line 24: Is this used?

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/TextPopupElementView.qml
```diff
@@ -0,0 +1,62 @@
...
```
jared-2016, line 24:
```suggestion
```

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/PopupViewController.cpp
jared-2016: Approach 1: I tried that idea with a new baseclass that all future PopupElementViewControllers would share which would just contain a PopupElement and the ability to check the popupElementType since that will be shared across all Controllers but no luck initially.

Approach 2: Also did a quick test subclassing QAbstractListModel and making it a little more specific to this use case by containing a QList<QVariant>. I got the TextPopupElements to display but this was quick and rough and there are bugs to work out.

I do feel like Approach 1 should work since it holds a QList<QObject*> which is what I am populating it with but getting that underlying controller object back to be set on the View just isn't working so far. I personally would like to go the first route because it's already there and seems like it should work.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/PopupViewController.cpp
```diff
@@ -62,6 +73,77 @@ PopupManager* PopupViewController::popupManager() const
...
+  if (m_popup)
+  {
+      disconnect(m_popup.data(), nullptr, this, nullptr);
+      m_popupElementsModel->removeRows(0, m_popupElementsModel->rowCount());
+  }
+
+  m_popup = popup;
+
+  if (m_popupManager)
+      qWarning() << "Both Popup and PopupManager have been set. The PopupView will default to PopupView which supports PopupElements.";
```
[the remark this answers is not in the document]
jared-2016: IDE formatter was fighting me on this as well because I think toolkit code styling is different from ours. I'll format and stick with toolkit styling.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/PopupViewController.h
```diff
@@ -37,8 +42,10 @@ namespace Toolkit
 class PopupViewController : public QObject
 {
   Q_OBJECT
+  Q_PROPERTY(Popup* popup READ popup WRITE setPopup NOTIFY popupChanged)
+  Q_PROPERTY(QAbstractListModel* popupElements READ popupElements NOTIFY popupChanged)
   Q_PROPERTY(PopupManager* popupManager READ popupManager WRITE setPopupManager NOTIFY popupManagerChanged)
-  Q_PROPERTY(QString title READ title NOTIFY popupManagerChanged)
+  Q_PROPERTY(QString title READ title NOTIFY titleChanged)
```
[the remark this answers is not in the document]
jared-2016: I think this was just done for connivence since it was directly linked to the popupManager. The only place `emit popupManagerChanged()` is emitted is in `setPopupManager` which I added `emit titleChanged()` there as well so im pretty confident it will behave the same.

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/TextPopupElementView.qml
jared-2016: No, Thank you for catching this.

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/PopupView.qml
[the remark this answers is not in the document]
jared-2016: I'll do it!

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/PopupViewController.cpp
JamesMBallard: Tough one. Either approach sounds viable but approach 1 sounds better if you can make it work.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/TextPopupElementViewController.cpp
```diff
@@ -0,0 +1,123 @@
...
+#include "TextPopupElementViewController.h"
+
+namespace Esri::ArcGISRuntime::Toolkit {
+
+/*!
+  \class Esri::ArcGISRuntime::Toolkit::TextPopupElementViewController
+  \inmodule ArcGISRuntimeToolkit
+  \ingroup ArcGISQtToolkitUiCppControllers
+  \brief In MVC architecture, this is the controller for the corresponding
+  \c TextPopupElementView.
```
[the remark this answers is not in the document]
jared-2016: Quick search of the toolkit codebase and I didn't see it anywhere.

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/PopupView.qml
```diff
@@ -73,6 +73,8 @@ Page {
      */
     property var popupManager: null
 
+    property var popup: null
```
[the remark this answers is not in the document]
jared-2016: I added it to this issue to create an issue for when all of these are completed to go and update that doc since I was thinking we wouldn't want to update the doc yet until the PopupView + PopupElements is completed.

jared-2016:
Label seems to work well but you lose a little bit of definition between each View.
LabelFrame

JamesMBallard:
>Label seems to work well but you lose a little bit of definition between each View. Makes it harder to tell them apart.

I like that better on the whole, but I think missing a clear divider between elements makes it a no-go for me. Especially in dark mode it looked like the divider disappeared completely. I would go with what you had previously, but I'll defer to @ldanzinger if he prefers Label.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/Internal/QmlEnums.h
```diff
@@ -0,0 +1,47 @@
...
+ ******************************************************************************/
+
+#ifndef ARCGIS_RUNTIME_TOOLKIT_CPP_QMLENUMS_H
+#define ARCGIS_RUNTIME_TOOLKIT_CPP_QMLENUMS_H
+
+// Qt headers
+#include <QObject>
+
+// STL headers
+#include <PopupTypes.h>
```
JamesMBallard, line 24: Git hooks strikes again. Wrong category.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/Internal/QmlEnums.h
```diff
@@ -0,0 +1,47 @@
...
+
+class QmlEnums : public QObject
+{
+  Q_OBJECT
+  public:
+  Q_INVOKABLE explicit QmlEnums(QObject* parent = nullptr);
+  ~QmlEnums() = default;
```
JamesMBallard, line 31: = default should appear in the .cpp file.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/Internal/QmlEnums.cpp
```diff
@@ -0,0 +1,28 @@
...
+#include "QmlEnums.h"
+
+/*!
+    \internal
+    \class QmlEnums
+    \brief This class exposes `QmlEnums.PopupElementType` such that it is accessible to QML.
+*/
+
+QmlEnums::QmlEnums(QObject* parent) :
+    QObject(parent)
```
JamesMBallard, line 26: Stick to consistent indent size.

JamesMBallard (commented):
Just some nitpicky stuff at this point. I think Lucas should give the all-clear to go with the new Label type, or go back to what you originally had.

jared-2016:
The more I look at it, the less it bothers me. Plus looking at AGOL, there are no clear dividers.

jared-2016:
@ldanzinger
Parameter doc for `setPopup` has been updated and the class brief for `QmlEnums` has been updated as well. I am trying to address the doc warnings/errors where possible but will write up a followup issue for this Epic to address all remaining doc warnings/errors once all individual PopupElementViews have been completed.

ldanzinger (approved):

PR #647, opened 2024-10-17
Jared/fields popup element view

jared-2016:
Captured some general doc fixes in this PR that are not directly applicable to FieldsPopupElement work but is related to the PopupView and PopupElements related work in general.

The primary work accomplished in this PR is the implementation of the FieldsPopupElementView and related controller. Most of the work is straight forward except two noteworthy things.

1. The creation of a single QVariantList of QVariantMap items so that we can have one model used to populate the ListView with the FieldsPopupElement `labels` and `formattedValues`. These are provided by the API in separate QList<QString> but in the case of how we wanted it to be displayed. Having it as two separate lists was counter productive. I will create a separate issue addressing this for a possible enhancement.
2. In the event the `formattedValue` contains a url to a website, the design specified that it should be presented as `View` and if clicked/touched it should open in the systems default browser like TextPopupElementView. This required some post processing of the formattedValue.

Demo:
FieldsPopupElementDemonstration_gif3


uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/FieldsPopupElementViewController.cpp
```diff
@@ -0,0 +1,99 @@
...
+                 std::back_inserter(combinedData),
+                 [](const QString &item1, const QString &item2) {
+    QVariantMap item;
+    item["label"] = item1;
+    item["formattedValue"] = item2;
+    return item;
+  });
+  emit fieldsPopupElementChanged();
+  return combinedData;
+}
```
JamesMBallard, line 84: I'm trying to wrap my head around this. It looks like you are iterating over both list1 and list2 at the same time. What happens if list1 and list2 have a different number of elements? Can that even happen?

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/FieldsPopupElementViewController.h
```diff
@@ -0,0 +1,60 @@
...
+ *  limitations under the License.
+ ******************************************************************************/
+#ifndef ESRI_ARCGISRUNTIME_TOOLKIT_FIELDSPOPUPELEMENTVIEWCONTROLLER_H
+#define ESRI_ARCGISRUNTIME_TOOLKIT_FIELDSPOPUPELEMENTVIEWCONTROLLER_H
+
+// Toolkit headers
+#include "PopupElementViewItem.h"
+
+// Qt headers
+#include <QAbstractListModel>
```
JamesMBallard, line 23: Why is this needed?

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/FieldsPopupElementViewController.h
```diff
@@ -0,0 +1,60 @@
...
+#include <QObject>
+#include <QPointer>
+
+// Maps SDK headers
+#include <FieldsPopupElement.h>
+#include <PopupElement.h>
+#include <PopupFieldListModel.h>
```
JamesMBallard, line 30: Can these be forward declared?

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/FieldsPopupElementViewController.h
```diff
@@ -0,0 +1,60 @@
...
+#ifndef ESRI_ARCGISRUNTIME_TOOLKIT_FIELDSPOPUPELEMENTVIEWCONTROLLER_H
+#define ESRI_ARCGISRUNTIME_TOOLKIT_FIELDSPOPUPELEMENTVIEWCONTROLLER_H
+
+// Toolkit headers
+#include "PopupElementViewItem.h"
+
+// Qt headers
+#include <QAbstractListModel>
+#include <QObject>
+#include <QPointer>
```
JamesMBallard, line 25: Can this be forward declared?

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/FieldsPopupElementViewController.cpp
```diff
@@ -0,0 +1,99 @@
...
+{
+  return static_cast<FieldsPopupElement*>(popupElement())->title();
+}
+
+/*!
+  \brief Returns the combination of each element from \l {Esri::ArcGISRuntime::FieldsPopupElement::labels} {labels} and
+  \l {Esri::ArcGISRuntime::FieldsPopupElement::formattedValues} {formattedValues} of the
+  \c FieldsPopupElement in a QList of QMap elements
+  */
+QVariantList FieldsPopupElementViewController::values()
```
ldanzinger, line 67: If this is something we expect most popup implementations would need, then we should consider if this is something we should put in the API itself

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/FieldsPopupElementViewController.cpp
jared-2016: @JamesMBallard I don't believe it's possible that they have different number of elements. I think it's possible the fields related formatted value might be empty or null if you didn't give it a value but the lists should always be the same.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/FieldsPopupElementViewController.h
jared-2016: It's not it was left over from some earlier testing that I missed.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/FieldsPopupElementViewController.cpp
JamesMBallard: In that case the `int size = qMin(list1.size(), list2.size());` is confusing to me if we can assume they're both the same size. I assume that is a safety check so we don't iterate beyond the end of either collection, but it leads the reader to believe they could be different sizes, which is what lead me to ask questions in the first place.

To me that means:
1. `int size = qMin(list1.size(), list2.size());` is unnecessary, or
2. We might leave off some elements if they're indeed different sizes.

My recommendation is to remove the `qMin` and use a Qt assert if the collections are different sizes. That should make it clearer that we are assuming they're the same size.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/FieldsPopupElementViewController.h
jared-2016: Yes, great suggestion!

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/FieldsPopupElementViewController.h
jared-2016: Yes, sorta, but looking at the code a little more closely it's not even needed since QPointer is included in `PopupElementViewItem.h` which is included by all PopupElementControllers because of their inheritance.

This leaves me wondering what benefit QPointer brings over just a raw pointer in this scenario. The Objects lifetime is managed by the parent which is the popup. I used it because it was used for m_popupManager in `PopupViewController.h` but it brings baggage with it and the benefits of QPointer are not being exercised. So I'm inclined to switch to just raw pointers for the stored PopupElement. Is there something I am missing here?

I don't think we can forward declare QPointer at all though since it's a template class and it requires it's full definition but I haven't worked with Template classes in a while so please correct me If I am wrong.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/FieldsPopupElementViewController.cpp
jared-2016: Ah I see where I caused confusion now. Thanks for pointing this out.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/FieldsPopupElementViewController.cpp
jared-2016: I think a ListModel with these accessible through roles would make a lot of sense to me since I imagine it would be really common to access both the label and it's value in the same context.

jared-2016: I'll elicit feedback from the team during the review.

ldanzinger (approved):
looks good from my perspective

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/FieldsPopupElementViewController.h
JamesMBallard: You can forward declare template types.

>Is there something I am missing here?

If you aren't using the benefits of QPointer (needing to know when it's been deleted), then I'd switch to a raw pointer.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/FieldsPopupElementViewController.cpp
JamesMBallard: >in a QList of QMap elements

This sounds more like a list model, if I understand the data layout correctly.

jared-2016: That's what I was trying to accomplish without going as far as subclassing QAbstractListModel just for this scenario because I thought this could be a useful addition to the API.

JamesMBallard (approved):

PR #661, opened 2024-12-17
James/fix build

JamesMBallard:
Some helpful header guards are coming to AuthenticationChallenge.h that I previously missed. Once added, the build will break. This include is not needed here, so it can be shuffled into the .cpp file.


uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/AuthenticationController.cpp
```diff
@@ -1,3 +1,4 @@
+
```
jared-2016, line 1:
```suggestion
```
Be weary of the pre-commit hook ghosts. Adding blank lines where no one asked for them.

jared-2016 (approved):
One tiny thing but lgtm.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/AuthenticationController.cpp
JamesMBallard: That's exactly what happened. Thanks.

PR #666, opened 2024-12-31
Guil8553/5327 duplicate

GuillaumeBelz:
Remove duplicate code in toolkit.

- move code from toolkitcpp to common
- remove code from toolkitwidgets
- same for images
- fix paths in config files
- remove `register_cpp` and `register_qml` and move code to `register`

General thoughts about naming. Maybe `toolkitcpp` is confusing. `toolkitwidgets` is a C++ API either. Maybe an explicit name with a reference to `QtQuick` is better? `toolkitquick` or `toolkitqml`?

There are 3 doc warnings:
- 2 about "Could not find the module header in include paths for module" because no module header is provided. A default module header is generated.
- 1 about OverviewMap. I don't know why.
```
toolkit/uitools/toolkitwidgets/src/OverviewMap.cpp:43: (qdoc) warning: 
Cannot tie this documentation to anything
[qdoc found a /*! ... */ comment, but there was no topic command (e.g., '\fn', '\page') in the 
comment and no function definition following the comment.]
```


augmented_reality/ArApi/source/ArcGISArView.cpp
```diff
@@ -90,7 +90,7 @@ Camera ArcGISArView::originCamera() const
 }
 
 /*!
-  \property ArcGISArView::originCamera
+  \property Esri::ArcGISRuntime::Toolkit::ArcGISArView::originCamera
```
GuillaumeBelz, line 93: Fix some doc issue like this. Namespace are required

uitools/toolkit.pri
```diff
@@ -25,9 +25,7 @@ HEADERS += $$REGISTERPATH/register.h \
 SOURCES += $$REGISTERPATH/register.cpp \
            $$REGISTERPATH/internal/register_cpp.cpp
 
-RESOURCES += \
-  $$PWD/toolkitcpp/images/esri_arcgisruntime_toolkit_images.qrc \
-  $$PWD/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/esri_arcgisruntime_toolkit_view.qrc
```
GuillaumeBelz, line 30: esri_arcgisruntime_toolkit_view.qrc doesn't exist (or I didn't find it?)

uitools/common/src/ArcGISAuthenticationController.cpp
```diff
@@ -40,7 +40,7 @@ namespace Esri::ArcGISRuntime::Toolkit {
 
 /*!
  \class Toolkit::ArcGISAuthenticationController
- \inmodule EsriArcGISRuntimeToolkit
+ \inmodule Esri.ArcGISRuntime.Toolkit
```
GuillaumeBelz, line 43: The name must be the same as declared in arcgisqttoolkit-overview.qdoc

JamesMBallard:
>Maybe an explicit name with a reference to QtQuick is better? toolkitquick or toolkitqml?

I like that idea, but can we do that without breaking any existing code written against the toolkit?

uitools/common.pri
JamesMBallard, line 1: Will deleting this file break existing user code? If so, we should keep it here as a shim that simply does an `include($$PWD/common/common.pri)`

uitools/toolkitcpp/src/register.h
JamesMBallard, line 1: Existing toolkit consuming apps do this:
```c++
#include "Esri/ArcGISRuntime/Toolkit/register.h"
```

and
```c++
Esri::ArcGISRuntime::Toolkit::registerComponents(engine);
```

There is more, too. We document what users need to do [here](https://github.com/Esri/arcgis-maps-sdk-toolkit-qt/tree/main/uitools/toolkitcpp). None of that can break.

JamesMBallard (commented):
Mostly looks good, but I am worried some of the changes will break client code.

GuillaumeBelz:
> but I am worried some of the changes will break client code.

If they copied only toolkitcpp or toolkitwidgets folder, they need to copy common folder. Except that, I guess it's possible to break nothing. I'm testing that for qmake and cmake projects, but for another issue.

uitools/common.pri
GuillaumeBelz: I can restore this file, but I'm not sure to see the purpose of this file. It looks like it should not be used by users.

GuillaumeBelz:
I tried with Samples with v.next branch, everything looks fine without any change.

GuillaumeBelz:
> I like that idea, but can we do that without breaking any existing code written against the toolkit?

No, if we don't want to break the code, I guess.

uitools/common.pri
JamesMBallard: I am not as familiar with the toolkit patterns as with the API patterns. If anything is an internal implementation detail, we can change it. We need to be careful about anything we've documented, and we should minimally ensure that none of our stuff breaks that uses the toolkit (gui tests, samples, etc...)

JamesMBallard:
>No, if we don't want to break the code, I guess.

Ok, let's wait on that idea then.

JamesMBallard (approved):
If the existing common.pri is not something we document, and there's no reasonable expectation users would be including that file directly, then you can remove it. I mean this one: https://github.com/Esri/arcgis-maps-sdk-toolkit-qt/blob/guil8553/5327-duplicate/uitools/common.pri

GuillaumeBelz:
> If the existing common.pri is not something we document

I missed something. It's not used, but it's referenced in documentation when users want to create their own controler.

PR #667, opened 2024-12-31
James/server trust auth

JamesMBallard:
The start of the Network Authentication story for the toolkit. This does the following:

 * Implements the "Server Trust" workflow, which is what we called "SSL Handshake" previously.
 * Simplifies the workflow so there is no "remember" option presented in the dialog. Any choice is persisted. 
 * Adds internal testing hooks so we can automate testing this toolkit component.
 * Now that there are two "Challenge Handlers", I want the ArcGISAuthenticationManager (better name pending) to implement both interfaces, but we can't since they both inherit from QObject. I moved to composition for that, which works fine but is more files/boilerplate.

Here is what the new dialog looks like. It's similar to the previous one.

Screenshot 2024-12-31 at 2 23 12 PM


uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/ArcGISAuthenticationController.cpp
```diff
@@ -138,28 +140,70 @@ void ArcGISAuthenticationController::handleArcGISAuthenticationChallenge(ArcGISA
...
+  std::lock_guard<std::mutex> lock(m_mutex);
+  challenge->setParent(nullptr);
+  m_currentNetworkChallenge = std::unique_ptr<NetworkAuthenticationChallenge>{challenge};
+
+  if (!m_currentNetworkChallenge)
+  {
+    return;
+  }
+
+  if (!m_currentNetworkChallenge->sslErrors().isEmpty())
```
JamesMBallard, line 157: We will need a better way to tell network challenges apart. That will come later once we have more of them.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/ArcGISAuthenticationChallengeRelay.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+namespace Esri::ArcGISRuntime::Toolkit {
+
+/*!
+ \class Toolkit::ArcGISAuthenticationChallengeRelay
+ \inmodule EsriArcGISRuntimeToolkit
+ \ingroup ArcGISQtToolkitUiCppControllers
+ \brief Simple implementation of ArcGISAuthenticationChallengeHandler.
+
+ \internal
```
ldanzinger, line 36: I like this 👍 

I believe this means that this class won't show up in the built doc, correct? Even still, maybe we could add a bit clearer of a brief than this. Maybe something along the lines of "An internal controller to handle ArcGIS challenges for the AuthenticationView"

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/NetworkAuthenticationChallengeRelay.cpp
```diff
@@ -0,0 +1,62 @@
...
+using namespace Esri::ArcGISRuntime;
+using namespace Esri::ArcGISRuntime::Authentication;
+
+namespace Esri::ArcGISRuntime::Toolkit {
+
+/*!
+ \class Toolkit::NetworkAuthenticationChallengeRelay
+ \inmodule EsriArcGISRuntimeToolkit
+ \ingroup ArcGISQtToolkitUiCppControllers
+ \brief Simple implementation of NetworkAuthenticationChallengeHandler.
```
ldanzinger, line 34: same comment here about the brief. Also, Handler -> Relay with the rename.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/ArcGISAuthenticationChallengeRelay.cpp
```diff
@@ -0,0 +1,62 @@
...
+using namespace Esri::ArcGISRuntime;
+using namespace Esri::ArcGISRuntime::Authentication;
+
+namespace Esri::ArcGISRuntime::Toolkit {
+
+/*!
+ \class Toolkit::ArcGISAuthenticationChallengeRelay
+ \inmodule EsriArcGISRuntimeToolkit
+ \ingroup ArcGISQtToolkitUiCppControllers
+ \brief Simple implementation of ArcGISAuthenticationChallengeHandler.
```
ldanzinger, line 34: Handler -> Relay

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/ServerTrustView.qml
```diff
@@ -0,0 +1,88 @@
...
+Dialog {
+
+    property ArcGISAuthenticationController controller: null
+
+    title: qsTr("Untrusted Host")
+
+    closePolicy: Popup.NoAutoClose
+
+    header: Label {
+        text: `<h1>${title}</h1>`
```
ldanzinger, line 32: this is neat

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/qmldir
```diff
@@ -54,6 +54,7 @@ Scalebar 100.13 Scalebar.qml
 Scalebar 200.0 Scalebar.qml
 SearchView 100.13 SearchView.qml
 SearchView 200.0 SearchView.qml
+ServerTrustView 200.0 ServerTrustView.qml
```
ldanzinger, line 57: Forgot about this file... With the new Qt 6 QML version changes, I wonder if there's opportunity to simplify this whole story? Obviously not as part of this PR...

GuillaumeBelz (approved):

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/ArcGISAuthenticationChallengeRelay.cpp
JamesMBallard: That is correct, it won't appear in the doc at all. I will beef up the explanation.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/NetworkAuthenticationChallengeRelay.cpp
JamesMBallard: It is an implementation of NetworkAuthenticationChallengeHandler, the interface from the API. I will reword it to make that clearer.

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/ServerTrustView.qml
JamesMBallard: I take no credit. This is simply ported from the existing SslHandshake view :-)

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/qmldir
JamesMBallard: Yeah good point. There are likely a few things we can clean up. I don't want to push things with this PR.

ldanzinger (approved):

PR #668, opened 2025-01-01
Jared/pe charts

jared-2016:
Well I made a mistake and got both branches up to date with v.next which seems to be including some of @JamesMBallard authentication work and im not sure how to fix it. I tried reverting the commits since it's not applicable at this point but source tree keeps telling me there is nothing to commit so 🤷 . Any suggestions on this is appreciated.

In other news, This PR adds the Charts to the WIP branch for Popup Elements. It also required refactoring of Images so that got shifted around a little but that functionality is the same.

Broke up individual chart types into their own classes(column and bar share the same class and the orientation is determined in QML by checking the type before loading the component).

Color is extracted from the json, if no color is provided there is a default theme set that is better than the default provided. If color is set on a series, slices etc. it overrides the theme but everything else stays e.g. white background etc.

White background is hard coded to match the WebMap View. I haven't commented on the inability to display negative and positive numbers in a BarSeries just yet. Was considering adding that to a later issue but in writing this, I have changed my mind and will add a comment now.


uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/ArcGISAuthenticationController.cpp
jared-2016, line 1: Not sure how to undo this.

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/MediaPopupElementView.qml
```diff
@@ -43,74 +44,221 @@ ColumnLayout {
...
+        model: mediaPopupElementView.controller.popupMediaItems
 
         Layout.fillWidth: true
-        Layout.leftMargin: mediaMargin
-        Layout.rightMargin: mediaMargin
+        Layout.leftMargin: mediaPopupElementView.mediaMargin
+        Layout.rightMargin: mediaPopupElementView.mediaMargin
 
         delegate: Item {
+            id: thisItem
```
ldanzinger, line 97: I think we can do better for an id name

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/MediaPopupElementView.qml
```diff
@@ -43,74 +44,221 @@ ColumnLayout {
...
+                    axisX: ValueAxis {
+                        id: xAxis
+                        max: listModelData.linePoints.length - 1
+                        tickInterval: 1
+                        labelsVisible: false
+                    }
+                    axisY: ValueAxis {
+                        id: yAxis
+                        max: listModelData.maxValue
+                        tickInterval: Math.abs(max) > Math.abs(min)? max / 3 : min / 3
```
ldanzinger, line 256: how was the 3 division decided on? does this work well in different cases?

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/MediaPopupElementView.qml
```diff
@@ -43,74 +44,221 @@ ColumnLayout {
...
+                    marginTop: 10
+                    marginBottom: 10
+                    marginLeft: 10
+                    marginRight: 10
+
+                    // Setting a default theme if no color information is provided on the LineSeries
+                    theme: GraphsTheme {
+                        backgroundColor: "white"
+                        plotAreaBackgroundColor: "white"
+                        theme: GraphsTheme.Theme.MixSeries
```
ldanzinger, line 244: Can you explain this? The background will be white, what will the other colors be?

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/ArcGISAuthenticationController.cpp
ldanzinger: Not sure either... @JamesMBallard?

ldanzinger (commented):
looking good, here is my feedback

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/LineChartPopupMediaItem.cpp
```diff
@@ -0,0 +1,122 @@
...
+    \li \a popupMedia - The \l {Esri::ArcGISRuntime::PopupMedia} {PopupMedia} used to populate the list of points.
+    \li \a jsonColorArray - The JSON array containing the color information for the line.
+    \li \a parent - The optional parent QObject.
+  \endlist
+ */
+LineChartPopupMediaItem::LineChartPopupMediaItem(PopupMedia* popupMedia, const QJsonArray& jsonColorArray, QObject* parent)
+  : PopupMediaItem{popupMedia, parent},
+    m_jsonArr{jsonColorArray}
+{
+  auto mediaValue = popupMediaItem()->value();
```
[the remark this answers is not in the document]
jared-2016: BarSeries and PieSeries both take ownership of their respective individual slices. I don't want to populate that QList<...*> until it's requested since the expectation is that if you request it, you need to take ownership of it(As their respective series will) and they will need to be populated every time this function is called. 

For LineChart it's a QList<QPoingtF> by value rather than pointer and the LineSeries doesn't take ownership. So since we can retain the original we only need to create that QList once and then a copy is provided to the LineSeries. We don't need to create that QList every time so it made sense to do it when the class was created rather than when the List was requested.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/LineChartPopupMediaItem.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+    const auto point = QPointF(i, value);
+    m_linePoints.append(point);
+  }
+
+  if (!m_jsonArr.isEmpty())
+  {
+    // color scheme is [Red, Green, Blue, Alpha]
+    m_color = QColor(m_jsonArr[0][0].toInt(), m_jsonArr[0][1].toInt(), m_jsonArr[0][2].toInt(), m_jsonArr[0][3].toInt());
+  }
```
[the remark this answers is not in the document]
jared-2016: I thought about this but the problem is that for pieslice we can create the slice and provide the label and value in the c'tor but for barset you can provide the label in the c'tor but not the value which needs to be appended.

Now that you brought it up, I didn't check to see if that barseries/barset workflow could apply to the PieSlice. I will double check.

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/MediaPopupElementView.qml
jared-2016: Given the initial display being a smaller window(preview if you will) I tested a few values that produced something similar to what I saw on the webmap and 3 was pretty close. When we move to the fullscreen view I am sure it will need changing but given the available screen real-estate, 3 looks good to me. I should have added some pictures to the PR title comment. I will add some for reference.

jared-2016:
Some images for reference:
Web MapViewer:


Android:
image
image
Macos:

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/MediaPopupElementView.qml
jared-2016: Good point. I can reference the Qt doc for `GraphsTheme.Theme.MixSeries`

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/ArcGISAuthenticationController.cpp
JamesMBallard: No need. This is meant to be part of v.next now. Automated OAuth tests are using it daily.

JamesMBallard:
>Well I made a mistake and got both branches up to date with v.next which seems to be including some of @JamesMBallard authentication work and im not sure how to fix it. I tried reverting the commits since it's not applicable at this point but source tree keeps telling me there is nothing to commit so 🤷 . Any suggestions on this is appreciated.

This is part of v.next now. There's no need to trim it out.
https://github.com/Esri/arcgis-maps-sdk-toolkit-qt/blob/v.next/uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/ArcGISAuthenticationController.h

uitools/toolkitcpp.pri
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 include($$PWD/common.pri)
 
-QT += quickcontrols2 webview svg
+QT += quickcontrols2 webview svg graphs
```
JamesMBallard, line 16: Is CMake support coming later?

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/ArcGISAuthenticationController.cpp
```diff
@@ -22,17 +22,19 @@
...
 #include <Error.h>
 #include <ErrorException.h>
-#include <OAuthUserConfiguration.h>
-#include <OAuthUserCredential.h>
-#include <OAuthUserLoginPrompt.h>
-#include <TokenCredential.h>
+#include <Authentication/OAuthUserConfiguration.h>
+#include <Authentication/OAuthUserCredential.h>
+#include <Authentication/OAuthUserLoginPrompt.h>
+#include <Authentication/TokenCredential.h>
```
JamesMBallard, line 33: If you merge the latest from v.next into the base `popupElementsWIP` then these differences should go away.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/BarChartPopupMediaItem.cpp
```diff
@@ -0,0 +1,116 @@
...
+#include <PopupMedia.h>
+#include <PopupMediaItem.h>
+#include <PopupMediaValue.h>
+
+namespace Esri::ArcGISRuntime::Toolkit {
+
+/*!
+  \class Esri::ArcGISRuntime::Toolkit::BarChartPopupMediaItem
+  \inmodule ArcGISRuntimeToolkit
+  \brief Represents the contents of a \c{MediaPopupElement::media} to be used for a BarSeries in QML.
```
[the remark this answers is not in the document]
JamesMBallard: That is my intent with the authentication stuff in the toolkit for now. I'm still in the process of changing a lot, so there's no point doc'ing it now. Also, it might make sense to just not doc it at all as a hint that the API is subject to change.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/BarChartPopupMediaItem.h
```diff
@@ -0,0 +1,61 @@
...
+  Q_OBJECT
+  Q_PROPERTY(QList<QBarSet*> barSets READ barSets NOTIFY barChartPopupMediaItemChanged)
+  Q_PROPERTY(qreal maxValue READ maxValue NOTIFY barChartPopupMediaItemChanged)
+  Q_PROPERTY(qreal minValue READ minValue NOTIFY barChartPopupMediaItemChanged)
+
+public:
+  explicit BarChartPopupMediaItem(PopupMedia* popupMedia, const QJsonArray& jsonColorArray, QObject* parent = nullptr);
+  ~BarChartPopupMediaItem() override = default;
+
+  QList<QBarSet*> barSets();
```
JamesMBallard, line 45: I think this should be private. That way the API is clear that it's only meant to be accessed via the property, in QML. In fact perhaps all these methods should be private throughout. That way it's clearer about the intent.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/BarChartPopupMediaItem.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+/*!
+  \brief Returns the list of \c QBarSet objects. You must take ownership of the returned QBarSets.
+ */
+QList<QBarSet*> BarChartPopupMediaItem::barSets()
+{
+  QList<QBarSet*> barSets;
+  auto mediaValue = popupMediaItem()->value();
+
+  for (int i = 0; i < mediaValue->data().length(); i++)
```
JamesMBallard, line 57: Consider storing the length once so you don't need to make a nested chain of calls for each loop iteration.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/BarChartPopupMediaItem.cpp
```diff
@@ -0,0 +1,116 @@
...
+    const auto label = mediaValue->labels().at(i);
+    const auto value = mediaValue->data().at(i).toReal();
+
+    if (value >= m_maxValue)
+      m_maxValue = value;
+
+    if (value <= m_minValue)
+      m_minValue = value;
+
+    const auto barset = new QBarSet(label, this);
```
JamesMBallard, line 68: Why does the doc state `You must take ownership of the returned QBarSets.`? They have a parent.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/BarChartPopupMediaItem.cpp
```diff
@@ -0,0 +1,116 @@
...
+    barset->append(value);
+    if (!m_jsonArr.isEmpty())
+    {
+      // color scheme is [Red, Green, Blue, Alpha]
+      auto color = QColor(m_jsonArr[i][0].toInt(), m_jsonArr[i][1].toInt(), m_jsonArr[i][2].toInt(), m_jsonArr[i][3].toInt());
```
JamesMBallard, line 73:
```suggestion
      const auto color = QColor(m_jsonArr[i][0].toInt(), m_jsonArr[i][1].toInt(), m_jsonArr[i][2].toInt(), m_jsonArr[i][3].toInt());
```

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/LineChartPopupMediaItem.cpp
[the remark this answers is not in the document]
JamesMBallard: Ditto - code duplication. At a minimum perhaps a QColor json helper.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/LineChartPopupMediaItem.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+} // namespace Esri::ArcGISRuntime::Toolkit
+
+/*!
+  \fn void Esri::ArcGISRuntime::Toolkit::LineChartPopupMediaItem::lineChartPopupMediaItemChanged()
+  \brief Signal emitted when the underlying \c PopupMedia changes.
+ */
+
+/*!
+  \property Esri::ArcGISRuntime::Toolkit::LineChartPopupMediaItem::linePoints
```
JamesMBallard, line 109: What does this document? That it just exists?

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/MediaPopupElementViewController.cpp
```diff
@@ -48,9 +55,40 @@ MediaPopupElementViewController::MediaPopupElementViewController(
...
+  for (int i = 0; i < media->size(); i++)
   {
-    m_popupMediaItems->append(new PopupMediaItem(media, this));
+    const QJsonDocument mediaPopupElementJson = QJsonDocument::fromJson(mediaPopupElement->toJson().toUtf8());
+    auto* popupMedia = media->at(i);
+
+    // Get color information for each popup media. If no color is selected when defining a popup, the default
+    // color will be used from the appropriate QtGraphs Series component.
+    // s.a. Web Map Spec: https://developers.arcgis.com/web-map-specification/objects/popupElement_media
+    const auto colors = mediaPopupElementJson.object()["mediaInfos"][i]["value"]["colors"].toArray();
```
JamesMBallard, line 67: I would make a helper for this that validates everything. This makes assumptions that will crash if they don't hold.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/MediaPopupElementViewController.cpp
```diff
@@ -48,9 +55,40 @@ MediaPopupElementViewController::MediaPopupElementViewController(
...
+        m_popupMediaItems->append(new PieChartPopupMediaItem(popupMedia, colors, media));
+        break;
+      case Esri::ArcGISRuntime::PopupMediaType::LineChart:
+        m_popupMediaItems->append(new LineChartPopupMediaItem(popupMedia, colors, media));
+        break;
+      case Esri::ArcGISRuntime::PopupMediaType::Unknown:
+        break;
+      default:
+        Q_UNIMPLEMENTED();
+        break;
```
JamesMBallard, line 90: Are there more types? Ideally we don't have default switch cases.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/PieChartPopupMediaItem.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+/*!
+  \brief Returns the list of \c QPieSlice objects. You must take ownership of the returned QPieSlices.
+ */
+QList<QPieSlice*> PieChartPopupMediaItem::pieSlices()
+{
+  QList<QPieSlice*> pieSlices;
+  auto mediaValue = popupMediaItem()->value();
+
+  for (int i = 0; i < mediaValue->data().length(); i++)
```
JamesMBallard, line 62: Ditto cache size.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/PieChartPopupMediaItem.cpp
```diff
@@ -0,0 +1,90 @@
...
+  {
+    const auto label = mediaValue->labels().at(i);
+    const auto value = mediaValue->data().at(i).toReal();
+
+    auto* pieSlice = new QPieSlice(label, value);
+    if (!m_jsonArr.isEmpty())
+    {
+      // color scheme is [Red, Green, Blue, Alpha]
+      auto color = QColor(m_jsonArr[i][0].toInt(), m_jsonArr[i][1].toInt(), m_jsonArr[i][2].toInt(), m_jsonArr[i][3].toInt());
```
JamesMBallard, line 71: This code appears multiple times. Definitely a helper for a QColor from json.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/PieChartPopupMediaItem.h
```diff
@@ -0,0 +1,53 @@
...
+class PieChartPopupMediaItem : public PopupMediaItem
+{
+  Q_OBJECT
+  Q_PROPERTY(QList<QPieSlice*> pieSlices READ pieSlices NOTIFY pieChartPopupMediaItemChanged)
+
+public:
+  explicit PieChartPopupMediaItem(PopupMedia* popupMedia, const QJsonArray& jsonColorArray, QObject* parent = nullptr);
+  ~PieChartPopupMediaItem() override = default;
+
+  QList<QPieSlice*> pieSlices();
```
JamesMBallard, line 41: I mentioned this on one property, but I think it could apply everywhere. Is there any case for other C++ code to call this? If not, make this method private so that its sole access is via the property in QML.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/PieChartPopupMediaItem.h
```diff
@@ -0,0 +1,53 @@
...
+
+signals:
+
+  void pieChartPopupMediaItemChanged();
```
JamesMBallard, line 45:
```suggestion
signals:
  void pieChartPopupMediaItemChanged();
```

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/PieChartPopupMediaItem.h
```diff
@@ -0,0 +1,53 @@
...
+#include <QObject>
+
+// Other headers
+#include "PopupMediaItem.h"
+
+class QPieSlice;
+
+namespace Esri::ArcGISRuntime {
+
+namespace Toolkit {
```
JamesMBallard, line 30: Unless there are Maps SDK forward declares to insert, just combine the namespaces

```c++
namespace Esri::ArcGISRuntime::Toolkit {
```

More changes needed at the bottom of the file.

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/MediaPopupElementView.qml
```diff
@@ -43,74 +44,221 @@ ColumnLayout {
...
+                    loader.sourceComponent = imageComp;
+                } else if (popupMediaType === QmlEnums.PopupMediaTypeColumnChart) {
+                    loader.sourceComponent = columnChartComp;
+                } else if (popupMediaType === QmlEnums.PopupMediaTypeBarChart) {
+                    loader.sourceComponent = barChartComp;
+                } else if (popupMediaType === QmlEnums.PopupMediaTypePieChart) {
+                    loader.sourceComponent = pieChartComp;
+                } else if (popupMediaType === QmlEnums.PopupMediaTypeLineChart) {
+                    loader.sourceComponent = lineChartComp;
+                }
```
JamesMBallard, line 116: Would this be easier to read as a switch?

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/MediaPopupElementView.qml
```diff
@@ -43,74 +44,221 @@ ColumnLayout {
...
+
+                    axisY: ValueAxis {
+                        max: listModelData.maxValue
+                        min: listModelData.minValue
+                        tickInterval: Math.abs(max) > Math.abs(min)? max / 3 : Math.abs(min) / 3
+                    }
+
+                    BarSeries {
+                        // BarSeries will take owership of the QList<QBarSet*> when we call append. This also applies to PieSeries and QPieSlice for PieChartPopupMediaItem.
+                        // s.a https://github.com/qt/qtgraphs/blob/ef90c963d10a1af30ed523483e9a589235eccef9/src/graphs2d/barchart/qbarseries.cpp#L670-L692
```
JamesMBallard, line 167: Is this documented, or only discovered by digging through the sources? We should work with Qt to get this documented.

uitools/toolkitcpp.pri
jared-2016: Yes I created a separate issue for the cmake integration once everything is complete with qmake.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/ArcGISAuthenticationController.cpp
jared-2016: I'll give that a shot.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/MediaPopupElementViewController.cpp
jared-2016: No there are not. I'll remove this.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/BarChartPopupMediaItem.cpp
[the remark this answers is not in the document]
jared-2016: I really like this idea. I think it makes sense to document PopupView and the PopupViewController class since those are the intended interfaces for displaying Popups with the toolkit. All the underlying controllers, views, etc. are not intended to be used directly.

> Also, it might make sense to just not doc it at all as a hint that the API is subject to change.

Agreed for anything not `PopupView.qml` and `PopupViewController.h/cpp`

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/BarChartPopupMediaItem.cpp
jared-2016: I was undecided on if I wanted to give these a parent or not since the Series takes ownership. If you look at `PieChartPopupMediaItem` you can see i didn't give that one a parent.

I was concerned over this getting leaked potentially but giving it a parent covers that.

JamesMBallard: If you are confident that the QML side takes ownership, I wouldn't provide any parent. Having a parent means that there are two owners, and if the BarChartPopupMediaItem goes away first, the QML side will have dangling pointers and could crash when it tries to delete them.

I think I commented elsewhere, but I would consider providing no parent at all, and make the method private so it can solely be executed by QML code via the property. That way it's clear that C++ code can't even invoke it (without some hacks, but the intent is clear).

Having a parent is a good insurance policy in most cases, but it could cause crash on exit types of problems that could be hard to diagnose.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/LineChartPopupMediaItem.cpp
jared-2016: Yeah, just that this `Q_PROPERTY` exists, https://doc.qt.io/qt-6/13-qdoc-commands-topics.html#property

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/MediaPopupElementView.qml
jared-2016: It looks like it is but not on the QML side of the documentation just the C++ side. Ill update the link to the C++ documentation which does indicate that function is available via QML

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/BarChartPopupMediaItem.cpp
jared-2016: > If you are confident that the QML side takes ownership

Yes BarSeries and PieSeries take ownership 100%. 

> I would consider providing no parent at all, and make the method private so it can solely be executed by QML code via the property

Done!

> pointers and could crash when it tries to delete them.

Ahh that's a good point as I was running into the reverse of that situation.

Thank you for the recommendation @JamesMBallard .

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/MediaPopupElementViewController.cpp
```diff
@@ -16,18 +16,81 @@
...
+#include <LineChartPopupMediaItem.h>
 
 namespace Esri::ArcGISRuntime::Toolkit {
 
+namespace {
+
+  template <typename T>
+  bool isNullOrUndefined(const T& value) {
+      return value.isNull() || value.isUndefined();
+  }
```
jared-2016, line 43: I don't think this is necessary to be templatized but I don't find the opportunity to do so very often so I went for it. 😄

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/MediaPopupElementViewController.cpp
```diff
@@ -16,18 +16,81 @@
...
+
+  // Get color information for each popup media. If no color is selected when defining a popup, GraphsTheme.Theme.MixSeries will
+  // be applied to the GraphsView.
+  // s.a. Web Map Spec: https://developers.arcgis.com/web-map-specification/objects/popupElement_media
+  QJsonArray colorsArrayFromJson(const QByteArray& mediaPopupElementJson, const int index)
+  {
+    if (mediaPopupElementJson.isEmpty())
+    {
+      qCritical() << QString("MediaPopupElement JSON is empty.");
```
jared-2016, line 52: The helper verifies all the assumptions that are made with the json scheme as it currently is. It's thorough but I think it's necessary since if one of these parts do fail, that information would be helpful to user even if we handle the fail gracefully by passing forward an empty QJsonArray.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/PopupMediaItem.cpp
```diff
@@ -52,52 +61,49 @@ QString PopupMediaItem::title() const
...
 }
 
-/*!
-  \brief Returns the popupMediaType of the \c MediaPopupElement.
- */
-PopupMediaType PopupMediaItem::popupMediaType() const
+QColor PopupMediaItem::jsonColorHelper(const QJsonArray& colorArray, const int index) const
 {
-  return m_popupMedia->popupMediaType();
+  if (!colorArray.isEmpty() && index < colorArray.size())
```
jared-2016, line 89: helper for extracting the color information.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/LineChartPopupMediaItem.cpp
```diff
@@ -0,0 +1,134 @@
...
+      m_maxValue = value;
+
+    if (value <= m_minValue)
+      m_minValue = value;
+
+    const auto point = QPointF(i, value);
+    m_linePoints.append(point);
+  }
+
+  m_color = jsonColorHelper(m_jsonColorArr, 0);
```
jared-2016, line 66: These are much nicer. I did consider making the parameter `jsonColorHelper(m_jsonColorArr[index])` but we lose the index verification. I don't mind this function signature but there is the possibility to remove a parameter so I wanted to acknowledge that.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/PopupMediaItem.h
```diff
@@ -35,24 +34,23 @@ namespace Toolkit
...
   explicit PopupMediaItem(PopupMedia* popupMedia, QObject* parent = nullptr);
 
   QString title() const;
-  QUrl linkUrl() const;
-  QUrl sourceUrl() const;
   QString caption() const;
   PopupMediaType popupMediaType() const;
+  PopupMedia* popupMediaItem() const;
+
+  QColor jsonColorHelper(const QJsonArray& colorArray, const int index) const;
```
JamesMBallard, line 50: I would consider making this static. It doesn't need anything from the class itself, so it shouldn't be required to have an object to use it.

Also, it's pretty clear already, but `jsonColorHelper` isn't as descriptive as it could be. Maybe something like `toColor` or `extractColor`? Your name is not terrible, so I'm ok if you prefer to keep it.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/LineChartPopupMediaItem.cpp
JamesMBallard: That makes sense. I like it the way you have it, since the helper takes on the burden of validating the json at that index, which makes the calling code more readable.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/PopupMediaItem.cpp
```diff
@@ -52,52 +61,49 @@ QString PopupMediaItem::title() const
...
 {
-  return m_popupMedia->caption();
+  return m_popupMedia;
 }
 
-/*!
-  \brief Returns the popupMediaType of the \c MediaPopupElement.
- */
-PopupMediaType PopupMediaItem::popupMediaType() const
+QColor PopupMediaItem::jsonColorHelper(const QJsonArray& colorArray, const int index) const
```
JamesMBallard, line 87: `const int` doesn't do much, but it does prevent you from changing `index` (even though it would have no side effects if you did). I've leaned towards just `int` rather than `const int`, but it does provide some value so let's go with what you have 👍

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/PopupMediaItem.cpp
```diff
@@ -52,52 +61,49 @@ QString PopupMediaItem::title() const
...
 {
-  return m_popupMedia->popupMediaType();
+  if (!colorArray.isEmpty() && index < colorArray.size())
+  {
+    // color scheme is [Red, Green, Blue, Alpha]
+    // https://developers.arcgis.com/web-map-specification/objects/color
+    return QColor(colorArray[index][0].toInt(),
+                  colorArray[index][1].toInt(),
+                  colorArray[index][2].toInt(),
+                  colorArray[index][3].toInt());
```
JamesMBallard, line 96: It's probably overkill, so this is optional, but consider something like this for readability.
```c++
const auto jsonColor = colorArray[index];
const auto r = colorArray[0].toInt();
const auto g = colorArray[1].toInt();
const auto b = colorArray[2].toInt();
const auto a = colorArray[3].toInt();
return QColor{r, g, b, a};
```

Might be too much for something as simple as this, so up to you. You could probably do some fancy looping, but that would be overkill IMO.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/MediaPopupElementViewController.cpp
JamesMBallard: What types would you expect to come through here? It has to be something that supports both `isNull()` and `isUndefined()`.

JamesMBallard (commented):
Looking better. See comments.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/MediaPopupElementViewController.cpp
jared-2016: I thought more of the QJson... types had both but it looks like `isUndefined` is unique to QJsonValue. I'll curb my enthusiasm and reverse course as it's for the best 😄

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/BarChartPopupMediaItem.cpp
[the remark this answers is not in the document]
jared-2016: @ldanzinger @JamesMBallard. I want to make sure we are all on the same page here before I make any drastic changes. Any objections to what I had mentioned?

uitools/toolkitcpp.pri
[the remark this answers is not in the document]
jared-2016: These are implemented using Qt Graphs. Am I miss understanding something @GuillaumeBelz ?

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/ArcGISAuthenticationController.cpp
jared-2016: I haven't tried merging v.next into the WIP branch just yet.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/BarChartPopupMediaItem.cpp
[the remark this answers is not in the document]
JamesMBallard: I am on board. I recommend the following workflow:
 * Fully document classes that are intended to be used directly in user code.
 * Internal classes not intended for direct use, mark everything with `\internal` and I would go further and add doc to the class (just once) indicating it's an internal implementation detail and subject to change.

Moving forward, we commit to not changing/breaking the public API (the documented classes), but are free to change any of the "internal" classes so long as it doesn't impact user code.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/PopupMediaItem.h
jared-2016: @JamesMBallard  `jsonColorExtractor`?

uitools/toolkitcpp.pri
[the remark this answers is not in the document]
GuillaumeBelz: I'm just crazy. I read "charts". Sorry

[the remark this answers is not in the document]
jared-2016: 👻

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/PopupMediaItem.cpp
jared-2016: Loop is definitely overkill but putting the values in descriptive variables is a good idea.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/BarChartPopupMediaItem.cpp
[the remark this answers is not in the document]
ldanzinger: Agree with @JamesMBallard's proposal

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/PopupMediaItem.h
JamesMBallard: How about `extractColorFromJson`? `jsonColorExtractor` sounds more like the name of a class/utility than the method itself. I find it useful to use verbs for the method names.

Like this:
```c++
struct DoesAThing
{
  void doTheThing();
};

DoesAThing helperThatDoesAThing;
helperThatDoesAThing.doTheThing();
```

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/BarChartPopupMediaItem.cpp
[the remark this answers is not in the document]
jared-2016: Full steam ahead on this now. I will keep the changes limited to what is in this PR then I will open another for the rest of the controllers/views/etc.

jared-2016:
All doc has added the `/internal` tag as well as a single description for each class that they are subject to change.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/BarChartPopupMediaItem.cpp
```diff
@@ -0,0 +1,130 @@
...
+#include <PopupMediaValue.h>
+
+// Toolkit headers
+#include <PopupMediaItem.h>
+
+namespace Esri::ArcGISRuntime::Toolkit {
+
+/*!
+  \internal
+  This class is an internal implementation detail and is subject to change.
```
JamesMBallard, line 33: Now that these are added, I would be very clear and remove everything else. I think `\internal` takes precedence in the doc build itself, but a user might overlook that and see the doc that's still here like the `\brief` below and assume this is fair game to use directly in their code.

JamesMBallard (commented):
I would be explicit and remove existing doc anywhere you've added `\internal` comments. Let's leave no doubt about what is internal API.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/BarChartPopupMediaItem.cpp
jared-2016: Done.

JamesMBallard (approved):

ldanzinger (commented):
couple questions. Also, will you be logging a follow up issue to add \internal to the other popup classes not related to charts?

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/MediaPopupElementViewController.cpp
```diff
@@ -16,86 +16,174 @@
...
+    */
+  bool isNullOrUndefined(const QJsonValue& value) {
+      return value.isNull() || value.isUndefined();
+  }
+
+  // Get color information for each popup media. If no color is selected when defining a popup, GraphsTheme.Theme.MixSeries will
+  // be applied to the GraphsView.
+  // s.a. Web Map Spec: https://developers.arcgis.com/web-map-specification/objects/popupElement_media
+  /*!
+    \internal
```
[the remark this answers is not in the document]
jared-2016: I did it as to leave no ambiguity but if one for just the entire class is sufficient and common practice, I would prefer that over putting internal doc on every function. @ldanzinger @JamesMBallard

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/ImagePopupMediaItem.cpp
```diff
@@ -0,0 +1,74 @@
...
+#include <PopupMedia.h>
+#include <PopupMediaValue.h>
+
+// Toolkit headers
+#include <PopupMediaItem.h>
+
+namespace Esri::ArcGISRuntime::Toolkit {
+
+/*!
+  \class Esri::ArcGISRuntime::Toolkit::ImagePopupMediaItem
```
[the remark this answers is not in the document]
jared-2016: Ah I missed all of ImagePopupMediaItem. Thank you.

uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/MediaPopupElementViewController.cpp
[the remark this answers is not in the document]
JamesMBallard: I am good with that, as long as the doc build is happy with it as well. It should be.

jared-2016:
> Also, will you be logging a follow up issue to add \internal to the other popup classes not related to charts?

I was just going to open a PR once this one was finished. It's low effort to get everything marked internal popup wise and I might as well do it while I am at it.

ldanzinger (approved):
lgtm

PR #670, opened 2025-01-07
Guil8553/fix build

GuillaumeBelz:
Fix toolkit build

The fix is in esri_arcgisruntime_toolkit_qml.qrc.

The rest is cleanup. In most of QML files, paths are given in relative path `source: "images/pin.svg"`. These paths depends on the configuration and can be broken if names are changed or folders moved. It's better to use URL to qrc file `source: "qrc:/Esri/Runtime/Toolkit/pin.svg"``


uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/esri_arcgisruntime_toolkit_qml.qrc
```diff
@@ -1,5 +1,5 @@
 <RCC>
-    <qresource prefix="/esri.com/qml/Esri/ArcGISRuntime/Toolkit">
+    <qresource prefix="/esri.com/imports/Esri/ArcGISRuntime/Toolkit">
```
GuillaumeBelz, line 2: This is the real fix. I made a mistake with this name.

JamesMBallard (approved):

PR #673, opened 2025-01-14
Cleanup and simplify the server trust workflow.

JamesMBallard:
Follow up to a recent conversation. This removes the concept of "remember" or "trust". We either create the credential if users wants to trust, or we don't and rather `continueAndFailWithError` if the user does chooses to reject/block.


uitools/common/src/ArcGISAuthenticationController.cpp
```diff
@@ -187,13 +187,19 @@ void ArcGISAuthenticationController::handleNetworkAuthenticationChallenge(Networ
...
-  if (credential)
+  if (trust)
   {
-    m_currentNetworkChallenge->continueWithCredential(credential);
+    auto* credential = ServerTrustCredential::createWithChallenge(m_currentNetworkChallenge.get(), this);
+    if (credential)
+    {
+      m_currentNetworkChallenge->continueWithCredential(credential);
+      m_currentNetworkChallenge.reset();
+    }
```
GuillaumeBelz, line 197: if no credential, m_currentNetworkChallenge is not reset ?

GuillaumeBelz (approved):

uitools/common/src/ArcGISAuthenticationController.cpp
JamesMBallard: You are right, that is not covered. In theory, `ServerTrustCredential::createWithChallenge` could return null if the challenge type doesn't match, so I covered that possibility.

Good catch, thanks.

GuillaumeBelz (approved):

ldanzinger (approved):

PR #675, opened 2025-01-15
exclude mobile from widgets

GuillaumeBelz:
exclude mobile platforms from toolkit widgets

### DIFF
--- a/uitools/toolkitwidgets/toolkitwidgets.pri
+++ b/uitools/toolkitwidgets/toolkitwidgets.pri
@@ -13,26 +13,27 @@
 # limitations under the License.
 include($$PWD/../common/common.pri)
 
+# toolkitwidgets is not supported on mobile platforms
 android|ios {
-    message("toolkitwidgets.pri is not usable on mobile platforms.")
-} else {
-    QT += widgets webenginewidgets svg
-
-    TOOLKITWIDGETS_BASE_SRC = $$PWD/src
-    TOOLKITWIDGETS_TOOLKIT_SRC = $$TOOLKITWIDGETS_BASE_SRC/Esri/ArcGISRuntime/Toolkit
-
-    INCLUDEPATH += $$TOOLKITWIDGETS_BASE_SRC
-
-    DEPENDPATH += $$TOOLKITWIDGETS_BASE_SRC
-
-    HEADERS += $$files($$TOOLKITWIDGETS_TOOLKIT_SRC/*.h) \
-               $$files($$TOOLKITWIDGETS_TOOLKIT_SRC/Internal/*.h)
-
-    SOURCES += $$files($$TOOLKITWIDGETS_TOOLKIT_SRC/*.cpp) \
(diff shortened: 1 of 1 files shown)


uitools/toolkitwidgets/toolkitwidgets.pri
```diff
@@ -13,22 +13,24 @@
 # limitations under the License.
 include($$PWD/../common/common.pri)
 
-QT += widgets webenginewidgets svg
+unix:win32:macx {
```
JamesMBallard, line 16: will android trigger the `unix` section? I recall we've had to make exceptions for that in the past.

Also, presumably we should throw an error or something if someone tries to use this on Android.

I don't think we should allow this to be included by platforms that should not use it, and just have it silently do nothing.

GuillaumeBelz:
Tested on Mac, Android and iOS, on local machine.

uitools/toolkitwidgets/toolkitwidgets.pri
```diff
@@ -13,22 +13,26 @@
 # limitations under the License.
 include($$PWD/../common/common.pri)
 
-QT += widgets webenginewidgets svg
+android || ios {
```
JamesMBallard, line 16:
```suggestion
android|ios {
```

I am frequently confused by qmake logical operators, but I checked the doc to confirm and it's only a single `|'.
https://doc.qt.io/qt-6/qmake-language.html#:~:text=There%20is%20also%20the%20%7C%20operator

uitools/toolkitwidgets/toolkitwidgets.pri
```diff
@@ -13,22 +13,26 @@
...
+    message("toolkitwidgets.pri is not usable on mobile platforms.")
+} else {
+    QT += widgets webenginewidgets svg
```
JamesMBallard, line 19: I should've mentioned this before, but I don't think we need the else block. It can all remain unindented. This is the qmake version of https://clang.llvm.org/extra/clang-tidy/checks/readability/else-after-return.html haha.

GuillaumeBelz: There is not return in qmake (I guess). Without `else`, that means only the message is displayed and there is a build error for mobile platforms due to webenginewidgets.

uitools/toolkitwidgets/toolkitwidgets.pri
GuillaumeBelz: Right. I didn't check the doc. As the code worked on my local test, I assumed it's correct.

uitools/toolkitwidgets/toolkitwidgets.pri
JamesMBallard: Oh my mistake, you are right. I thought `message` was `error`.  🤦‍♂️ 

Use `error` and that will halt the build on mobile platforms. Would that work?

GuillaumeBelz: Done

JamesMBallard (approved):

PR #683, opened 2025-01-20
Update README.md

GuillaumeBelz:
Fix snippet in doc

### DIFF
--- a/uitools/toolkitwidgets/README.md
+++ b/uitools/toolkitwidgets/README.md
@@ -87,7 +87,6 @@ A good way to start using the toolkit is to use one of the ArcGIS Maps SDK for Q
 > auto northArrow = new Esri::ArcGISRuntime::Toolkit::NorthArrow(this);
 > northArrow->setMapView(m_mapView); // for a 2D map
 > // or northArrow->setSceneView(m_sceneView); // for a 3D scene
-> northArrow->show();
 > ```
 
 **STEP 6:** When you run your app, you should now see the UI for the Qt toolkit component in your app. For example:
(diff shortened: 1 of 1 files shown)


uitools/toolkitwidgets/README.md
```diff
@@ -83,10 +83,11 @@ A good way to start using the toolkit is to use one of the ArcGIS Maps SDK for Q
...
 > #include "Esri/ArcGISRuntime/Toolkit/NorthArrow.h"
 >  ...
->
+> // Add your NorthArrow to your UI here
 > auto northArrow = new Esri::ArcGISRuntime::Toolkit::NorthArrow(this);
-> northArrow->setMapView(m_mapView);
-> // Add your NorthArrow to your UI here!
+> northArrow->setMapView(m_mapView); // for a 2D map
+> // or northArrow->setSceneView(m_sceneView); // for a 3D scene
+> northArrow->show();
```
bri12415, line 90: Do we need to call `show()` in this case? For me it defaults to being visible when I construct it.

I see this in the QWidget doc
```
If you add a child widget to an already visible widget you must explicitly show the child to make it visible.
```
So maybe it is best practice to call this.

bri12415 (approved):

uitools/toolkitwidgets/README.md
GuillaumeBelz: No, you're right

uitools/toolkitwidgets/README.md
```diff
@@ -83,10 +83,10 @@ A good way to start using the toolkit is to use one of the ArcGIS Maps SDK for Q
 > // Locate this existing line in your project
```
bri12415, line 83: I missed this, but I think this is also incorrect.

This line wont exist in the user's project, they'll need to add it themselves.

PR #686, opened 2025-02-10
Jared/attachments

jared-2016:
This PR adds Attachments for PopupElements. There is a known limitation currently for mobile which I have an open support ticket about but I have not heard back so I am proceeding with what we got for now.

Adds some new svg images to calcite to be uses for the attachments types we support through the API. Lists the attachments and once downloaded locally if it's clicked again it will attempt to open them using an external application specified by the host system.

popupELementsAttach


jared-2016:
This PR is ready for review. @ldanzinger @JamesMBallard. I've updated the demo gif at the top that more accurately reflects it's current state. I'll do one for iOS and Android as well in the morning.

jared-2016:
iOS - 
iosDemo

uitools/common/src/Internal/PopupAttachmentImageProvider.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+  const auto key = item->id();
+
+  if (m_itemMap.contains(key))
+    return false;
+
+  m_itemMap.insert(key, item);
+  QObject::connect(item, &QObject::destroyed, m_internalObject, [this, key]
+  {
+    m_itemMap.remove(key);
```
JamesMBallard, line 179: Is there any risk with the async image provider stuff that there could be any data races on `m_itemMap`? I assume the connections will either be direct, or queued if there is threading involved and everything will happen on the same thread. Thoughts?

uitools/common/src/PopupAttachmentItem.h
```diff
@@ -0,0 +1,83 @@
...
+  QImage thumbnail() const;
+  QUrl localData() const;
+  QUuid id() const;
+  void setThumbnail(const QImage& thumbnail);
+
+signals:
+  void popupAttachmentItemChanged();
+
+private:
+  bool m_fetchingAttachment;
```
JamesMBallard, line 73:
```suggestion
  bool m_fetchingAttachment = false;
```

uitools/common/src/PopupAttachmentItem.h
```diff
@@ -0,0 +1,83 @@
...
+  PopupAttachment* m_popupAttachment = nullptr;
+  QImage m_thumbnail;
+  QUrl m_localData;
+  QUuid m_id;
```
JamesMBallard, line 77: If any of these never change (m_id?), consider making them const.

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/AttachmentsPopupElementView.qml
```diff
@@ -0,0 +1,212 @@
...
+
+            MouseArea {
+                anchors.fill: parent
+                hoverEnabled: true
+
+                onClicked: {
+                    if (!model.listModelData.dataFetched) {
+                        model.listModelData.downloadAttachment();
+                    } else {
+                        if( model.listModelData.popupAttachmentType === QmlEnums.PopupAttachmentTypeImage) {
```
JamesMBallard, line 132:
```suggestion
                        if (model.listModelData.popupAttachmentType === QmlEnums.PopupAttachmentTypeImage) {
```

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/AttachmentsPopupElementView.qml
```diff
@@ -0,0 +1,212 @@
...
+                        elide: Text.ElideRight
+                    }
+                }
+
+                Image {
+                    id: downloadImage
+                    fillMode: Image.PreserveAspectFit
+                    Layout.preferredHeight: fileInfoColumn.height
+                    Layout.preferredWidth: fileInfoColumn.height
+                    visible: !fetchAttachment.running ? !model.dataFetched : !fetchAttachment.running
```
JamesMBallard, line 195:
```suggestion
                    visible: !fetchAttachment.running && !model.dataFetched
```

I think this is equivalent. The robot says so, but please double check.

uitools/common/src/Internal/PopupAttachmentImageProvider.cpp
```diff
@@ -0,0 +1,198 @@
...
+      case PopupAttachmentType::Document:
+        icon = QIcon(":/esri.com/imports/Calcite/images/file.svg");
+        break;
+      case PopupAttachmentType::Other:
+        [[fallthrough]];
+      default:
+        icon = QIcon(":/esri.com/imports/Calcite/images/other.svg");
+        break;
+    }
+
```
ldanzinger, line 124:
```suggestion
```

uitools/common/src/Internal/PopupAttachmentImageProvider.h
```diff
@@ -0,0 +1,49 @@
+/*******************************************************************************
+ *  Copyright 2012-2021 Esri
```
ldanzinger, line 2:
```suggestion
 *  Copyright 2012-2025 Esri
```

uitools/common/src/Internal/PopupAttachmentImageProvider.cpp
```diff
@@ -0,0 +1,198 @@
...
+  // Handles edge case for Mobile since QAbstractFileIconProvider returns a Null QIcon.
+#if defined(Q_OS_IOS) || defined(Q_OS_ANDROID)
+  if (thumbnail.isNull() || m_popupAttachmentItem->popupAttachmentType() != PopupAttachmentType::Image)
+  {
+    switch (m_popupAttachmentItem->popupAttachmentType()) {
+      case PopupAttachmentType::Image:
+        icon = QIcon(":/esri.com/imports/Calcite/images/image.svg");
+        break;
+      case PopupAttachmentType::Video:
+        icon = QIcon(":/esri.com/imports/Calcite/images/video.svg");
```
ldanzinger, line 113: assuming these work, even without qrc path defined? sometimes they get tripped up on mobile if not referenced inside of the qrc

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/AttachmentsPopupElementView.qml
```diff
@@ -0,0 +1,212 @@
...
+                right: parent.right
+            }
+            fillMode: Image.PreserveAspectFit
+        }
+
+        Button {
+            id: closeButton
+            text: "Close"
+            anchors.bottom: parent.bottom
+            anchors.horizontalCenter: parent.horizontalCenter
```
ldanzinger, line 66: recommend grouping

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/AttachmentsPopupElementView.qml
```diff
@@ -0,0 +1,212 @@
...
+
+                onClicked: {
+                    if (!model.listModelData.dataFetched) {
+                        model.listModelData.downloadAttachment();
+                    } else {
+                        if( model.listModelData.popupAttachmentType === QmlEnums.PopupAttachmentTypeImage) {
+                            fullScreenImage.source = model.listModelData.localData;
+                            fullScreenImageDialog.visible = true;
+                        } else {
+                            if (Qt.platform.os !== "android" && Qt.platform.os !== "ios") {
```
ldanzinger, line 136: probably worthy of a comment about the differing behavior and possibly a link to any bugs/enhancements we are tracking, as we ideally can support this on mobile in the future

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/AttachmentsPopupElementView.qml
```diff
@@ -0,0 +1,212 @@
...
+                                Qt.openUrlExternally(model.listModelData.localData);
+                            }
+                        }
+                    }
+                }
+                onEntered: {
+                    if (model.listModelData.dataFetched && nameLabel.text !== "Click to open")
+                        nameLabel.text = "Click to open";
+
```
ldanzinger, line 145:
```suggestion
```

ldanzinger (commented):
Looking good, just a couple of comments

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/AttachmentsPopupElementView.qml
jared-2016: Ah yes. I was waiting to hear back about the thumbnails as well. I'll add a bug link and explanation here.

uitools/common/src/Internal/PopupAttachmentImageProvider.cpp
jared-2016: They have been working but i'll put it in just in case.

calcite/Calcite/images/download.svg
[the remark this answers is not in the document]
jared-2016: I have not followed up with the iOS team for suggestions yet. I will do that while this is in review. Thanks

uitools/common/src/PopupAttachmentItem.h
jared-2016: Good idea, the id does not change.

uitools/common/src/PopupAttachmentItem.h
jared-2016: Is this needed in the header? It's part of the constructor initialization list, https://github.com/Esri/arcgis-maps-sdk-toolkit-qt/blob/ef1f7812416cc7a7b8f85aaee6660fd498b379ea/uitools/common/src/PopupAttachmentItem.cpp#L7

uitools/common/src/PopupAttachmentItem.cpp
```diff
@@ -0,0 +1,185 @@
...
+  QString formatFileSize(qint64 bytes) {
+    if (bytes < 0) {
+        return "Invalid size";
+    }
+    if (bytes == 0) {
+        return "0 B";
+    }
+
+    const char* units[] = { "B", "KB", "MB", "GB", "TB" };
+    auto size = static_cast<double>(bytes);
```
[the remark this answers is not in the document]
jared-2016: I can't because size is modified in the following while loop.

uitools/common/src/AttachmentsPopupElementViewController.cpp
```diff
@@ -0,0 +1,74 @@
...
+{
+  return m_popupAttachmentItems;
+}
+
+void AttachmentsPopupElementViewController::popuplateAttachments()
+{
+  for (auto* popupAttachment : static_cast<AttachmentsPopupElement*>(popupElement())->attachments())
+  {
+    m_popupAttachmentItems->append(new PopupAttachmentItem(popupAttachment, this));
+  }
```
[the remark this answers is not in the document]
jared-2016: This only happens once. Only one possible instance of AttachmentsPopupElementViewController per popup and only one fetchAttachmentsAsync in the constructor.

uitools/common/src/Internal/PopupAttachmentImageProvider.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+    auto pixmap = icon.pixmap(32, 32);
+    thumbnail = pixmap.toImage();
+    return QQuickTextureFactory::textureFactoryForImage(thumbnail);
+  }
+#endif
+
+  if (thumbnail.isNull() && m_popupAttachmentItem->dataFetched())
+  {
+    QAbstractFileIconProvider iconProvider;
```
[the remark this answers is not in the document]
jared-2016: Wasn't aware of `QFileSystemModel` but that worked nicely. `QFileIconProvider` requires the addition of widgets which wasn't ideal.

uitools/common/src/Internal/PopupAttachmentImageProvider.cpp
jared-2016: Ah this doesn't work on mobile with the `qrc` passed into the QIcon constructor. I was following BasemapGalleryImageProvider for this, https://github.com/Esri/arcgis-maps-sdk-toolkit-qt/blob/2193418c264fb8e841d017b753f84c4aa131b572/uitools/common/src/Internal/BasemapGalleryImageProvider.cpp#L135. I am guessing it's because of how QIcon looks up the path.https://doc.qt.io/qt-6/qicon.html#QIcon-5

jared-2016:
@JamesMBallard Lots of questions regarding the ImageProvider Class so I will summarize my response in one comment, for the most part. The whole ImageProvider concept within Qt is very new to me and I used Neils work on the `BasemapGalleryImageProvider` class as my baseline, https://github.com/Esri/arcgis-maps-sdk-toolkit-qt/blob/v.next/uitools/common/src/Internal/BasemapGalleryImageProvider.cpp. So my goals with this was to use that implementation as a starting place and I tried to strip it down to what I needed in my situation.

All that to say I don't have any particular reason behind the choices other than it worked for basemap gallery so I just used that as my starting point. I'll take a good pass and apply your suggestions to the provider class. Thank you for the suggestions!

uitools/common/src/PopupAttachmentItem.cpp
```diff
@@ -0,0 +1,185 @@
...
+
+  void registerItem(PopupAttachmentItem* self)
+  {
+    PopupAttachmentImageProvider::instance()->registerItem(self);
+  }
+
+  void deregisterItem(PopupAttachmentItem* self)
+  {
+    PopupAttachmentImageProvider::instance()->deregisterItem(self);
+  }
```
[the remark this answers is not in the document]
jared-2016: This was another thing I followed from a previous example but it looks like this just make the calling code a little nicer. In any case, I don't think it's nice enough to keep and it kind of obfuscates what it's actually doing and not in a good way from my perspective.

uitools/common/src/PopupAttachmentItem.h
JamesMBallard: No, it is not technically needed here, but I still recommend you add it anyway. It is fine to have in both places if you prefer. This is preventing future mistakes. In some hypothetical future, if someone adds a new constructors but forgets to initialize that bool, that's a time-consuming error to track down.

Initializing it in the header is more future-proof, but you are correct that your branch properly handles all cases that currently exist.

uitools/common/src/PopupAttachmentItem.cpp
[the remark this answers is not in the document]
JamesMBallard: Honestly I don't have strong opinions on this. If it's needed, then keep it. I just don't understand what the registration system does. Can you explain that part?

JamesMBallard:
Thanks @jared-2016, no need to reinvent the wheel if we have something that works already. Neil laid a good foundation that is likely as solid as it gets, so unless it's too cumbersome to understand or adapt, let's go with that.

uitools/common/src/Internal/PopupAttachmentImageProvider.cpp
jared-2016: I don't believe so but I need to think about this a little more.

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/AttachmentsPopupElementView.qml
jared-2016: bugs/enhancements added and a comment about their current state.

calcite/Calcite/images/download.svg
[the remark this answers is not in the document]
jared-2016: updated.

calcite/Calcite/images/arrow-circle-down.svg
jared-2016, line 1: This is what replaced the cloud/arrow download icon. not as great but I think it gets the message across well enough and it's part of the calcite styling.

JamesMBallard (approved):

ldanzinger (approved):
LGTM

PR #687, opened 2025-02-14
Add 3d basemaps to gallery

tanneryould:
Adds support for 3D basemaps to be displayed in a Basemap Gallery when the Gallery is applied to a Scene.

### DIFF
--- a/uitools/common/src/BasemapGalleryController.cpp
+++ b/uitools/common/src/BasemapGalleryController.cpp
@@ -239,6 +239,30 @@ namespace Esri::ArcGISRuntime::Toolkit {
       }
     }
 
+    void sort3DBasemapsAndAddToGallery(BasemapGalleryController* self, BasemapListModel* basemaps)
+    {
+      qDebug() << self->geoModel();
+      // Convert BasemapListModel into a Basemap* vector and sort basemaps alphabetically using the title
+      std::vector<Basemap*> basemapsVector;
+      basemapsVector.reserve(basemaps->rowCount());
+      std::copy(std::cbegin(*basemaps), std::cend(*basemaps), std::back_inserter(basemapsVector));
+      std::sort(std::begin(basemapsVector), std::end(basemapsVector), [](Basemap* b1, Basemap* b2)
+      {
+        // Check validity of basemap->item() and if title() is empty. If either is true, push to end of list.
+        if (!b1->item() || b1->item()->title() == "")
+          return false;
+        else if (!b2->item() || b2->item()->title() == "")
+          return true;
+        else
+          return b1->item()->title() < b2->item()->title();
+      });
+
+      // For each discovered map, add it to our gallery.
+      for (auto basemap : basemapsVector)
+      {
+        self->append(basemap, true);
+      }
+    }
     /*!
       \internal
       Calls Portal::fetchDeveloperBasemapsAsync on the portal. Note that we do
@@ -264,6 +288,19 @@ namespace Esri::ArcGISRuntime::Toolkit {
                            // Notify the demo that the basemaps have changed.
                            emit self->basemapsChanged();
                          });
+
+                         if (qobject_cast<Scene*>(self->geoModel()))
+                         {
+                           portal->fetch3DBasemapsAsync().then(
+                                 [portal, self]()
+                           {
+                             // Sort and append the basemaps to the gallery.
+                             BasemapListModel* basemaps = portal->basemaps3D();
+                             sort3DBasemapsAndAddToGallery(self, basemaps);
+                             // Notify the demo that the basemaps have changed.
+                             emit self->basemapsChanged();
+                           });
+                         }
                        });
       portal->load();
     }
@@ -466,6 +503,27 @@ namespace Esri::ArcGISRuntime::Toolkit {
                      emit basemapsChanged();
                    });
                  }
+
+                 if (qobject_cast<Scene*>(m_geoModel))
+                 {
+                   if (m_portal->basemaps3D()->rowCount() > 0)
+                   {
+                     for (auto basemap : *m_portal->basemaps3D())
+                     {
+                       append(basemap);
+                     }
+                   }
+                   else
+                   {
+                     m_portal->fetch3DBasemapsAsync().then(
+                           [this]()
+                     {
+                       BasemapListModel* basemaps = m_portal->basemaps3D();
+                       sort3DBasemapsAndAddToGallery(this, basemaps);
+                       emit basemapsChanged();
+                     });
+                   }
+                 }
                });
     }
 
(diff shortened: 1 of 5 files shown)


uitools/common/src/BasemapGalleryController.cpp
```diff
@@ -466,6 +502,27 @@ namespace Esri::ArcGISRuntime::Toolkit {
...
+                   {
+                     for (auto basemap : *m_portal->basemaps3D())
+                     {
+                       append(basemap);
+                     }
+                   }
+                   else
+                   {
+                     m_portal->fetch3DBasemapsAsync().then(
+                           [this]()
```
bri12415, line 518: I think the formatting got messed up on the capture clause for the new lambdas in this file.

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/BasemapGallery.qml
```diff
@@ -233,6 +233,28 @@ Pane {
                 height: basemapGallery.internal.defaultCellSize
                 color: "transparent"
             }
+            Rectangle {
```
bri12415, line 236: Seems like a lot of magic numbers in the section below, like `-8`, `14`, `-4`. Is there anyway to base these values on other properties' size?

uitools/common/src/BasemapGalleryController.cpp
```diff
@@ -264,6 +287,19 @@ namespace Esri::ArcGISRuntime::Toolkit {
                            // Notify the demo that the basemaps have changed.
                            emit self->basemapsChanged();
                          });
+
+                         if (qobject_cast<Scene*>(self->geoModel()))
+                         {
+                           portal->fetch3DBasemapsAsync().then(
```
bri12415, line 293: If the call to `portal->fetch3DBasemapsAsync` here and `portal->fetchDeveloperBasemapsAsync` above both finish around the same time, I think they'll both call their respective sort basemaps function, and both be acting on the `BasemapGalleryController` "self" at the same time.

I think that means they can both be calling `self->append`, which eventually ends up [here](https://github.com/Esri/arcgis-maps-sdk-toolkit-qt/blob/main/uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/Internal/GenericListModel.cpp#L571) appending the `m_objects` QList at the same time, and I don't think QList is thread safe.

@JamesMBallard Does that make sense? I might be incorrect in my thinking here.

uitools/common/src/BasemapGalleryController.cpp
```diff
@@ -466,6 +502,27 @@ namespace Esri::ArcGISRuntime::Toolkit {
...
+                   if (m_portal->basemaps3D()->rowCount() > 0)
+                   {
+                     for (auto basemap : *m_portal->basemaps3D())
+                     {
+                       append(basemap);
+                     }
+                   }
+                   else
+                   {
+                     m_portal->fetch3DBasemapsAsync().then(
```
bri12415, line 517: I think the same race may be possible here as described above.

tanneryould:
I need to do some refactoring on the icon tiles because it doesn't look great on mobile 
image

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/BasemapGallery.qml
tanneryould: You're absolutely right, but I couldn't get the anchors working on the icon properly, hence the need for magic numbers. I went back though and I updated the component entirely.

uitools/common/src/BasemapGalleryController.cpp
tanneryould: @bri12415 I made use of QMetaObject's `Qt::QueuedConnection` to handle the QList::append. What do you think?

tanneryould:
image
Much better!

uitools/common/src/BasemapGalleryController.cpp
bri12415: I think it might be simpler to add a mutex in the `BasemapGalleryController::append` functions, something like

```
  bool BasemapGalleryController::append(Basemap* basemap, QImage thumbnail, QString tooltip)
  {
    std::lock_guard<std::mutex> lock(m_galleryAccessMutex);
    return m_gallery->append(new BasemapGalleryItem(basemap, std::move(thumbnail), std::move(tooltip), this));
  }
```

bri12415 (commented):
Sorry, I had forgotten to actually click the review button. My feedback is [here](https://github.com/Esri/arcgis-maps-sdk-toolkit-qt/pull/687#discussion_r1966972395)

uitools/common/src/BasemapGalleryController.cpp
JamesMBallard: >@JamesMBallard Does that make sense? I might be incorrect in my thinking here.

Sorry, I missed this notification. @bri12415 is right that QList is not thread safe. You can get into trouble mutating it from multiple threads at the same time.

However, if using QFuture continuations and passing the [`this`](https://doc.qt.io/qt-6/qfuture.html#then-1) arg, it should queue up the logic to "this"'s thread and synchronize it automatically. That's one way, but Brian's idea of using a mutex is more explicit. I like the mutex idea.

uitools/common/src/BasemapGalleryItem.cpp
```diff
@@ -295,6 +311,16 @@ namespace Esri::ArcGISRuntime::Toolkit {
     m_tooltip = std::move(tooltip);
     emit tooltipOverrideChanged();
   }
+  
+  bool BasemapGalleryItem::is3D() const
+  {
+      return m_is3D;
```
ldanzinger, line 317: indentation is different here

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/BasemapGallery.qml
```diff
@@ -226,36 +227,112 @@ Pane {
...
+                    verticalAlignment: basemapDelegate.isGrid ? Qt.AlignTop : Qt.AlignVCenter
+                    horizontalAlignment: basemapDelegate.isGrid ? Qt.AlignHCenter : Qt.AlignLeft
+                }
+
+                // 3D Tag
+                Rectangle {
+                    id: basemap3DTag
+                    anchors {
+                        top: basemapIcon.top;
+                        right: basemapIcon.right;
```
ldanzinger, line 262:
```suggestion
                        top: basemapIcon.top
                        right: basemapIcon.right
```

ldanzinger (approved):

uitools/common/src/BasemapGalleryController.cpp
```diff
@@ -224,18 +224,18 @@ namespace Esri::ArcGISRuntime::Toolkit {
...
                     return false;
-                  else if (!b2->item() || b2->item()->title() == "")
+                  else if (!b2->item() || b2->item()->title().isEmpty())
                     return true;
                   else
                     return b1->item()->title() < b2->item()->title();
                 });
 
-      // For each discovered map, add it to our gallery.
+      // Ensure appending happens on the main thread
```
bri12415, line 235:
```suggestion
```

bri12415 (approved):

PR #691, opened 2025-03-14
Deprecate all the AR components.

JamesMBallard:
This PR deprecates in doc and in code all AR components. Anyone using the pri file will get a qmake warning as well as compiler warnings.

PR #692, opened 2025-03-18
James/doc shuffle internal classes

JamesMBallard:
* For Qt Quick, all the cpp classes are internal. Only the Qml files are expected to be used directly.
 * Shuffle source files around to remove the common/src/internal folder since it's all internal source code.
 * This also deprecates the AR classes in doc. I had only done that in sources with https://github.com/Esri/arcgis-maps-sdk-toolkit-qt/pull/691
 * Build-related changes to absorb the shuffling of the source files.

I tested the Qt Quick changes with the existing toolkit example, and with a toolkit locally consuming the toolkit for widgets. For widgets I only made sure it builds.


uitools/examples/src/BaseDemo.cpp
```diff
@@ -22,7 +23,6 @@
 #include <MapViewTypes.h>
 #include <Scene.h>
 #include <SceneQuickView.h>
-#include <stdexcept>
```
JamesMBallard, line 25: I couldn't find anything needing this. I didn't go bananas pruning includes, but this one seemed as safe as it gets.

PR #693, opened 2025-03-19
Jared/signal data

jared-2016:
Graphs on iOS is still under investigation but getting the PR up as I don't believe it's related to anything i've done. 

What is encapsulated in this PR:
1. Added full screen viewing for MediaPopupElements (Images and Graphs)
2. Hovering for Graphs with a Qt Popup that displays the value of the graph component being hovered over(except for `PieSeires`, bug will be logged and added to the code as a comment once it's submitted)
3. Hand with finger pointing when hovering over and item that will do something if clicked or tapped(Media Images, Graphs, hyper links)
4. properties in PopupView for users to disable default behaviors for opening urls, displaying images in a full screen view, opening attachments externally.
5. Added signals that bubble up to PopupViewController so users can connect to them and handle these scenarios in `4.` that can be disabled.
5.1. two signals for general url's when hyper links or images, `sourceUrl` for the image displayed and `linkUrl` for when the Image is clicked on.
5.2. signal for attachments data once they have been fetched.

Some warning cleanup was done as well.


uitools/common/src/ImagePopupMediaItem.h
```diff
@@ -41,6 +41,8 @@ class ImagePopupMediaItem : public PopupMediaItem
 
 signals:
   void imagePopupMediaItemChanged();
+  void clickedUrl(const QUrl& url);
```
jared-2016, line 44: This is used for the `linkUrl` for when an image is displayed, the user can click on it to view more information.
translates to the following image of the WebMap

uitools/common/src/ImagePopupMediaItem.h
```diff
@@ -41,6 +41,8 @@ class ImagePopupMediaItem : public PopupMediaItem
...
+  void mediaImageSourceUrl(const QUrl& sourceUrl);
```
jared-2016, line 45: This is the sourceUrl for the image displayed. translates to the following images of the WebMap:

ldanzinger:
> Graphs on iOS is still under investigation

@jared-2016 can you elaborate on what is under investigation?

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/AttachmentsPopupElementView.qml
```diff
@@ -132,25 +128,39 @@ ColumnLayout {
...
-                                Qt.openUrlExternally(model.listModelData.localData);
+                            if (popupView.openAttachmentsExternally) {
+                                // This is a known limitation iOS but there are a few mentioned workarounds, https://bugreports.qt.io/browse/QTBUG-42942
+                                // Android currnetly has a bug logged against this, https://bugreports.qt.io/browse/QTBUG-133702
+                                // We plan on supporting these platforms in the future when it's implemented on iOS and fixed on Android.
+                                if (Qt.platform.os !== "android" && Qt.platform.os !== "ios") {
+                                    Qt.openUrlExternally(model.listModelData.localData);
+                                }
+                            } else {
+                                // user disabled default behavior, so we do nothing
```
ldanzinger, line 146: does the clicked event for both cases? I don't see where that happens

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/MediaPopupElementView.qml
```diff
@@ -37,9 +36,90 @@ ColumnLayout {
...
+
+        ColumnLayout {
+            anchors.fill: parent
+            Layout.alignment: Qt.AlignHCenter
+
+            Label {
+                visible: fullScreenImageDialog.imageTitle !== ""
+                text: fullScreenImageDialog.imageTitle
+                font.weight: Font.Bold
+                color: "black"
```
ldanzinger, line 65: do these hardcoded colors work in light/dark mode?

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/PopupView.qml
```diff
@@ -84,6 +85,30 @@ Page {
...
+    /*!
+       \brief Boolean that controls if the PopupView will attempt to open URL's with an external browser.
+
+       Defaults to true.
+       \qmlproperty bool openUrlsExternally
+     */
+    property bool openUrlsExternally: true
+
+    /*!
+       \brief Boolean that controls if the PopupView will attempt to open images with an internal viewer.
```
ldanzinger, line 97: would it be accurate to say "in a full screen takeover", or something similar

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/PopupView.qml
```diff
@@ -84,6 +85,30 @@ Page {
...
+
+       Defaults to true.
+       \qmlproperty bool openImagesInternally
+     */
+    property bool openImagesInternally: true
```
ldanzinger, line 102: maybe `openImagesInApp` or similar would be a little more descriptive?

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/PopupView.qml
```diff
@@ -84,6 +85,30 @@ Page {
...
+
+    /*!
+       \brief Boolean that controls if the PopupView will attempt to open attachments with an external viewer.
```
ldanzinger, line 105: perhaps link to the qt open method, which should provide more details about what behavior they can expect

uitools/common/src/AttachmentsPopupElementViewController.cpp
```diff
@@ -49,7 +49,8 @@ AttachmentsPopupElementViewController::~AttachmentsPopupElementViewController()
 
 QString AttachmentsPopupElementViewController::title() const
 {
-  return static_cast<AttachmentsPopupElement*>(popupElement())->title();
+  const auto title = static_cast<AttachmentsPopupElement*>(popupElement())->title();
+  return title != QStringLiteral("") ? title : QStringLiteral("Attachments");
```
JamesMBallard, line 53:
```suggestion
  return !title.isEmpty() ? title : QStringLiteral("Attachments");
```

uitools/common/src/FieldsPopupElementViewController.cpp
```diff
@@ -36,7 +36,8 @@ FieldsPopupElementViewController::~FieldsPopupElementViewController() = default;
 
 QString FieldsPopupElementViewController::title() const
 {
-  return static_cast<FieldsPopupElement*>(popupElement())->title();
+  const auto title = static_cast<FieldsPopupElement*>(popupElement())->title();
+  return title != QStringLiteral("") ? title : QStringLiteral("Fields");
```
JamesMBallard, line 40: Use `isEmpty` as the standard "empty string" check.

jared-2016:
> > Graphs on iOS is still under investigation
> 
> @jared-2016 can you elaborate on what is under investigation?

@ldanzinger Only LineSeries is working currently. Isolating it to see if it's a qt bug or something with my implementation. Seems likely to be a qt bug I must have missed previously. If it is a bug, the fallback for iOS will be to use RTC generated graphs.

uitools/common/src/AttachmentsPopupElementViewController.h
```diff
@@ -48,6 +48,7 @@ class AttachmentsPopupElementViewController : public PopupElementViewItem
 
 signals:
   void attachmentPopupElementChanged();
+  void attachmentDataFetched(const QByteArray& attachmentData, const QString& name);
```
JamesMBallard, line 51: Is this mandatory? I assume large files are moving around a lot of data here (MBs?), especially photos and videos. Who is consuming this data? If this is "just in case someone wants it", I would consider using adding some additional logic to only emit if someone is connected.

That is nontrivial to add, so we should discuss before you make any decisions on this.

This is very similar to the workflow mentioned in the doc [here](https://doc.qt.io/qt-6/qobject.html#isSignalConnected): 
>you can use this function to avoid expensive initialization or emitting a signal that nobody listens to

uitools/common/src/MediaPopupElementViewController.cpp
```diff
@@ -95,7 +112,8 @@ QString MediaPopupElementViewController::description() const
 
 QString MediaPopupElementViewController::title() const
 {
-  return static_cast<MediaPopupElement*>(popupElement())->title();
+  auto title = static_cast<MediaPopupElement*>(popupElement())->title();
+  return title != QStringLiteral("") ? title : QStringLiteral("Media");
```
JamesMBallard, line 116: use `isEmpty`. I won't comment on each if there are more.

uitools/common/src/PopupAttachmentItem.cpp
```diff
@@ -136,8 +135,16 @@ void PopupAttachmentItem::downloadAttachment()
...
+  m_popupAttachment->attachment()->fetchDataAsync().then([this] (const QByteArray& attachmentData)
   {
+    if (attachmentData.isEmpty())
+    {
+      m_fetchingAttachment = false;
+      emit popupAttachmentItemChanged();
+      return;
+    }
+
+    emit attachmentDataFetched(attachmentData, name());
```
JamesMBallard, line 148: Ditto - what needs this data? If it is necessary for the popup to function, then we need to keep it.

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/AttachmentsPopupElementView.qml
jared-2016: The signal is emitted from the `fetchDataAsync` call in PopupAttachmentItem, https://github.com/Esri/arcgis-maps-sdk-toolkit-qt/blob/b270df915e5cb408ee47cb5e1ce5a5faf5a6bbcb/uitools/common/src/PopupAttachmentItem.cpp#L147

i just added a comment there for clarification as i did it everywhere else.

uitools/common/src/PopupViewController.cpp
```diff
@@ -137,26 +137,54 @@ void PopupViewController::setPopup(Popup* popup)
...
       {
         case Esri::ArcGISRuntime::PopupElementType::TextPopupElement:
-          m_popupElementControllerModel->append(
-                new TextPopupElementViewController(static_cast<TextPopupElement*>(element), m_popup));
+        {
+          const auto controller = new TextPopupElementViewController(static_cast<TextPopupElement*>(element), m_popup);
+          m_popupElementControllerModel->append(controller);
+          // bubble up signal to controller for users to connect too
+          connect(controller, &TextPopupElementViewController::clickedUrl,
+                  this, &PopupViewController::clickedUrl);
```
JamesMBallard, line 145: I am seeing a pattern emerge that looks like it could lead to maintenance issues in the future. A few times I have seen objects get created, followed by comments like
```
// bubble up signal to controller...
```
and a connect statement. 

I am sensing that this could be brittle, and if someone else later on needs to work on this code, they will not know to do this.

Is it possible to tweak the constructor for `TextPopupElementViewController` (for example) to take the `PopupViewController` and have `TextPopupElementViewController` itself wire up the signal? That way, the code enforces that this must happen, and future devs in this codebase won't need to remember anything (hopefully).

I could be oversimplifying things and maybe that won't work, or won't expand nicely to other classes.

Give it some thought. Maybe there's a better or different way to do this?

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/PopupView.qml
```diff
@@ -65,6 +65,7 @@ Page {
     id: popupView
 
     /*!
+       \deprecated
```
JamesMBallard, line 68: Is this intentional?

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/PopupView.qml
JamesMBallard: I feel similarly about these. It is not obvious what externally and internally mean.

jared-2016: @ldanzinger @JamesMBallard how about `openUrlsWithSystemDefaultApplication`. It's long but descriptive. Attachments would be `openAttachmentsWithSystemDefaultApplication`.

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/PopupView.qml
jared-2016: Yes, we plan on the deprecating popupManager with the release of the PopupElements toolkit components. 

Probably better to separate the depreciation of popupManager and the rest that goes with it into it's own PR. I'll take it out of this one.

uitools/common/src/AttachmentsPopupElementViewController.h
jared-2016: Emitting the data as a signal I would say is mandatory. For mobile platforms the only attachments we can do anything with are images which we can display with a full screen takeover. The rest just download and do nothing currently. We want the user to be able to connect to the signal and do what they want with the data. It's a very good point that you made though that it's MB's being moved around and could be a lot more in the future so we should be cognizant of that.

I like the idea of only emit when connected so the user isn't paying the price of moving around large files for nothing. Granted in my testing of attachments I was unable to upload attachments bigger than ~20MB. I didn't try all sizes in between but i eventually got a ~10MB picture to upload so it's somewhere between 10-20MB as far as I was able to test. Just wanted to add that additional information.

Not sure what it would take to only emit when connected but I think it's worth while. Maybe it's not something we need to do in this PR since the maximum file size seems limited to something reasonable but it's something we should eventually do. I would like to get a more definite answer on the file size limits first.

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/PopupView.qml
JamesMBallard: >how about openUrlsWithSystemDefaultApplication. It's long but descriptive. Attachments would be openAttachmentsWithSystemDefaultApplication.

I like that better.

uitools/common/src/AttachmentsPopupElementViewController.h
JamesMBallard: I want to clarify my previous statement. It sounds like there is some nuance here I didn't understand.

1. We are emitting to enable things to work correctly.
2. We are emitting in case users want to do something with that information, but it isn't necessary for things to work properly.

For 1, no question we don't want to break things. For 2, I do not think we should emit unless someone is connected. Possibly `QObject::isSignalConnected` is enough to check that. Just a handful of photos could mean pushing around many MBs of data for no reason.

uitools/common/src/PopupViewController.cpp
jared-2016: Done and it worked great. Thanks for the suggestion!

uitools/common/src/AttachmentsPopupElementViewController.h
jared-2016: Makes perfect sense. I'll look into `QObject::isSignalConnected`.

> signal must be a signal member of this object, otherwise the behaviour is undefined.

This might be tricky to work with as I have to make connections to make the signal accessible to the user but the emission happens all the way down from `PopupViewController->AttachmentsPopupViewController->PopupAttachmentItem
calling `QObject::isSignalConnected(QMetaMethod::fromSignal(&PopupAttachmentItem::attachmentDataFetched))` I think this will always return true since there is the above mentioned plumbing.

JamesMBallard: Ok, in that case this will get more complicated. We do this in the API, but this is where more internal plumbing is needed to ensure all the multiple levels of connections cooperate. 

It's inefficient, but adding all this extra logic will take more time and complicate things. I think we can live without it for now.

QByteArray is implicitly shared (I forgot about that), so if the data already exists and we send a copy of the QByteArray somewhere that does nothing with it, that should not incur overhead like I worried it might.

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/PopupView.qml
ldanzinger: Yeah, something along those lines works for me

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/MediaPopupElementView.qml
jared-2016: good catch!

uitools/common/src/AttachmentsPopupElementViewController.h
jared-2016: Yeah this data exists no matter what since we must call `fetchDataAsync` which returns the QByteArray in the QFuture. Guess this is less of a worry than we though, good to know.

jared-2016:
> > > Graphs on iOS is still under investigation
> > 
> > 
> > @jared-2016 can you elaborate on what is under investigation?
> 
> @ldanzinger Only LineSeries is working currently. Isolating it to see if it's a qt bug or something with my implementation. Seems likely to be a qt bug I must have missed previously. If it is a bug, the fallback for iOS will be to use RTC generated graphs.

@ldanzinger Getting closer. It seems when the data is requested from QML it creates it just fine but when I access the `QList<>::length` property in QML it's coming back undefined on iOS only. Seems like a Qt bug to me, ill put in a workaround for this in the mean time.

jared-2016:
Looks like QVariantList might be a better way of exposing the list to qml. I gave that a try on ios and I don't get undefined anymore.

jared-2016:
Made one other general UI enhancement for Labels. Fixed the Graphs issue on ios, e.g. switched from QList<PieSlices*> to QVariantList and the same goes for all other graphs.

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/MediaPopupElementView.qml
```diff
@@ -177,25 +175,32 @@ ColumnLayout {
...
-                id: loader
+                id: mediaLoader
             }
 
             MouseArea {
                 anchors.fill: parent
 
                 onClicked: {
-                    model.listModelData.mediaImageSourceUrl(model.listModelData.sourceUrl);
+                    if(model.popupMediaType === QmlEnums.PopupMediaTypeImage) {
```
ldanzinger, line 185:
```suggestion
                    if (model.popupMediaType === QmlEnums.PopupMediaTypeImage) {
```

ldanzinger (approved):
LGTM

JamesMBallard (approved):

PR #698, opened 2025-03-21
Jared/cmake

jared-2016:
I think given the finishing of PopupElements and the signals/properties now being exposed it would be a good time to merge into v.next so I don't have to maintain the WIP branch any longer. It's getting more and more complicated with some of the refactoring work @JamesMBallard has done recently.

In order to test cmake it was easier to start from v.next and merge the latest into it and make the necessary fixes. This PR includes:
1. CMake for Popup Elements(tested on mobile and desktop)
2. moved the svgs used to the toolkit common images so they can be used without calcite

All of the controllers, views, and support classes have been reviewed so i would just do a quick glance.

Things to look out for:
1. Unrelated files have been changed, I think I caught all of these but just wanted to point it out.
2. Any doc that might have slipped back in since this work was done prior to James refactoring the doc. I think I caught it all but a 2nd/3rd set of eyes doesn't hurt.


uitools/common/src/PopupViewController.cpp
```diff
@@ -169,4 +254,18 @@ void PopupViewController::setAttachmentThumbnailHeight(int height)
...
+
+/*!
+  \fn void Esri::ArcGISRuntime::Toolkit::PopupViewController::clickedUrl(const QUrl& url)
+  \brief Signal emitted when a URL has been clicked. This is exposed to the user if they want to define their own handler for the url.
+ */
+
+/*!
+  \fn void Esri::ArcGISRuntime::Toolkit::PopupViewController::mediaImageSourceUrl(const QUrl& sourceUrl)
+  \brief Signal emitted when a media image has been clicked. This is exposed to the user if they want to define their own handler for the image.
+ */
```
JamesMBallard, line 270: I think this was removed intentionally.
```suggestion
```

JamesMBallard (approved):

ldanzinger:
> moved the svgs used to the toolkit common images so they can be used without calcite

I see all the svgs in a calcite folder/qrc with this PR?

ldanzinger (commented):
Checked over the files modified at a high level, everything looks in place, except for my question about Calcite

jared-2016:
> > moved the svgs used to the toolkit common images so they can be used without calcite
> 
> I see all the svgs in a calcite folder/qrc with this PR? Do they need to be there and Common, or can they just be in common?

@ldanzinger They are only used by Popups so they only need to reside in common. They are pulled from the calcite styling so I was split on if they should also live in the Calcite portion of the toolkit or not. Nothing else uses them so I am leaning towards only common inside the toolkit components.

I'll get a change up to remove the references from calcite.

ldanzinger:
@jared-2016  I lean towards putting them just in the one place we know we need them now (common), and if we get use cases for needing these in our calcite style later, we can figure out a solution at that point

ldanzinger (approved):

jared-2016:
@ldanzinger @JamesMBallard 

If you could both take a look at the last two commit's that would be great.

1. James made a good observation about one of the signals during a call so i changed how data was passed when an image was clicked. Now when a Image Popup Media is clicked one signal is emitted with all relevant data(sourceUrl and linkUrl).

2. Since controllers are not documented, i re-emitted the signals from the controller in the view itself so it's more readily available for the user. I added documentation to go with the signals.

Tested these changes on ios, android, and macos.

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/PopupView.qml
```diff
@@ -99,12 +132,57 @@ Page {
...
+       \brief Signal emitted when a url or hyperlink is clicked.
+       The \a url of the hyperlink from the Popup that was clicked on.
+     */
+    signal clickedUrl(var url)
+
+    /*!
+       \qmlsignal PopupView::imageClicked(var sourceUrl, var linkUrl)
+       \brief Signal emitted when a Image Popup Media is clicked.
+       The \a sourceUrl of the image that was clicked on.
+       The \a linkUrl of the image that was clicked on.
```
JamesMBallard, line 154: I would document what the difference is between the two args.

JamesMBallard (approved):
One comment about the doc, but otherwise looks good.

ldanzinger (approved):

uitools/common/src/PieChartPopupMediaItem.cpp
```diff
@@ -0,0 +1,85 @@
...
+  int green = sliceColor.green();
+  int blue = sliceColor.blue();
+
+  // Calculate the relative luminance of the color using
+  // Rec. 709, https://en.wikipedia.org/wiki/Rec._709
+  double luminance = (0.2126 * red + 0.7152 * green + 0.0722 * blue) / 255;
+
+  // Check if the luminance is closer to black or white
+  return (luminance > 0.5) ? QColor(0, 0, 0) : QColor(255, 255, 255);
+}
```
jared-2016, line 44: algorithm for determining our high contrast text color. Bar Graph doesn't need it(adapts automatically). LineSeries doesn't have a way of doing this nicely. The individual points don't have a concept of an associate label like the rest of the Graph series.

PR #701, opened 2025-03-25
fix popup view scroll behavior

jared-2016:
Prevents the wheel event from propagating from the PopupView to the MapView.


uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/PopupView.qml
```diff
@@ -155,6 +155,14 @@ Page {
      */
     signal imageClicked(var sourceUrl, var linkUrl)
 
+    MouseArea {
+        anchors.fill: parent
+        // stop whell event from scrolling the map once we reach the end of the PopupView
```
JamesMBallard, line 160:
```suggestion
        // stop wheel event from scrolling the map once we reach the end of the PopupView
```

PR #704, opened 2025-03-26
Jared/fix fields popup

jared-2016:
Waiting on @bri12415 to confirm the manual setting of the hyper link color works for him.

One other minor change is setting a maximum height on the busy indicator.

### DIFF
--- a/uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/AttachmentsPopupElementView.qml
+++ b/uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/AttachmentsPopupElementView.qml
@@ -214,6 +214,7 @@ ColumnLayout {
                     id: fetchAttachment
                     running: model.listModelData.fetchingAttachment
                     visible: model.listModelData.fetchingAttachment
+                    Layout.maximumHeight: fileInfoColumn.height
                 }
 
                 MenuSeparator {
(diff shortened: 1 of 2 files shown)


uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/AttachmentsPopupElementView.qml
```diff
@@ -214,6 +214,7 @@ ColumnLayout {
                     id: fetchAttachment
                     running: model.listModelData.fetchingAttachment
                     visible: model.listModelData.fetchingAttachment
+                    Layout.maximumHeight: fileInfoColumn.height
```
jared-2016, line 217: I did a quick test of a few different styles and noticed that the styling size for the busy indicator changed its size dramatically which would alter the grid layout row size but would not go back down dynamically. Setting the maximum size on this ensures consistency regardless of style. If the icon is small it fits just fine but it won't get overly large and cause distortion of the grid layout now. Low risk as it just set's an upper limit that's used elsewhere on elements in this grid layout.

bri12415:
Here is what it looks like now on my end. Pixel 7.

Screenshot_20250326-123851

Screenshot_20250326-123851

jared-2016:
It slipped my mind but FIeldsPopupElementView is not the only view that displays hyper links. The TextPopupElementView also could contain hyper links and that's not a simple to fix. We have to manually create hyperlinks for the fields view because the data comes in as a url but we would have to search and fix the TextPopupElementView which makes me lean towards not including the color change.

JamesMBallard (approved):

jared-2016:
maybe it's not as difficult as I thought to modify the text string. Let's see what the robot says.

jared-2016:
Easier than I thought. I don't like that we have to do it but it's a better guaranteed experience across all platforms.

JamesMBallard (approved):

ldanzinger (approved):

PR #706, opened 2025-03-27
fix menu separators. consistent across all views.

jared-2016:
Adjusted the menu separators. Attachments was doing the same thing so I made the fix there as well.


jared-2016:
I think this might be the last one?

aji13187 (approved):

PR #709, opened 2025-04-01
Disable deprecation warnings for PopupViewController.cpp

bri12415:
Disable deprecation warnings stemming from our internal `PopupViewController.cpp` source file.

### DIFF
--- a/uitools/common/src/PopupViewController.cpp
+++ b/uitools/common/src/PopupViewController.cpp
@@ -14,6 +14,8 @@
  *  See the License for the specific language governing permissions and
  *  limitations under the License.
  ******************************************************************************/
+#define QRT_DISABLE_DEPRECATED_WARNINGS
+
 #include "PopupViewController.h"
 
 // Qt headers


JamesMBallard:
Does this get the warnings down to zero? There may be other files to apply the same pattern to.

bri12415:
> Does this get the warnings down to zero? There may be other files to apply the same pattern to.

No, there are still existing warnings. I intentionally didn't add this elsewhere for those warnings, heres why:

* `PopupViewController.h`: If we add that define here, I think users won't get actual deprecation warnings if they use a deprecated function declared in this header. This is where the majority of the warnings are from. 
* `PopupAttachmentListModel.h`: The only constructor available for this class is one that takes a deprecated class. I'm not sure if it would make sense or not have that define here. Users don't really have an option not to use that constructor, so maybe it does make sense to use that define here.
* `PopupViewDemo.cpp`: This demo does invoke deprecated functions. I'm not sure if we consider this code internal or not. If we do, then I guess it does make sense to add the define here.

JamesMBallard:
I want to make a distinction between these codebases:
1.  The toolkit implementation .cpp files.
2. The toolkit public API that user work with directly - its headers and qml files that we document.

I would like to address all warnings in 1. If there are toolkit deprecations (category 2), user code should see those warnings.

We should not add `#define QRT_DISABLE_DEPRECATED_WARNINGS` to PopupViewController.h, or any toolkit header since it will spread that decision to other, unrelated areas.

Is the `PopupAttachmentListModel.h` header exposed via other toolkit headers? If it is only included in toolkit .cpp files, we may be able to hide warnings related to it. Unclear, and it may be possible but not worth the code complications.

PopupViewDemo.cpp is gray area. I don't think we should hide warnings, but I don't know which warnings are appearing from that file. Can you share more details about this one?

bri12415:
> Is the PopupAttachmentListModel.h header exposed via other toolkit headers?

I don't think so, it's [forward declared](https://github.com/Esri/arcgis-maps-sdk-toolkit-qt/blob/main/uitools/toolkitcpp/cpp/Esri/ArcGISRuntime/Toolkit/PopupViewController.h#L31) in `PopupViewContoller.h` (but also used in a `Q_MOC_INCLUDE` as well), and then actually included in `PopupViewController.cpp`. This is its only usage.

> Can you share more details about this one?

Yes, in `PopupViewDemo.cpp` we [instantiate](https://github.com/Esri/arcgis-maps-sdk-toolkit-qt/blob/main/uitools/examples/src/demos/PopupViewDemo.cpp#L139) a new `PopupManager` which is deprecated and so is throwing that warning.

JamesMBallard:
Are there any reasonable mitigations for the warnings coming from PopupAttachmentListModel.h? If that is only coming from PopupViewController.cpp, I don't see why `#define QRT_DISABLE_DEPRECATED_WARNINGS` wouldn't disable it.

bri12415:
> If that is only coming from PopupViewController.cpp, I don't see why #define QRT_DISABLE_DEPRECATED_WARNINGS wouldn't disable it.

I don't believe the warnings are coming from the `PopupViewController.cpp` file, they're coming from the `PopupAttachmentListModel.h` file itself, because its using deprecated classes.

JamesMBallard:
If you restructure the includes for PopupViewController.h like this, does it get rid of those warnings?
```c++
// Qt headers
#include <QAbstractListModel>
#include <QObject>
#include <QPointer>

// Maps SDK headers
#include <Deprecated.h>
#include <Popup.h>
#include <PopupElement.h>

// Other headers
#include "GenericListModel.h"

Q_MOC_INCLUDE("PopupAttachmentListModel.h")
Q_MOC_INCLUDE("PopupManager.h")
```

So, drop the direct include of `PopupManager.h`, but add a `Q_MOC_INCLUDE` for it since it's used directly in a Q_PROPERTY.

bri12415:
Yes, that does remove those warnings. I'll add that it now.

JamesMBallard (approved):

uitools/common/src/PopupViewController.h
```diff
@@ -25,12 +25,12 @@
 #include <Deprecated.h>
 #include <Popup.h>
 #include <PopupElement.h>
-#include <PopupManager.h>
 
 // Other headers
 #include "GenericListModel.h"
 
 Q_MOC_INCLUDE("PopupAttachmentListModel.h")
+Q_MOC_INCLUDE("PopupManager.h")
```
bri12415, line 33: This was already forward declared below fyi.

PR #715, opened 2025-04-09
calcite add property to honor system theme.

jared-2016:
Added boolean `honorSystemTheme` property to calcite which will override the existing Calcite dark/light mode toggle if true which is the default state. Otherwise the user can specify the property to be false and maintain existing behavior.

Unfortunately this wasn't an option originally but now that Qt provides a way to honor system theme, we should use it. Example apps that utilize calcite have been updated accordingly. 

Tested on Android, iOS, and macOS.
systemThemeEx


calcite/Calcite/Calcite.qml
```diff
@@ -26,12 +26,18 @@ import QtQuick
...
-    }
+    Unknown,
+    Light,
+    Dark
+}
+
+    property int theme: honorSystemTheme ? Application.styleHints.colorScheme : Calcite.Theme.Light
 
-    property int theme: Calcite.Theme.Light
+    property bool honorSystemTheme: true
```
JamesMBallard, line 40: This is just wording preference, but I would go with `useSystemTheme` or something like that. I don't think `honorSystemTheme` is bad and I am good if we stick with it.

calcite/docs/dark_mode.png
JamesMBallard, line 1: Both screenshots are significantly smaller now. Not sure if that will have any negative impacts anywhere.

JamesMBallard (commented):
Mostly looks good, but I am not sure what "Enabled" means in that demo gif. Also I am unsure why the "Honor System Theme" checkbox is "checked" or "unchecked but line through it". That may be a Qt control question, but it looks odd.

calcite/Calcite/Calcite.qml
jared-2016: > `honorSystemTheme` is bad

Say no more 😉 . In all seriousness, I like useSystemTheme better as well.

calcite/docs/dark_mode.png
jared-2016: I just took the default window size but I can re-do it to a similar size easily enough :)

jared-2016, line 1: I didn't want to cover over the watermark. But the size is much closer. Yay or Nay? @JamesMBallard @ldanzinger

calcite/Calcite/Calcite.qml
```diff
@@ -26,12 +26,18 @@ import QtQuick
...
+    // Unknown was added it to more accurately match Application.styleHints.colorScheme
+    // This is an internal implementation detail.
+    // https://doc.qt.io/qt-6/qt.html#ColorScheme-enum
     enum Theme {
-        Light,
-        Dark
-    }
+    Unknown,
+    Light,
+    Dark
```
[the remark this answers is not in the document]
jared-2016: I reorganized them to match Qt, https://doc.qt.io/qt-6/qt.html#ColorScheme-enum so the existing logic to determine the styles color choices below would align and still work if either Calcite theme enum is used or the Qt style hints enum. From a users perspective I imagine this wouldn't change anything as long as they are using the enums.

jared-2016:
> Mostly looks good, but I am not sure what "Enabled" means in that demo gif. Also I am unsure why the "Honor System Theme" checkbox is "checked" or "unchecked but line through it". That may be a Qt control question, but it looks odd.

@JamesMBallard If you are talking about the images for the readme, I was split between showing it honoring system theme vs the calcite theme toggle. The first set of pictures was showing the toggle which is why honor system theme had the dash `-` through it. The specific styling, I believe it's just mimicking the calcite styling since the indicator is defined. https://github.com/Esri/arcgis-maps-sdk-toolkit-qt/blob/f730922a5703f3c5bc248d9aeaef9b4fe00baca2/calcite/Calcite/CheckBox.qml#L30

Enabled just enables/disables some but not all of the controls in this demo app 🤷

JamesMBallard:
>...vs the calcite theme toggle

Ok that's what it is. I was confused because "enabled" could be "turn dark mode on", but that helps me understand. This is just for the demo app so I think it's ok.

calcite/docs/dark_mode.png
JamesMBallard: New images look good to me.

ldanzinger (approved):

PR #716, opened 2025-04-10
Jared/iosffmpeg fix for calcite demo app

jared-2016:
in doing my other calcite work I noticed this was missing from this apps project file.

### DIFF
--- a/calcite/demo/calcite_test.pro
+++ b/calcite/demo/calcite_test.pro
@@ -74,6 +74,32 @@ macx {
 
 ios {
     include (iOS/iOS.pri)
+
+    # workaround for https://bugreports.qt.io/browse/QTBUG-129651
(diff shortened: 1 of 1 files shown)


JamesMBallard (commented):
I don't know if we should do this or not. Qt now has a qmake solution, and we're planning to incorporate it this release. This will be another thing we need to roll back later if we do it now. For comparison, we put this in for sample viewer, but not each individual sample.

Thoughts?

jared-2016:
>Qt now has a qmake solution, and we're planning to incorporate it this release

I forgot about this. Let's hold off on this then, is there an issue I can update to make sure this project isn't missed?

jared-2016:
will opt for Qt's qmake solution when the time is right. https://bugreports.qt.io/browse/QTBUG-129651

PR #720, opened 2025-04-28
PKI Client Certificate toolkit component

JamesMBallard:
This is the new authentication system PKI toolkit component. This is referred to as "ClientCertificate" authentication.

Here is the workflow:
1. Detect ClientCertificate challenge, display UI for users.
2. Once user selects the certificate, we attempt it immediately with no password. Same as the previous toolkit component.
3. If that doesn't work, we prompt the user for a password.
4. Keep iterating on that until the password is correct, or the user cancels.

Differences compared to the existing toolkit component:
 * There is no longer a list of certificates. That is not how the NetworkCredentialStore is structured. There is now one cert associated with one host, so you can only select and attempt one at a time. The NetworkCredentialStore will house all certificates after they are added.
 * Purely an implementation detail, but the password dialog is now a persisted component that is invisible by default. That allowed managing state better.
 * If the user enters a password that doesn't work, we display `Invalid password - try again` in red text.
 
Worth noting but the new controller class is getting very large. It's intended to be the one-stop-shop for all authentication, and users don't need to interact with it directly. I mention this only in the sense that lots of more additions and we will need to consider breaking it down somehow. Perhaps more composition to encapsulate more.

## ~~No automation yet~~

~~I plan to try to add automation support but it's not there yet. If it ends up being a lot, I will consider splitting it into a second PR. It's not clear if this is automatable due to the number of possible paths users could take.~~

It wasn't that bad. Added to this PR.


uitools/common/src/ArcGISAuthenticationController.cpp
```diff
@@ -163,7 +165,17 @@ void ArcGISAuthenticationController::handleNetworkAuthenticationChallenge(Networ
     }
     case NetworkChallengeType::ClientCertificate:
     {
-      Q_UNIMPLEMENTED();
+      const auto sslBackend = QSslSocket::activeBackend();
+      if (QSslSocket::activeBackend() != "openssl")
+      {
+        qWarning() << QString("ClientCertificate authentication is not supported with the current SSL backend (%1). ").arg(sslBackend) +
```
jared-2016, line 171: Would `QStringLiteral` work here? I tend to avoid them when I have runtime args but I am curious so I wanted to ask.

jared-2016 (approved):
One question but LGTM.

uitools/common/src/ArcGISAuthenticationController.cpp
JamesMBallard: I pushed the change. I also realized that just logging to the console is insufficient and we need to fail the challenge since we cannot handle it in these cases.

PR #723, opened 2025-05-21
Out of process OAuth2 on mobile

JamesMBallard:
This implements the OAuth out of process browser login via Qt's newer network authentication OAuth classes.

Historically OAuth has always been an in-process experience with Qt, primarily because it was quite complicated to implement on our own.

Now with these changes, on iOS and Android we can support the out-of-process browser login experience by leveraging:
 * [QOAuthUriSchemeReplyHandler](https://doc.qt.io/qt-6/qoauthurischemereplyhandler.html)
 * [QAbstractOAuth2](https://doc.qt.io/qt-6/qabstractoauth2.html) (a custom implementation)

With Qt's concrete classes they handle everything, which is too much for our use case. That's why we have a custom implementation of `QAbstractOAuth2`.

This does not impact the current in-process workflows. We just look for specific redirectURI to determine which method to attempt.

The bulk of the logic is straight from Qt's example logic, seen [here](https://doc.qt.io/qt-6/qoauthurischemereplyhandler.html#details).


uitools/common/src/CustomOAuth2AuthorizationCodeFlow.cpp
```diff
@@ -0,0 +1,48 @@
+
```
aji13187, line 1: An extra newline

uitools/common/src/ArcGISAuthenticationController.cpp
```diff
@@ -23,6 +23,8 @@
 #include <QSslError>
 #include <QSslSocket>
 #include <QStringLiteral>
+#include <QOAuthUriSchemeReplyHandler>
+#include <QDesktopServices>
```
aji13187, line 27: Minor, Include alphabetic order can be fixed. Line 48 also.

uitools/common/src/ArcGISAuthenticationController.cpp
```diff
@@ -421,4 +435,54 @@ int ArcGISAuthenticationController::previousFailureCount_() const
...
+    oauthFlow->setAuthorizationCode(code);
+    emit oauthFlow->granted();
+  });
+
+  oauthFlow->setAuthorizationUrl(m_currentOAuthUserLoginPrompt->authorizeUrl());
+  oauthFlow->setClientIdentifier(m_currentOAuthUserConfiguration->clientId());
+
+  connect(oauthFlow, &QAbstractOAuth::authorizeWithBrowser, this, &QDesktopServices::openUrl);
+  connect(oauthFlow, &QAbstractOAuth::granted, this, [callbackReplyHandler, oauthFlow, this]()
+  {
```
aji13187, line 463: Is there an error case to handle if not granted ?

aji13187 (approved):
LGTM

uitools/common/src/ArcGISAuthenticationController.cpp
JamesMBallard: Good question. I can wire up some defensive logic, but I am not sure what cases are possible. I suppose these are all possible, but possibly others:
1. The browser opens up, but the user returns to the app without the redirect workflow (for example if they don't know the credentials and just gave up). I am not sure how to detect this.
2. The browser opens up, but the user decides to abandon the login and just cancel. **This is already covered** if we cannot find the "code" from a successful login.
3. `callbackReplyHandler->listen()` returns false - I don't know why this could fail, but it returns false so presumably it can. I will add logic for this.

As far as I know, as long as `oauthFlow->grant();` is called, and the completes the login process (successfully, or unsuccessfully), then `QAbstractOAuth::granted` will always emit.

jared-2016 (approved):
lgtm

aji13187 (approved):

PR #724, opened 2025-05-22
Bri12415/6084 authenticator

bri12415:


bri12415:
Tested workflows using our `QuickToolkitTester` app.

bri12415:
Some merge conflicts have showed up, will resolve.

uitools/common/src/AuthenticatorController.cpp
```diff
@@ -139,7 +139,7 @@ void ArcGISAuthenticationController::handleArcGISAuthenticationChallenge(ArcGISA
       {
         if (!m_currentArcGISChallenge)
         {
-          return;
+        return;
```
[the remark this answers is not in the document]
bri12415: Thanks

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/ServerTrustView.qml
JamesMBallard, line 1: All the other views were stripped of their "view" name. We should be consistent on that. If we go with "Authenticator" prefix everywhere else, then this one should be AuthenticatorServerTrust, but I do think it would make sense for all the subcomponents to keep their "view" names.

bri12415: I had removed the "view" names based on some conversation in the internal issue. Specifically:

> For the view components should we stick with the "View" moniker, as AuthenticatorView.qml?
> > I don't think we have to unless we think it is needed. 

My understanding of "components" was all these pieces that I've modified so far (and `ServerTrustView` which I had missed). I'll rename these back to using the "View" suffix.

PR #729, opened 2025-06-12
Run continuations on the object's thread.

JamesMBallard:
This can cause problems if we don't when there's a need to set parents across thread on new objects.

In the case of BasemapGallery, new items were being created on a non-UI thread, resulting in a crash:
```
QQmlEngine: Illegal attempt to connect to Esri::ArcGISRuntime::Toolkit::BasemapGalleryItem(0x600002c6e6f0) that is in a different thread than the QML engine QQmlApplicationEngine(0x16ddbeee8.
17:15:04: The process crashed.
```

These changes fix that crash.


jared-2016 (approved):
👍 good catch

PR #733, opened 2025-06-17
continueAndFailWithError is deprecated and replaced with continueWithError

aji13187:
Replaced the deprecated `continueAndFailWithError` with `continueWithError`

### DIFF
--- a/uitools/common/src/AuthenticatorController.cpp
+++ b/uitools/common/src/AuthenticatorController.cpp
@@ -279,8 +279,10 @@ void AuthenticatorController::continueWithUsernamePasswordArcGIS_(const QString&
       return;
     }
     emit previousFailureCountChanged();
-    m_currentArcGISChallenge->continueWithError(e.error());
-    m_currentArcGISChallenge.reset();
+    auto* arcgisChallenge = m_currentArcGISChallenge.release();
+    arcgisChallenge->setParent(this);
+    arcgisChallenge->deleteLater();
+    arcgisChallenge->continueWithError(e.error());
   });
 }
 
--- a/uitools/common/src/AuthenticatorController.cpp
+++ b/uitools/common/src/AuthenticatorController.cpp
@@ -143,7 +143,7 @@ void AuthenticatorController::handleArcGISAuthenticationChallenge(ArcGISAuthenti
           return;
         }
 
-        m_currentArcGISChallenge->continueAndFailWithError(e.error());
+        m_currentArcGISChallenge->continueWithError(e.error());
         m_currentArcGISChallenge.reset();
       });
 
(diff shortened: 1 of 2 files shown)


JamesMBallard (approved):

uitools/common/src/AuthenticatorController.cpp
```diff
@@ -135,16 +135,15 @@ void AuthenticatorController::handleArcGISAuthenticationChallenge(ArcGISAuthenti
...
-        m_currentArcGISChallenge.reset();
       }).onFailed(this, [this](const ErrorException& e)
       {
         if (!m_currentArcGISChallenge)
         {
           return;
         }
 
-        m_currentArcGISChallenge->continueAndFailWithError(e.error());
-        m_currentArcGISChallenge.reset();
```
aji13187, line 147: @JamesMBallard - Not sure about the `m_currentArcGISChallenge.reset();` do we need to retain here?
I think the reset was there to allow `continueAndFailWithError()` only once.

uitools/common/src/AuthenticatorController.cpp
```diff
@@ -269,35 +270,18 @@ void AuthenticatorController::continueWithUsernamePasswordArcGIS_(const QString&
...
-    {
-      m_arcGISPreviousFailureCountsForUrl.remove(requestUrl);
-      m_currentArcGISChallenge->continueAndFailWithError(e.error());
-      m_currentArcGISChallenge.reset();
       return;
     }
-
     emit previousFailureCountChanged();
+    m_currentArcGISChallenge->continueWithError(e.error());
     emit displayUsernamePasswordSignInView();
```
JamesMBallard, line 283: I don't know if this is correct anymore. Previously we kept track of when to display this dialog with one challenge, but now more challenges will be issued if this is needed. Now we should just call `continueWithError` and let additional challenges happen, or not happen as appropriate.

JamesMBallard (approved):

uitools/common/src/AuthenticatorController.cpp
```diff
@@ -143,7 +143,8 @@ void AuthenticatorController::handleArcGISAuthenticationChallenge(ArcGISAuthenti
           return;
         }
 
-        m_currentArcGISChallenge->continueAndFailWithError(e.error());
+        emit previousFailureCountChanged();
+        m_currentArcGISChallenge->continueWithError(e.error());
         m_currentArcGISChallenge.reset();
```
JamesMBallard, line 148: Does this code path need the same fix of `deleteLater`? I assume this will crash in the OAuth workflow without your fix.

JamesMBallard (approved):

PR #738, opened 2025-07-14
Add internal test helper.

JamesMBallard:
In some automated tests, we will get the server trust challenge. That is fine, but when we're not testing that we should be able to silently ignore it.

There are inherent risks with allowing any way to disable or bypassing challenges. I feel this is ok because:
1. The intent is clear - this is a private method that can only be accessed by a friend method that doesn't exist in this codebase.
2. This class is not meant to be used directly anyways, and we internally document it can change any time.
3. This repo is open source. We cannot prevent misuse (purposeful or otherwise) since anyone can clone the repo and change anything they want.

If someone does manage to accidentally invoke this test-only functionality, it will loudly log to the console this is for testing purposes only. We will see that in all our automated tests that use this.

### DIFF
--- a/uitools/common/src/AuthenticatorController.cpp
+++ b/uitools/common/src/AuthenticatorController.cpp
@@ -17,6 +17,7 @@
 #include "AuthenticatorController.h"
 
 // Qt headers
+#include <QDebug>
 #include <QDesktopServices>
 #include <QFuture>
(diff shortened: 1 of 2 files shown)


aji13187 (approved):

JamesMBallard:
The work for this has been postponed. Closing this out for now.

PR #745, opened 2025-07-18
Remove the Authentication View demo from UiToolsExamples

bri12415:
The `Authentication View` component is deprecated and part of the legacy authentication workflows. We are moving the `UiToolsExample` sample app to use core auth, so we are removing the `Authentication View` example from it.

This doesn't remove the deprecated component itself, just the example.


uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/AuthenticationView.qml
```diff
@@ -68,7 +68,6 @@ import QtQuick.Controls
         {Qt WebView}.
 
   \image authenticationview_qml.png authentication_view
-  \snippet qml/demos/AuthenticationViewDemoForm.qml Set up Authentication View QML
```
bri12415, line 71: This snippet has to get removed because it refers to the removed demo. Should we set up a replacement?

JamesMBallard: Yes, we should. Maybe we just copy that snippet into this location.

JamesMBallard (commented):
We should keep the snippet that was lost, even though it's a deprecated item.

JamesMBallard (approved):

PR #751, opened 2025-08-07
Jared/popup doc fixmain

jared-2016:
James brought it to my attention that the markdown doc was incorrect for popupview so I got this change up to fix it. Then decided we should also change over the example app to use it as well which was straight forward even though the data is not interesting.

I want to update the image in the markdown but the url points to `https://developers.arcgis.com/qt/toolkit/api-reference/images/popupview.png` and im not sure how to update this.

### DIFF
--- a/uitools/examples/qml/demos/PopupViewDemoForm.qml
+++ b/uitools/examples/qml/demos/PopupViewDemoForm.qml
@@ -21,30 +21,6 @@ import Esri.ArcGISRuntime.Toolkit
 import DemoApp
 
 DemoPage {
-    sceneViewContents: Component {
-        SceneView {
-            id: view
-
-            //! [Set up Popup View QML]
-            PopupView {
-                id:popupView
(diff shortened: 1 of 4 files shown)


JamesMBallard:
@dkemlage do you have any details on what @jared-2016 asked above?

JamesMBallard:
It may make more sense to check this into v.next first, then cherry pick it to main. We will want to verify the changes are good first. Going to main first we risk disrupting users using 200.8 if anything is wrong.

JamesMBallard (commented):
I recommend merging to v.next first, but all the changes look fine.

JamesMBallard (approved):
I verified the changes on v.next. This is good to merge to main.

jared-2016:
I would rather cherry pick from v.next to be safer since that was verified. https://github.com/Esri/arcgis-maps-sdk-toolkit-qt/pull/753

PR #755, opened 2025-08-19
Callout component fixes

min14751:
- Callout background follows system theme.
- Callout text defined by a variable as a workaround to update text color when switching themes on windows.


min14751:
The Callout itself doesnt expose a color property we just have variables for position size and the text/image to be displayed.

min14751:
but yes the existing colors can be overridden if needed and at which point the theme colors will also be changed.

PR #757, opened 2025-08-21
Hover effects disabled when components are disabled.

min14751:
- Calcite components have hover effects when component is disabled.
- Slider and its components do not have a visual indicator for the disabled state.

The calcite test project can be used to review these changes. We have a check box for disabling the components.

### DIFF
--- a/calcite/Calcite/CheckIndicator.qml
+++ b/calcite/Calcite/CheckIndicator.qml
@@ -16,15 +16,16 @@
 import QtQuick
 
 Rectangle {
-    property var control;
+    property var control
+    property bool isHoveredAndEnabled: control.hovered && control.enabled
     width: 24
     height: 24
     border {
         width: 2
-        color: control.checkState !== Qt.Unchecked ? control.hovered && control.enabled ? Calcite.brandHover : Calcite.brand :
-        control.hovered && control.enabled ? Calcite.text3 : Calcite.borderInput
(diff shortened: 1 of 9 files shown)


calcite/Calcite/CheckIndicator.qml
```diff
@@ -21,10 +21,10 @@ Rectangle {
     height: 24
     border {
         width: 2
-        color: control.checkState !== Qt.Unchecked ? control.hovered ? Calcite.brandHover : Calcite.brand :
-        control.hovered ? Calcite.text3 : Calcite.borderInput
+        color: control.checkState !== Qt.Unchecked ? control.hovered && control.enabled ? Calcite.brandHover : Calcite.brand :
```
jared-2016, line 24: Just putting my thoughts out there but this ternary logic is starting to get out of hand but I see this isn't new as I see them all other the place in this Component. Nothing to change unless you feel otherwise.

jared-2016: maybe you could combine `control.hovered && control.enabled` into a single property and use that in place since that one is repeated quite a bit? not sure how that works with property binding but just an idea to reduce the length of these.

jared-2016 (approved):
one comment on ternary length and an idea to make them shorter but not required.

bri12415 (approved):

PR #761, opened 2025-08-23
Remove PopupStackView documentations and CMakeLists.txt entry

bri12415:
When removing `PopupStackView` from the toolkit due to deprecation, I had missed some items related to it that should have been removed. Namely, documentation files and an entry in our CMakeLists.txt.


jared-2016 (approved):
Lgtm

PR #771, opened 2025-10-06
Calcite: progress bar implementation

min14751:
qml dir includes a progress bar but the component is not present.
This pr adds this.

### DIFF
--- a/calcite/Calcite/ProgressBar.qml
+++ b/calcite/Calcite/ProgressBar.qml
@@ -17,7 +17,7 @@ import QtQuick
 import QtQuick.Templates as T
 
 T.ProgressBar {
-    id: control2
+    id: control
 
     implicitWidth: Math.max(implicitBackgroundWidth + leftInset + rightInset,
                             implicitContentWidth + leftPadding + rightPadding)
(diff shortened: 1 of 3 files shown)


bri12415 (approved):
Would you mind including some screenshots in this PR of what this looks like? Would be good for record keeping in my opinion. Not necessary though.

min14751:
@bri12415 
Screenshot 2025-10-08 at 1 44 43 PM

Pretty standard progress bar that uses Calcite colors

bri12415:
@min14751 

Looks good, thanks.

calcite/Calcite/ProgressBar.qml
```diff
@@ -0,0 +1,46 @@
+/*******************************************************************************
+ *  Copyright 2012-2021 Esri
```
jared-2016, line 2: is this copyright year right?

jared-2016 (approved):
one question but looks good to me.

calcite/Calcite/ProgressBar.qml
min14751: Nope not updated. We could create a new issue to update it. But for now its consistent with the other components.

PR #785, opened 2025-11-17
Ldanzinger/overview fix

ldanzinger:
Fixes the overview map initially displaying zoomed out

Since the inset map creation/viewpoint setting is not happening in the constructor, we need to make sure to set it in the case that the map is already loaded and if it is not yet loaded.

### DIFF
--- a/uitools/common/src/OverviewMapController.cpp
+++ b/uitools/common/src/OverviewMapController.cpp
@@ -22,6 +22,7 @@
 #include "SingleShotConnection.h"
 
 // ArcGISRuntime headers
+#include <Error.h>
 #include <Geometry.h>
 #include <Graphic.h>
 #include <GraphicListModel.h>
@@ -32,6 +33,7 @@
 #include <MapViewTypes.h>
 #include <Point.h>
 #include <Polygon.h>
+#include <Scene.h>
 #include <SimpleFillSymbol.h>
 #include <SimpleLineSymbol.h>
 #include <SimpleMarkerSymbol.h>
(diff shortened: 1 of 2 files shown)


uitools/common/src/OverviewMapController.cpp
```diff
@@ -111,6 +111,42 @@ namespace Esri::ArcGISRuntime::Toolkit {
...
+
+      // if the map is already loaded, setup the inset map
+      if (sceneView->arcGISScene()->loadStatus() == LoadStatus::Loaded)
+      {
+        setupInsetMapForScene(sceneView);
+      }
+      // set up the inset map once the scene is done loading
+      else
+      {
+        connect(sceneView->arcGISScene(), &Scene::doneLoading, this, [setupInsetMapForScene, sceneView](Error e)
```
JamesMBallard: Double check this compiles, but you can avoid a hefty copy here.

```suggestion
        connect(sceneView->arcGISScene(), &Scene::doneLoading, this, [setupInsetMapForScene = std::move(setupInsetMapForScene), sceneView](Error e)
```

uitools/common/src/OverviewMapController.cpp
```diff
@@ -136,6 +172,44 @@ namespace Esri::ArcGISRuntime::Toolkit {
...
+
+      // if the map is already loaded, setup the inset map
+      if (mapView->map()->loadStatus() == LoadStatus::Loaded)
+      {
+        setupInsetMapForMap(mapView);
+      }
+      // set up the inset map once the map is done loading
+      else
+      {
+        connect(mapView->map(), &Map::doneLoading, this, [setupInsetMapForMap, mapView](Error e)
```
JamesMBallard:
```suggestion
        connect(mapView->map(), &Map::doneLoading, this, [setupInsetMapForMap = std::move(setupInsetMapForMap), mapView](Error e)
```

JamesMBallard: Likely this will rarely/never happen but I suppose it's theoretically possible that the map/scene loads after you check `if (mapView->map()->loadStatus() == LoadStatus::Loaded)`, but before you connect to doneLoading. No great solutions to that that wouldn't involve a lot more code.

I think we should prevent a possible data race if possible, but this could scope creep your PR. What do you think?

ldanzinger: would it make sense just to swap the order around? First make the connection (which may never emit), then check if it's already loaded after. That would guarantee the connection was always setup, but then catch the odd case that it already emitted before it was hooked up

JamesMBallard: yeah that should work, but then you have be careful not to move the lambda into the connection and try to use it after that point. You could make it a function.

aji13187 (approved):
lgtm

uitools/common/src/OverviewMapController.cpp
ldanzinger: changed to function and switched up the order @JamesMBallard

JamesMBallard (approved):

PR #787, opened 2025-11-19
Fix - qmldir warning on Android

aji13187:
`qmldir` helps QML engine to locate QML components / types.
We register all the types in [register.cpp](https://github.com/Esri/arcgis-maps-sdk-toolkit-qt/blob/main/uitools/toolkitcpp/src/Esri/ArcGISRuntime/Toolkit/register.cpp), so qmldir is not needed.

But for Android, androiddeployqt looks for `qmldir` for `import Esri.ArcGISRuntime` for some reason.
So added a minimal placeholder qmldir file to avoid error during deployment.

The `qmldir` for `import Esri.ArcGISRuntime.Toolkit` is already available [here](https://github.com/Esri/arcgis-maps-sdk-toolkit-qt/blob/main/uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/qmldir).

`:-1: error: qmldir file not found at "/qt/sdk/arcgis-maps-sdk-toolkit-qt/uitools/toolkitcpp/import/Esri/ArcGISRuntime"`


JamesMBallard:
This doesn't feel like the right solution, but maybe it is. Is it possible that our existing qmldir file is just in the wrong place? Should that one be one directory up where this new one is?

aji13187:
> Is it possible that our existing qmldir file is just in the wrong place?

The existing qmldir is at correct place.
The existing `qmldir` is for `import Esri.ArcGISRuntime.Toolkit`, so the qmldir path should be `Esri/ArcGISRuntime/Toolkit/qmldir` (`.` is replaced by` /` in the import statement)

The error is showing for `import Esri.ArcGISRuntime`, it is expecting a `qmldir` file under `Esri/ArcGISRuntime/qmldir'

JamesMBallard:
I think I understand now.

Of all of these, https://github.com/search?q=repo%3AEsri%2Farcgis-maps-sdk-toolkit-qt%20import%20Esri.ArcGISRuntime&type=code I think `import Esri.ArcGISRuntime` is an obsolete import statement that referred to our QML API. Do things work fine if you simply delete all those?

aji13187:
The current changes fixes the problem.
Looks like if the directory (import/Esri/Examples) doesn't exist, there is no error.
If I add an 'Examples' directory the error is showing.

Now the two imports in the qml files will be as below, Two different names,`Esri.Examples` and `Esri.ArcGISRuntime.Toolkit`
```
import Esri.Examples
import Esri.ArcGISRuntime.Toolkit
```
Is that fine? If we rename `import Esri.ArcGISRuntime.Toolkit` to `import Esri.Examples.Toolkit` the same error will come.

JamesMBallard:
>Is that fine?

Yes. I don't see that rename happening. Good investigation figuring out the root cause.

JamesMBallard (approved):

jared-2016 (approved):

PR #790, opened 2025-12-08
Calcite - Icon Integration.

min14751:
This PR contains the following changes:

- Added calcite-design-system repo as a submodule for the icons.
- Updated Readme files
- Created calcite-icons.qrc which references icons from the submodule. This will be included in the calcite_test build only if the submodule folder is found. 
- Updated calcite_test project to display/demo calcite icon usage. 
    - main.qml: added dialog and button to open icons showcase menu
    - IconShowcase.qml: newly added qml file that displays some icons.

### DIFF
--- a/calcite/demo/qml/main.qml
+++ b/calcite/demo/qml/main.qml
@@ -37,6 +37,13 @@ ApplicationWindow {
         //Calcite.theme = Calcite.Dark;
     }
 
+    // Check if Calcite icons submodule is initialized
+    property bool calciteIconsAvailable: {
+        // Check if icon resources are available
+        var testPath = "qrc:/calcite/icons/check-24.svg"
+        return Qt.resolvedUrl(testPath).toString().length > 0
+    }
+
     // add a mapView component
     MapView {
         id: mapView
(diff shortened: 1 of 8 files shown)


calcite/demo/qml/main.qml
```diff
@@ -32,9 +32,11 @@ ApplicationWindow {
...
     }
 
-    Component.onCompleted: {
-
-        //Calcite.theme = Calcite.Dark;
+    // Check if Calcite icons submodule is initialized
+    property bool calciteIconsAvailable: {
+        // Check if icon resources are available
+        var testPath = "qrc:/calcite/icons/check-24.svg"
+        return Qt.resolvedUrl(testPath).toString().length > 0
```
JamesMBallard, line 39: Since this is a resource, can this ever be false?

JamesMBallard (approved):
Just one question, but I don't think it requires another review.

We should have a CMake example, so can you please log an issue to add one. You don't need to add it with this PR.

bri12415 (approved):

calcite/demo/qml/main.qml
min14751: Yes, when the submodule is not available calcite-icons.qrc wont be included so this will fail

PR #792, opened 2025-12-15
Calcite Demo - Scale fixes for mobile, Hide open folder button for andriod/ios.

min14751:
This Pr has the following changes:
- `main.qml ` - aded `scale` property to frame and panes. Added top and bottom inset for android. Added black tinted bg when dialogue is open.
- ` IconShowcase.qml` - added `canOpenLocalFolders` to track platform.  Open Folder button is visible/hidden based on this value. Fixed open folder functionality for windows.
- `calcite_test.pro` - Removed multimedia for iOS. This was causing builds to fail for iOS.

Note:- majority of the changes as shown in the diff are due to indentation. the above mentioned changes are the only additions.

### DIFF
--- a/calcite/demo/calcite_test.pro
+++ b/calcite/demo/calcite_test.pro
@@ -82,6 +82,8 @@ macx {
 
 ios {
     include (iOS/iOS.pri)
+    QT -= multimedia
+    DEFINES += QT_NO_MULTIMEDIA_WIDGETS
 }
 
 android {
--- a/calcite/demo/calcite_test.pro
+++ b/calcite/demo/calcite_test.pro
@@ -82,8 +82,33 @@ macx {
 
 ios {
     include (iOS/iOS.pri)
-    QT -= multimedia
-    DEFINES += QT_NO_MULTIMEDIA_WIDGETS
+    PLATFORM = "iOS"
+
+    # workaround for https://bugreports.qt.io/browse/QTBUG-129651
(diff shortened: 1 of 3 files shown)


calcite/demo/calcite_test.pro
```diff
@@ -82,6 +82,8 @@ macx {
 
 ios {
     include (iOS/iOS.pri)
+    QT -= multimedia
+    DEFINES += QT_NO_MULTIMEDIA_WIDGETS
 }
```
min14751, line 112: Not sure why iOS doesn't build for me without this.

bri12415: This is a known issue with the ffmpeg binaries and how Qt used to ship them with their framework vs how they do now. I can't link the issue in our repo here.

I don't think we should keep this QT -= multimedia and the like here because I think we made an announcement on this, but I could be wrong on both accounts. @JamesMBallard?

JamesMBallard: Brian's right we don't want to keep perpetuating `QT -= multimedia`.

I think we can add this, including the qt bug report link 
https://github.com/Esri/arcgis-maps-sdk-samples-qt/blob/main/SampleViewer/SampleViewer.pro#L121-L149

That way we'll know that all these workarounds can be removed when we move up our minimum build version of Qt.

har13205 (approved):

bri12415 (approved):

PR #796, opened 2025-12-23
running clang format against entire repo

ldanzinger:
@GuillaumeBelzPro @JamesMBallard could you please review? I modified the clang format script to run against the entire repo:

`stagedFiles=$(find . -type f \( -name '*.cpp' -o -name '*.h' -o -name '*.mm' \))`

I then ran the precommit hooks against the entire repo:

`bash .git/hooks/pre-commit --all-files`

I spot checked the result, but could use additional eyes to make sure it's what we expect


uitools/common/src/ArcGISAuthenticationChallengeRelay.h
```diff
@@ -16,26 +16,26 @@
 #ifndef ESRI_ARCGISRUNTIME_TOOLKIT_ARCGISAUTHENTICATIONCHALLENGERELAY_H
 #define ESRI_ARCGISRUNTIME_TOOLKIT_ARCGISAUTHENTICATIONCHALLENGERELAY_H
 
-// ArcGIS Maps SDK headers
+// STL headers
```
JamesMBallard, line 19: This isn't right, but I think this is a separate problem with our code formatter. I head to tweak a bunch of these by hand and disable the hooks to get them right.

I don't think it's worth abandoning the entire PR for this.

uitools/common/src/LocatorSearchSource.h
```diff
@@ -16,19 +16,19 @@
...
 
-// Toolkit headers
-#include "SearchSourceInterface.h"
+// Qt headers
+#include <QFuture>
+#include <QList>
+#include <QObject>
 
-// ArcGISRuntime headers
+// STL headers
```
JamesMBallard, line 24: These are wrong, too. Just comments though, so we can do another sweep or you can find/fix all the offenders in this PR (easier said than done).

uitools/common/src/NetworkAuthenticationChallengeRelay.h
```diff
@@ -16,26 +16,26 @@
 #ifndef ESRI_ARCGISRUNTIME_TOOLKIT_NETWORKAUTHENTICATIONCHALLENGERELAY_H
 #define ESRI_ARCGISRUNTIME_TOOLKIT_NETWORKAUTHENTICATIONCHALLENGERELAY_H
 
-// ArcGIS Maps SDK headers
+// STL headers
```
JamesMBallard, line 19: Another one. I won't comment on these after this one. I assume there will be a handful scattered around.

JamesMBallard (approved):
A few include section comments got switched to the wrong thing, but there's an outstanding problem with these.

uitools/common/src/ArcGISAuthenticationChallengeRelay.h
GuillaumeBelzPro: The formater use `#include "..."` for ArcGIS headers and `#include <...>` for STL headers. It's not a bug, it's a feature :) It must be updated to support the style used in samples and toolkit.

uitools/common/src/AttachmentsPopupElementViewController.cpp
```diff
@@ -1,3 +1,4 @@
+
```
GuillaumeBelzPro, line 1: I think this empty line is added because the formater doesn't recognize the copyright used in toolkit as a copyright block. If I remember correctly, the formater expects the copyright lines start with `\\`.

uitools/common/src/AttachmentsPopupElementViewController.cpp
```diff
@@ -31,41 +32,41 @@
...
-  This class is an internal implementation detail and is subject to change.
- */
-AttachmentsPopupElementViewController::AttachmentsPopupElementViewController(
-    AttachmentsPopupElement* attachmentsPopupElement, PopupViewController* popupViewController, QObject* parent)
-  : PopupElementViewItem{attachmentsPopupElement, parent},
+  /*!
+    \internal
+    This class is an internal implementation detail and is subject to change.
+   */
+  AttachmentsPopupElementViewController::AttachmentsPopupElementViewController(
```
GuillaumeBelzPro: I was surprised by the indentation in this file, and I realized there is a `.clang-format` setting file in toolkit, which is different that what we use for our code. Is this intentional?

GuillaumeBelzPro (changes requested):
Expect if you think it's better to have a different `.clang-format` setting file for toolkit, I recommend to use the same settings we use for our code.

GuillaumeBelzPro:
There is no `.clang-format` setting file in samples. Maybe we should standardize that and have a setting file in both repos. Or in neither of the two.

JamesMBallard:
I think we should have a .clang-format in both repos and adapt our current internal one so things are consistent.

ldanzinger:
@GuillaumeBelzPro updated to use new clang-format file, and re-ran on the repo

GuillaumeBelzPro (approved):
Looks good

PR #797, opened 2025-12-27
Fixed Keyboard Nav for Combobox

min14751:
Removed conditional check for `model` in `ListView`  `control.popup.visible ? control.delegateModel : null` which was preventing keyboard navigation.
Border for the `ListView` is set to `1` regardless of the current theme so we get a border for dark theme as well. This was causing the `ListView` to blend with the background menu.

### DIFF
--- a/calcite/Calcite/ComboBox.qml
+++ b/calcite/Calcite/ComboBox.qml
@@ -26,6 +26,27 @@ T.ComboBox {
                              Math.max(contentItem.implicitHeight,
                                       indicator ? indicator.implicitHeight : 0) + topPadding + bottomPadding)
 
+    property bool _justOpened: false
+
+    Keys.onPressed: (event) => {
+                        if (event.key === Qt.Key_Return || event.key === Qt.Key_Enter) {
+                            if (!control.popup.visible) {
+                                control.popup.open();
+                                _justOpened = true;
+                                event.accepted = true;
+                            }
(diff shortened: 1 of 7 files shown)


calcite/Calcite/ComboBox.qml
```diff
@@ -26,6 +26,27 @@ T.ComboBox {
                              Math.max(contentItem.implicitHeight,
                                       indicator ? indicator.implicitHeight : 0) + topPadding + bottomPadding)
 
+    property bool _justOpened: false
+
```
min14751: Private property to keep the combobox list view open after pressing Enter/Return

calcite/Calcite/RadioButton.qml
```diff
@@ -25,6 +25,25 @@ T.RadioButton {
...
+
+    Keys.onReleased: (event) => {
+                         control.down = false;
+                         event.accepted = true;
+                     }
+
+    onActiveFocusChanged: {
+        if (!activeFocus) {
+            control.down = false;
+        }
```
min14751: Ensures control.down is set to false when `Keys.onReleased` fails to trigger. This happens when we open another UI and focus shits to that preventing `Keys.onReleased` from triggering here.

jared-2016 (approved):

aji13187 (approved):

PR #798, opened 2026-01-06
Callout changes and add sceneLocationVisibility enum to QmlEnums.

min14751:
This pr adds callout support for Scene view and Local scene view.

- Added enum for sceneLocationVisibility
- Opacity support for callout in SV and LSV.

### DIFF
--- a/uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/Callout.qml
+++ b/uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/Callout.qml
@@ -75,12 +75,12 @@ Pane {
     id: root
 
     opacity: {
-        if (!calloutData || calloutData.sceneLocationVisibility === undefined)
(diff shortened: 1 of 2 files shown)


uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/Callout.qml
```diff
@@ -73,6 +74,15 @@ import QtQuick.Shapes
 Pane {
     id: root
 
+    opacity: {
+        if (!calloutData || calloutData.sceneLocationVisibility === undefined)
```
GuillaumeBelzPro: I don't remember the guidelines in QML, if we need to use brackets in one line statement, but we do that for C++ code. I guess it's the same for QML.

GuillaumeBelzPro (approved):

JamesMBallard (approved):

PR #799, opened 2026-01-07
Identity-aware proxy

JamesMBallard:
This introduces new functionality for identity-aware proxy support, and a minor refactoring of the existing iOS OAuth logic to accommodate some quirks of the IAP workflows.

There are two main quirks of IAP that necessitated migrating to a native iOS implementation.
1. redirect URL matching.
2. IAP logout workflow shortcomings.

For the first issue, there are some issues discovered in testing IAP where Qt's auth APIs reject the redirect URL because the server does not respond with the exact match of what was requested. The new iOS implementation allows resiliency against that since it does not validate it.

For the second issue, iOS provides no "back button" navigation experience like Android does if you are sent over to Safari with no clean way back. The new iOS implementation shows Safari as a popup from within the same app session while still being out of process. This allows users to get back to the app cleanly.

## Summary of changes

 * New iOS OAuth backend to handle the browser experience.
 * New OAuth logout callbacks hooked up and enabled. This allows for signing out of IAP workflows.
 * Embedded browser logout is not implemented - if this happens the toolkit will immediately fail the request.
 * Both Android and iOS implementations of logout are basic. As mentioned, the callback mechanism does not work as expected (it cannot redirect to an app), so the logic assumes it is successful. That matches other SDK team's implementations.

### DIFF
--- a/uitools/common/src/AuthenticatorController.cpp
+++ b/uitools/common/src/AuthenticatorController.cpp
@@ -26,6 +26,7 @@
 #include <QStringLiteral>
 #include <QUrl>
 #include <QtGlobal>
+#include <QGuiApplication>
 
 // Maps SDK headers
 #include <ArcGISRuntimeEnvironment.h>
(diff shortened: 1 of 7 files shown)


JamesMBallard:
@ldanzinger I want to tag you so you are aware of the iOS changes I've made here. It will change the existing OAuth sign in experience we've already tested, and require retesting. Due to the shortcomings mentioned above, I think it's a superior experience for IAP, and perhaps a more modern approach for OAuth. It only affects the out of process OAuth experience.

uitools/common/src/AuthenticatorController.cpp
```diff
@@ -57,6 +64,15 @@ using Esri::ArcGISRuntime::Authentication::AuthenticationManager;
 namespace Esri::ArcGISRuntime::Toolkit
 {
 
+  namespace
+  {
+    bool isInProcessOAuthRedirect(const QString& redirectUri)
+    {
+      return redirectUri == QStringLiteral("urn:ietf:wg:oauth:2.0:oob") || // this is the default value for "oob"
+             redirectUri.contains(QStringLiteral("oob")); // this is what the Qt docs indicate to check for
```
min14751: isn't this too broad? Will we ever reach a case where a regular uri might pass this check unintentionally?

JamesMBallard: I was going by this doc - https://doc.qt.io/qt-6/qabstractoauthreplyhandler.html#callback

but perhaps that is too specific and could catch the wrong thing. We are using `QOAuthUriSchemeReplyHandler` and it doesn't mention that, so I can tighten the check for exactly "urn:ietf:wg:oauth:2.0:oob". The secondary check was purely a defensive check, but as far as I know we've never hit it before. All our 'oob' testing and services use the longer uri format.

I will tighten it up.

min14751 (approved):

har13205 (approved):

uitools/common/src/IOS/IOSWebAuthenticationSession.mm
```diff
@@ -0,0 +1,195 @@
...
+      {
+        return windowScene.windows.firstObject;
+      }
+    }
+
+    return nil;
+  }
+} // namespace
+
+@interface QRTWebAuthPresentationContextProvider : NSObject<ASWebAuthenticationPresentationContextProviding>
```
GuillaumeBelzPro: Why "QRT"?

GuillaumeBelzPro (approved):
Looks good, but I'm not familiar enough with Apple Web API.

uitools/common/src/IOS/IOSWebAuthenticationSession.mm
JamesMBallard: The robot chose that name. I didn't nitpick the names, but "QRT" is misleading. I will change it to simply `QtWebAuthPresentationContextProvider`

PR #802, opened 2026-01-26
Standardize OAuth system browser approach

JamesMBallard:
Don't validate the incoming info, just pass it through.

Previously we did two steps:
1. `QOAuthUriSchemeReplyHandler::callbackReceived`, which only provides the parsed arguments from the browser session.
2. `QAbstractOAuth::granted`, where we need to reassemble the url with the code we parsed in the previous step.

We can simplify this by just using `QOAuthUriSchemeReplyHandler::callbackDataReceived` which has the info we're looking for to begin with.

This brings the Android implementation in line with the newer iOS implementation which just passes whatever the browser returned directly.


min14751 (approved):

jared-2016 (approved):
👍

PR #803, opened 2026-02-21
Popup edit summary.

min14751:
This PR adds editSummary to the Popup.

### DIFF
--- a/uitools/common/src/PopupViewController.cpp
+++ b/uitools/common/src/PopupViewController.cpp
@@ -119,11 +119,17 @@ namespace Esri::ArcGISRuntime::Toolkit
 
     emit popupChanged();
     emit titleChanged();
+    emit editSummaryChanged();
   }
 
   QString PopupViewController::title() const
   {
     return m_popup ? m_popup->title() : QString();
+  }
+
+  QString PopupViewController::editSummary() const
+  {
+    return m_popup ? m_popup->editSummary() : QString("");
   }
 
 } // namespace Esri::ArcGISRuntime::Toolkit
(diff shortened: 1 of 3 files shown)


uitools/common/src/PopupViewController.cpp
```diff
@@ -119,11 +119,17 @@ namespace Esri::ArcGISRuntime::Toolkit
...
   }
 
   QString PopupViewController::title() const
   {
     return m_popup ? m_popup->title() : QString();
   }
 
+  QString PopupViewController::editSummary() const
+  {
+    return m_popup ? m_popup->editSummary() : QString("");
```
JamesMBallard:
```suggestion
    return m_popup ? m_popup->editSummary() : QString{};
```

JamesMBallard (approved):
One small thing.

uitools/common/src/PopupViewController.cpp
min14751: Wont this return a null value rather than an empty string?

Then in qml i will have to null check [here](https://github.com/Esri/arcgis-maps-sdk-toolkit-qt/pull/803/changes#diff-3e36caa9ad64c8801b1eb4180e1e8e12c3ae9bbe89e17ac9c9e5f7d7192ad26aR300) `text:controller.editSummary`

JamesMBallard: I don't think so. Double check. QString{} should just be a slightly optimized way to indicate an empty string. QML should treat it the same.

min14751: Yes you are right `QString{}` is an empty string.

jared-2016 (approved):

PR #805, opened 2026-03-05
Update popup when AttributeListModel changes.

min14751:
The following changes have been made in this PR:

- Added `refreshPopupContent_` private function to update popup.
- When Popup is set in the controller we additionally listen for changes to the `AttributeListModel` to update popup.

Instead of recreating controllers to update view on every refresh, the order of the views are checked to see if we can update them instead. Updating the individual views instead of recreating them prevents the scrollview from being reset.

### DIFF
--- a/uitools/common/src/PopupViewController.cpp
+++ b/uitools/common/src/PopupViewController.cpp
@@ -46,6 +46,80 @@
 namespace Esri::ArcGISRuntime::Toolkit
 {
 
+  namespace
+  {
+    bool isReusableElementType(PopupElementType type)
+    {
+      return type == PopupElementType::TextPopupElement || type == PopupElementType::FieldsPopupElement;
+    }
+
+    PopupElementViewItem* popupElementViewItemAt(GenericListModel* model, int index)
+    {
+      auto* object = model->element(model->index(index, 0));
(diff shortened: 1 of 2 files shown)


uitools/common/src/PopupViewController.cpp
```diff
@@ -45,6 +46,80 @@
...
+      switch (type)
+      {
+        case PopupElementType::TextPopupElement:
+          QMetaObject::invokeMethod(object, "textPopupElementChanged", Qt::DirectConnection);
+          break;
+        case PopupElementType::FieldsPopupElement:
+          QMetaObject::invokeMethod(object, "fieldsPopupElementChanged", Qt::DirectConnection);
+          break;
+        default:
+          break;
```
jared-2016, line 97: maybe Q_UNIMPLEMENTED(); here?

jared-2016 (approved):
one small suggestion but looks good. Nice work.

aji13187 (approved):

uitools/common/src/PopupViewController.cpp
min14751: Added `Q_UNIMPLEMENTED();`

PR #806, opened 2026-03-13
Cancel outstanding challenges.

JamesMBallard:
Ideally the controller can detect this automatically, but there are too many edge cases to be confident
we can do it correctly and not misinterpret what actually happened.

For now, we will allow consuming code to call this when appropriate. I did not add any logic for the OAuth login/logout prompts, because abandoning those workflows would be more complicated to detect since they can be out of process.


JamesMBallard:
Requires https://github.com/Esri/arcgis-maps-sdk-samples-qt/pull/1958

jared-2016 (approved):

GuillaumeBelzPro (approved):

PR #807, opened 2026-03-13
Add support for refresh intervals for PopupMedia Images(PopupElements)

min14751:
This PR adds refresh interval support for PopupMedia Images. A fallback has also been added when an image fails to load.

Note: - In `MediaPopupElementView.qml`, only the `imageComp` control has been modified to add a timer and an image fallback. The other diffs are from `cmd+i` indent.

### DIFF
--- a/uitools/common/src/ImagePopupMediaItem.cpp
+++ b/uitools/common/src/ImagePopupMediaItem.cpp
@@ -1,4 +1,3 @@
-
 /*******************************************************************************
  *  Copyright 2012-2025 Esri
  *
@@ -42,7 +41,10 @@ namespace Esri::ArcGISRuntime::Toolkit
     connect(this, &ImagePopupMediaItem::imagePopupMediaItemChanged, this, &ImagePopupMediaItem::setupRefreshTimer);
   }
 
-  ImagePopupMediaItem::~ImagePopupMediaItem() = default;
+  ImagePopupMediaItem::~ImagePopupMediaItem()
+  {
+    m_refreshTimer.stop();
+  }
 
   QUrl ImagePopupMediaItem::sourceUrl() const
   {
--- a/uitools/common/src/ImagePopupMediaItem.cpp
+++ b/uitools/common/src/ImagePopupMediaItem.cpp
@@ -50,4 +50,9 @@ namespace Esri::ArcGISRuntime::Toolkit
     return popupMediaItem()->value()->linkUrl();
   }
 
+  quint64 ImagePopupMediaItem::imageRefreshInterval() const
+  {
+    return popupMediaItem()->imageRefreshInterval();
+  }
+
 } // namespace Esri::ArcGISRuntime::Toolkit
--- a/uitools/common/src/ImagePopupMediaItem.cpp
+++ b/uitools/common/src/ImagePopupMediaItem.cpp
@@ -36,6 +36,10 @@ namespace Esri::ArcGISRuntime::Toolkit
   {
     // bubble up imageClicked signal to PopupViewController. This is the sourceUrl & linkUrl used for ImagePopupMediaItems.
     connect(this, &ImagePopupMediaItem::imageClicked, popupViewController, &PopupViewController::imageClicked);
+
+    setupRefreshTimer();
+
+    connect(this, &ImagePopupMediaItem::imagePopupMediaItemChanged, this, &ImagePopupMediaItem::setupRefreshTimer);
   }
 
   ImagePopupMediaItem::~ImagePopupMediaItem() = default;
@@ -53,6 +57,20 @@ namespace Esri::ArcGISRuntime::Toolkit
   quint64 ImagePopupMediaItem::imageRefreshInterval() const
   {
     return popupMediaItem()->imageRefreshInterval();
+  }
+
+  void ImagePopupMediaItem::setupRefreshTimer()
+  {
+    m_refreshTimer.stop();
+
+    const quint64 interval = imageRefreshInterval();
+    if (interval > 0)
+    {
+      m_refreshTimer.setInterval(static_cast<int>(interval));
+      m_refreshTimer.setSingleShot(false);
+      connect(&m_refreshTimer, &QTimer::timeout, this, &ImagePopupMediaItem::refreshImage, Qt::UniqueConnection);
+      m_refreshTimer.start();
+    }
   }
 
 } // namespace Esri::ArcGISRuntime::Toolkit
--- a/uitools/common/src/ImagePopupMediaItem.cpp
+++ b/uitools/common/src/ImagePopupMediaItem.cpp
@@ -43,7 +43,10 @@ namespace Esri::ArcGISRuntime::Toolkit
 
   ImagePopupMediaItem::~ImagePopupMediaItem()
   {
-    m_refreshTimer.stop();
+    if (m_refreshTimer.isActive())
+    {
+      m_refreshTimer.stop();
+    }
   }
 
   QUrl ImagePopupMediaItem::sourceUrl() const
--- a/uitools/common/src/ImagePopupMediaItem.h
+++ b/uitools/common/src/ImagePopupMediaItem.h
@@ -19,6 +19,7 @@
 // Qt headers
 #include <QJsonObject>
 #include <QObject>
+#include <QTimer>
 
 // Other headers
 #include "PopupMediaItem.h"
@@ -44,9 +45,13 @@ namespace Esri::ArcGISRuntime::Toolkit
     QUrl sourceUrl() const;
     quint64 imageRefreshInterval() const;
 
+    void setupRefreshTimer();
+    QTimer m_refreshTimer;
+
   signals:
     void imagePopupMediaItemChanged();
     void imageClicked(const QUrl& sourceUrl, const QUrl& linkUrl);
+    void refreshImage();
   };
 
 } // namespace Esri::ArcGISRuntime::Toolkit
--- a/uitools/common/src/ImagePopupMediaItem.h
+++ b/uitools/common/src/ImagePopupMediaItem.h
@@ -33,6 +33,7 @@ namespace Esri::ArcGISRuntime::Toolkit
     Q_OBJECT
     Q_PROPERTY(QUrl sourceUrl READ sourceUrl NOTIFY imagePopupMediaItemChanged)
     Q_PROPERTY(QUrl linkUrl READ linkUrl NOTIFY imagePopupMediaItemChanged)
+    Q_PROPERTY(quint64 imageRefreshInterval READ imageRefreshInterval NOTIFY imagePopupMediaItemChanged)
 
   public:
     explicit ImagePopupMediaItem(PopupMedia* popupMedia, PopupViewController* popupViewController, QObject* parent = nullptr);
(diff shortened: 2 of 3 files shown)


min14751:
Fall back when the image fails to load:

Dark mode:
image

Light mode:
image

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/MediaPopupElementView.qml
```diff
@@ -206,35 +206,86 @@ ColumnLayout {
...
-                    source: listModelData.sourceUrl
+                    id: popupImage
+                    property url currentSourceUrl: listModelData.sourceUrl
+                    property bool hasLoggedImageError: false
+                    source: currentSourceUrl
                     fillMode: fullScreenImageDialog.visible ? Image.PreserveAspectFit : Image.PreserveAspectCrop
                     asynchronous: true
-                    cache: true
+                    // Disable cache when refresh interval is set so timed reloads fetch updated content.
+                    cache: listModelData.imageRefreshInterval === 0
```
JamesMBallard, line 236: The doc for the `cache` property isn't great IMHO.

If it's cached, I would assume it should still refresh if we change it. Can we instead just emit some changed signal when it's been refreshed?

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/MediaPopupElementView.qml
```diff
@@ -206,35 +206,86 @@ ColumnLayout {
...
+                    // Image fallback
+                    Rectangle {
+                        anchors.fill: parent
+                        visible: popupImage.status === Image.Error
+                        color: palette.base
+                        border.color: palette.dark
+                        border.width: 1
+
+
+
```
JamesMBallard:
```suggestion
                        border.width: 1

```

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/MediaPopupElementView.qml
```diff
@@ -206,35 +206,86 @@ ColumnLayout {
...
+                    Timer {
+                        running: listModelData.imageRefreshInterval > 0
+                        repeat: true
+                        interval: listModelData.imageRefreshInterval
+                        onTriggered: {
+                            popupImage.currentSourceUrl = "";
+                            popupImage.currentSourceUrl = listModelData.sourceUrl;
+                        }
+                    }
```
JamesMBallard: Is it possible to run the timer on the C++ side of things and then just notify the QML side when it needs to be refreshed? That way I'd think `listModelData.sourceUrl` can just emit that it changed and the QML side just picks up the changes.

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/MediaPopupElementView.qml
```diff
@@ -206,35 +206,86 @@ ColumnLayout {
...
                 Image {
-                    source: listModelData.sourceUrl
+                    id: popupImage
+                    property url currentSourceUrl: listModelData.sourceUrl
+                    property bool hasLoggedImageError: false
+                    source: currentSourceUrl
                     fillMode: fullScreenImageDialog.visible ? Image.PreserveAspectFit : Image.PreserveAspectCrop
                     asynchronous: true
-                    cache: true
+                    // Disable cache when refresh interval is set so timed reloads fetch updated content.
```
tanneryould, line 235: So if `cache` is `true` then whenever we fetch updated content, it would just pull from the cache?
I'm confused what caching would do if the interval was 0? Are we calling `fetch updated content` outside of the refresh interval?

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/MediaPopupElementView.qml
```diff
@@ -206,35 +206,86 @@ ColumnLayout {
...
+                        }
+                    }
+
+                    Timer {
+                        running: listModelData.imageRefreshInterval > 0
+                        repeat: true
+                        interval: listModelData.imageRefreshInterval
+                        onTriggered: {
+                            popupImage.currentSourceUrl = "";
+                            popupImage.currentSourceUrl = listModelData.sourceUrl;
```
tanneryould: `popupImage.currentSourceUrl` is only ever set to `listModelData.sourceUrl` -- what does unsetting then resetting it do?

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/MediaPopupElementView.qml
```diff
@@ -206,35 +206,86 @@ ColumnLayout {
...
             }
 
             Component {
                 id: imageComp
                 Image {
-                    source: listModelData.sourceUrl
+                    id: popupImage
+                    property url currentSourceUrl: listModelData.sourceUrl
+                    property bool hasLoggedImageError: false
+                    source: currentSourceUrl
```
tanneryould: Is there a reason we define `currentSourceUrl` and don't use `popupImage.source` directly?

uitools/common/src/ImagePopupMediaItem.h
```diff
@@ -41,6 +42,7 @@ namespace Esri::ArcGISRuntime::Toolkit
   private:
     QUrl linkUrl() const;
     QUrl sourceUrl() const;
+    quint64 imageRefreshInterval() const;
 
   signals:
     void imagePopupMediaItemChanged();
```
tanneryould, line 52: I ran a search and `imagePopupMediaItemChanged` is defined but never called anywhere in the repo -- is that intended?

tanneryould (commented):
I'm not sure I have enough context about this feature to do a proper review, but formatting-wise things look fine. I have a lot of questions though.

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/MediaPopupElementView.qml
min14751: That would be the case for when the image name/source changes(not tested). For sources such as this :
https://video.dot.state.mn.us/video/image/metro/C30279
If we refresh it does not update when it is cached

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/MediaPopupElementView.qml
min14751: it reloads the image. if we just have 
```
onTriggered: {
                            popupImage.currentSourceUrl = listModelData.sourceUrl;
                        }
```

There is no change in the image source and it does not update.

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/MediaPopupElementView.qml
min14751: if interval is 0 refresh is disabled and cache is enabled.

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/MediaPopupElementView.qml
JamesMBallard: And if we refresh it via emitting a signal indicating that the source has changed, that still doesn't work if `cached: true`?

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/MediaPopupElementView.qml
min14751: Even with cache false it does not reload the value if the `listModelData.sourceUrl` doesnt change. Take this for example: https://video.dot.state.mn.us/video/image/metro/C30279 . the Url does not change but the image updates when you refresh it.

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/MediaPopupElementView.qml
min14751: > And if we refresh it via emitting a signal indicating that the source has changed, that still doesn't work if `cached: true`?

No the image does not update.

uitools/common/src/ImagePopupMediaItem.h
min14751: Not sure on this.

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/MediaPopupElementView.qml
min14751: Just for convenience. Removed it to use `popupImage.source` directly.

uitools/toolkitcpp/import/Esri/ArcGISRuntime/Toolkit/MediaPopupElementView.qml
min14751: moved timer to c++

uitools/common/src/ImagePopupMediaItem.cpp
```diff
@@ -50,4 +54,23 @@ namespace Esri::ArcGISRuntime::Toolkit
...
+  {
+    m_refreshTimer.stop();
+
+    const quint64 interval = imageRefreshInterval();
+    if (interval > 0)
+    {
+      m_refreshTimer.setInterval(static_cast<int>(interval));
+      m_refreshTimer.setSingleShot(false);
+      connect(&m_refreshTimer, &QTimer::timeout, this, &ImagePopupMediaItem::refreshImage, Qt::UniqueConnection);
+      m_refreshTimer.start();
```
JamesMBallard, line 77: In the class's destructor should you stop the timer if it's still running?

JamesMBallard (commented):
One last question about if/when to stop the timer if it's running.

uitools/common/src/ImagePopupMediaItem.cpp
min14751: Since its a member of `ImagePopupMediaItem` parent child relationship should take care of stoping/destroying the timer on destruct(not 100% sure on this). Added stop on destruct just to be sure.

uitools/common/src/ImagePopupMediaItem.h
min14751: Going forward it gives us the option to update images in the popup without reconstructing the object. So it does not hurt to have it.

uitools/common/src/ImagePopupMediaItem.cpp
```diff
@@ -42,7 +41,10 @@ namespace Esri::ArcGISRuntime::Toolkit
     connect(this, &ImagePopupMediaItem::imagePopupMediaItemChanged, this, &ImagePopupMediaItem::setupRefreshTimer);
   }
 
-  ImagePopupMediaItem::~ImagePopupMediaItem() = default;
+  ImagePopupMediaItem::~ImagePopupMediaItem()
+  {
+    m_refreshTimer.stop();
```
JamesMBallard: Does it matter if you stop a timer that isn't running? I don't think so, but I don't want our users to see any console warnings related to it.

JamesMBallard (approved):

tanneryould (approved):
LGTM thank you for your explanations!

PR #808, opened 2026-03-16
Local Scene Support for Toolkit Components

tanneryould:
Adds support for LocalSceneView in Toolkit components

I just did a search for "SceneView" and "SceneViewToolkit" and added LocalSceneView support any time "SceneView" was mentioned.

There was a case in BasemapGalleryController where I also had to add a check for the Global SceneViewingMode, because LocalSceneView supports other Spatial References, like MapView.


uitools/common/src/OverviewMapController.h
```diff
@@ -62,14 +62,17 @@ namespace Esri::ArcGISRuntime::Toolkit
   private:
     void applyInsetNavigationToMapView(MapViewToolkit* view);
     void applyInsetNavigationToSceneView(SceneViewToolkit* view);
+    void applyInsetNavigationToLocalSceneView(LocalSceneViewToolkit* view);
 
     void applyMapNavigationToInset(MapViewToolkit* view);
     void applySceneNavigationToInset(SceneViewToolkit* view);
+    void applySceneNavigationToInset(LocalSceneViewToolkit* view);
```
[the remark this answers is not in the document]
tanneryould: LocalSceneView and SceneView are distinct classes of GeoView whereas the other method names specify the Scene GeoModel, which both LocalSceneView and SceneView use as their GeoModel. This is correct as written.

uitools/common/src/BasemapGalleryController.cpp
```diff
@@ -601,25 +601,28 @@ namespace Esri::ArcGISRuntime::Toolkit
...
-      switch (svts)
-      {
-        case SceneViewTilingScheme::Geographic:
-          return sp2.isGeographic();
+        const auto sp2 = basemap->baseLayers()->first()->spatialReference();
+        if (sp2.isEmpty()) //case used by the listview painter
+        {
+          return true;
+        }
+        auto svts = scene->sceneViewTilingScheme();
```
ldanzinger: I know these were the original variable names, but I think better variable names could provide clarity to what we are comparing, the actual current geomodel's spatial reference vs the basemap gallery basemap's spatial reference, etc

uitools/common/src/BasemapGalleryController.cpp
```diff
@@ -601,25 +601,28 @@ namespace Esri::ArcGISRuntime::Toolkit
...
 
-        case SceneViewTilingScheme::WebMercator:
-          return sp2 == SpatialReference::webMercator();
+          case SceneViewTilingScheme::WebMercator:
+            return sp2 == SpatialReference::webMercator();
 
-        default:
-          qDebug() << "a new sceneviewTilingScheme has been used";
-          break;
+          default:
```
ldanzinger: do we want to account for the new Automatic value in this switch?

uitools/common/src/BasemapGalleryController.cpp
```diff
@@ -640,10 +643,13 @@ namespace Esri::ArcGISRuntime::Toolkit
     {
       setGeoModel(sceneView->arcGISScene());
     }
+    else if (auto localSceneView = qobject_cast<LocalSceneViewToolkit*>(view))
```
ldanzinger: Have you tested this component with local scene view? One of my concerns is that the LSV doesn't support VTL yet, so this tool will be of limited use for now. I guess people could set their own basemaps in it that only use support layers, but I wonder if we want to think more deeply about how to handle it. For example, knowing that LSV doesn't support VTL, and knowing that most basemaps other than imagery use VTL, should we try and indicate through the UI which basemaps actually can be used?

uitools/common/src/OverviewMapController.cpp
```diff
@@ -162,17 +210,17 @@ namespace Esri::ArcGISRuntime::Toolkit
       {
         if (!e.isEmpty())
         {
-          qDebug() << "Error. Main map did not load" << e.message() << e.additionalMessage();
+          qDebug() << "Error. Main scene did not load" << e.message() << e.additionalMessage();
```
ldanzinger: not sure this should've been changed

ldanzinger (commented):
overall, looks ok, but I'd like to understand better what testing has been done for each and if there are design implications we need to consider for each, and not just adding another switch case for the new view type. For example, does everything work with basemap gallery, bookmarks, etc with LSV? Are there special cases to consider to the UX of the tool for LSV given some of the layer limitations that exist today?

uitools/common/src/GeoViews.h
```diff
@@ -18,23 +18,27 @@
 
 #ifdef WIDGETS_ARCGISRUNTIME_TOOLKIT
 
+#include <LocalSceneGraphicsView.h>
```
GuillaumeBelzPro: `LocalSceneGraphicsView` doesn't exist. It's `LocalSceneWidget`.

GuillaumeBelzPro:
> I just did a search for "SceneView" and "SceneViewToolkit" and added LocalSceneView support any time "SceneView" was mentioned.

Unfortunately, that's probably not enough. The derived classes can be used directly in the code, not only MapViewToolkit, SceneViewToolkit, and LocalSceneViewToolkit. For example https://github.com/Esri/arcgis-maps-sdk-toolkit-qt/blob/main/uitools/toolkitwidgets/src/Esri/ArcGISRuntime/Toolkit/CoordinateConversion.h#L49.

I think it's necessary to looking for derived classes too: MapView, SceneView, LocalSceneView, MapQuickView, MapGraphicsView, SceneQuickView, SceneGraphicsView, LocalSceneQuickView, and LocalSceneWidget.

uitools/common/src/BasemapGalleryController.cpp
```diff
@@ -565,72 +566,39 @@ namespace Esri::ArcGISRuntime::Toolkit
...
       }
-    }
-
-    const auto layers = basemap->baseLayers();
-    if (layers->size() <= 0)
-    {
-      return false;
-    }
+      return geoModel->spatialReference();
+    }(m_geoModel);
```
tanneryould: This new pattern follows the [Swift implementation](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/v.next/Sources/ArcGISToolkit/Components/BasemapGallery/BasemapGalleryViewModel.swift#L224-L237)

```swift
extension SpatialReferenceMismatchError: Equatable {}

extension GeoModel {
    /// The actual spatial reference of the geo model.
    /// - Remark:
    /// - For `Map`, it is map's `spatialReference`.
    /// - For `Scene`, if the `sceneViewTilingScheme` is `webMercator`, then `actualSpatialReference`
    /// is `webMercator`. Otherwise scene's `spatialReference`.
    var actualSpatialReference: SpatialReference? {
        if let scene = self as? ArcGIS.Scene,
           scene.tilingScheme == .webMercator {
            return .webMercator
        } else {
            return spatialReference
        }
    }
}
```

uitools/common/src/OverviewMapController.cpp
tanneryould: This is in reference to the scene geoview, not the map inset

tanneryould:
At this point, this PR makes Local Scenes work in Toolkit components. We'll likely need to do a follow up to review design, documentation, and any special cases for Local Scenes to include.

uitools/toolkitwidgets/src/Esri/ArcGISRuntime/Toolkit/CoordinateConversion.cpp
```diff
@@ -120,6 +121,17 @@ namespace Esri::ArcGISRuntime::Toolkit
     m_controller->setGeoView(sceneView);
   }
 
+    /*!
+    \brief Set the \c GeoView.
+    \list
+      \li \a localSceneView Sets the \c GeoView to a \c LocalSceneView.
+    \endlist
+   */
+  void CoordinateConversion::setLocalSceneView(LocalSceneWidget* localSceneView)
```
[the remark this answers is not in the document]
tanneryould: A `LocalSceneWidget` is an alias for `LocalSceneView` in widgets

uitools/common/src/BasemapGalleryController.cpp
```diff
@@ -565,72 +566,39 @@ namespace Esri::ArcGISRuntime::Toolkit
...
-        case SceneViewTilingScheme::WebMercator:
-          return sp2 == SpatialReference::webMercator();
-
-        default:
-          qDebug() << "a new sceneviewTilingScheme has been used";
-          break;
-      }
-      return false;
+      return true;
     }
```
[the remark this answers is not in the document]
tanneryould: @koushik-hajra if a Global Scene view uses `SceneViewTilingScheme::Geographic`, can it accept any layer with a Geographic SR?
If so, Swift's implementation has a bug.

uitools/common/src/BasemapGalleryController.cpp
```diff
@@ -565,72 +566,53 @@ namespace Esri::ArcGISRuntime::Toolkit
 
   bool BasemapGalleryController::basemapMatchesCurrentSpatialReference(Basemap* basemap) const
   {
-    if (!basemap)
+    if (!basemap || basemap->baseLayers()->isEmpty())
```
GuillaumeBelzPro: `basemap->baseLayers()` is not supposed to be nullptr. In this case, I like to have an assert to guarantee that.

uitools/common/src/BasemapGalleryController.cpp
```diff
@@ -565,72 +566,53 @@ namespace Esri::ArcGISRuntime::Toolkit
...
+      if (scene->sceneViewTilingScheme() == SceneViewTilingScheme::Geographic)
       {
-        return sp == item->spatialReference();
+        return basemapSR.isGeographic();
       }
     }
 
-    const auto layers = basemap->baseLayers();
-    if (layers->size() <= 0)
+    SpatialReference geoModelSR = [](GeoModel* geoModel)
```
GuillaumeBelzPro: Can be const?

uitools/common/src/BookmarksViewController.cpp
```diff
@@ -187,6 +189,15 @@ namespace Esri::ArcGISRuntime::Toolkit
           disconnect(scene->bookmarks(), nullptr, m_bookmarks, nullptr);
         }
       }
+      else if (auto localSceneView = qobject_cast<LocalSceneViewToolkit*>(m_geoView))
+      {
+        auto scene = localSceneView->arcGISScene();
```
GuillaumeBelzPro: Is clang-tidy enabled on toolkit? Usually, it recommends to use `auto*` here.

tanneryould: It recommends `auto*` on a _lot_ of places. Would you like me to update them all? Because I can, I just didn't want to bloat this PR

uitools/common/src/CoordinateConversionController.cpp
```diff
@@ -127,6 +127,22 @@ namespace Esri::ArcGISRuntime::Toolkit
         }
       });
     }
+    else if (auto localSceneView = qobject_cast<LocalSceneViewToolkit*>(m_geoView))
+    {
+      connect(localSceneView, &LocalSceneViewToolkit::mouseClicked, this, [localSceneView, this](QMouseEvent& event)
```
GuillaumeBelzPro, line 132: General comment for all toolkit. After a geoview is used with coordinate conversion controller, it's never disconnect until the controller or the view is deleted. If there are several geoviews or dynamic creation/deletion, maybe it would need a way to disconnect?

uitools/examples/src/demos/BasemapGalleryDemo.h
```diff
@@ -44,10 +45,12 @@ class BasemapGalleryDemo : public BaseDemo
 
 protected:
   Esri::ArcGISRuntime::Map* initMap_(QObject* parent) const override;
-  Esri::ArcGISRuntime::Scene* initScene_(QObject* parent) const override;
+  Esri::ArcGISRuntime::Scene* initGlobalScene_(QObject* parent) const override;
+  Esri::ArcGISRuntime::Scene* initLocalScene_(QObject* parent) const override;
 
 private:
   Esri::ArcGISRuntime::Toolkit::BasemapGalleryController* m_controller;
+  QScopedPointer<QObject> m_parent;
```
GuillaumeBelzPro: Usually, we prefer std::unique_ptr, rather than Qt's smart ptrs.

uitools/examples/src/demos/BasemapGalleryDemo.cpp
```diff
@@ -24,12 +23,15 @@
...
+#include "SceneViewTypes.h"
+
+using namespace Esri::ArcGISRuntime;
 
 BasemapGalleryDemo::BasemapGalleryDemo(QObject* parent) :
   BaseDemo(parent),
   m_controller(new Esri::ArcGISRuntime::Toolkit::BasemapGalleryController(this))
 {
-  using namespace Esri::ArcGISRuntime;
+  m_parent.reset(new QObject(this)); // Used as the parent for the initial GeoModels
```
GuillaumeBelzPro: Smart pointers and QObject parent-children relationship are not compatible. If you're using smart ptr, don't assign a QObject parent.

uitools/common/src/BookmarksViewController.cpp
GuillaumeBelzPro: No, it's fine. There is no requirement (for the moment) for clang-tidy in toolkit.

But there is no guideline against using `auto` in samples and toolkit? Maybe it's just for samples?

tanneryould: We avoid using `auto` in Samples, _for now_, but we use it freely in the Toolkit

uitools/examples/src/proxies/LocalSceneQuickViewProxy.cpp
```diff
@@ -0,0 +1,57 @@
...
+  m_scene(nullptr)
+{
+  if (m_localSceneQuickView)
+  {
+    using namespace Esri::ArcGISRuntime;
+    connect(m_localSceneQuickView, &LocalSceneQuickView::sceneChanged, this, &LocalSceneQuickViewProxy::sceneChangedProxy);
+  }
+}
+
+LocalSceneQuickViewProxy::~LocalSceneQuickViewProxy()
```
GuillaumeBelzPro: `= default`?

uitools/examples/src/main.cpp
```diff
@@ -64,6 +66,7 @@ int main(int argc, char* argv[])
   // Register ArcGIS types with QML.
   qmlRegisterExtendedType<Esri::ArcGISRuntime::MapQuickView, MapQuickViewProxy>("Esri.Examples", 200, 0, "MapView");
   qmlRegisterExtendedType<Esri::ArcGISRuntime::SceneQuickView, SceneQuickViewProxy>("Esri.Examples", 200, 0, "SceneView");
+  qmlRegisterExtendedType<Esri::ArcGISRuntime::LocalSceneQuickView, LocalSceneQuickViewProxy>("Esri.Examples", 200, 0, "LocalSceneView");
```
GuillaumeBelzPro: `200`? Not `300`?

GuillaumeBelzPro (approved):
Looks good. Just some more comments

uitools/examples/src/demos/BasemapGalleryDemo.h
tanneryould: I changed the way this is handled, but thank you for reminding me of this

uitools/examples/src/main.cpp
tanneryould: Good catch!

uitools/common/src/CoordinateConversionController.cpp
tanneryould: Every demo has this code block for `setGeoView` so that should cover us?

```cpp
  void NorthArrowController::setGeoView(QObject* geoView)
  {
    if (geoView == m_geoView)
    {
      return;
    }

    if (m_geoView)
    {
      disconnect(m_geoView, nullptr, this, nullptr);
    }

    m_geoView = geoView;

    // and so on...
```

uitools/examples/src/proxies/LocalSceneQuickViewProxy.cpp
tanneryould: The other QuickViewProxy classes use the other pattern. Would you like me to update those as well?
https://github.com/Esri/arcgis-maps-sdk-toolkit-qt/blob/main/uitools/examples/src/proxies/MapQuickViewProxy.cpp#L41-L43

uitools/common/src/CoordinateConversionController.cpp
GuillaumeBelz: I missed that 👍

uitools/common/src/BasemapGalleryController.cpp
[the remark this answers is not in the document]
koushik-hajra: @tanneryould If the scene's tiling scheme was already Geographic, then you can only add any basemap that is in geographic tiling scheme. Any spatial reference in that tiling scheme should work.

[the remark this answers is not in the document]
koushik-hajra: But we should test that out to be sure.

uitools/common/src/BookmarksViewController.cpp
```diff
@@ -187,6 +189,15 @@ namespace Esri::ArcGISRuntime::Toolkit
           disconnect(scene->bookmarks(), nullptr, m_bookmarks, nullptr);
         }
       }
+      else if (auto* localSceneView = qobject_cast<LocalSceneViewToolkit*>(m_geoView))
+      {
+        auto scene = localSceneView->arcGISScene();
```
ldanzinger: strange how auto* showed up in so many other places, but not here where the comment was made

uitools/common/src/NorthArrowController.cpp
```diff
@@ -43,30 +43,41 @@ namespace Esri::ArcGISRuntime::Toolkit
...
     {
       Camera currentCamera = sceneView->currentViewpointCamera();
       Camera updatedCamera = currentCamera.rotateTo(heading, currentCamera.pitch(), currentCamera.roll());
       auto future = sceneView->setViewpointCameraAsync(updatedCamera, 0.50);
       Q_UNUSED(future)
     }
+    else if (auto* localSceneView = qobject_cast<LocalSceneViewToolkit*>(m_geoView))
+    {
+      Camera currentCamera = localSceneView->currentViewpointCamera();
+      Camera updatedCamera = currentCamera.rotateTo(heading, currentCamera.pitch(), currentCamera.roll());
```
ldanzinger: could be marked const

uitools/examples/qml/demos/BasemapGalleryDemoForm.qml
```diff
@@ -44,6 +44,29 @@ DemoPage {
             onSceneChanged: gallery.setGeoModelFromGeoView(view)
         }
     }
+    localSceneViewContents: Component {
+        LocalSceneView {
+            id: view
+
+            //! [Set up BasemapGallery QML]
```
ldanzinger: I'm not aware of if/where we use this qdoc tag, but we shouldn't use an identical tag that is used in the same file as that could break the doc build if used.

uitools/examples/qml/demos/CoordinateConversionDemoForm.qml
```diff
@@ -42,6 +42,27 @@ DemoPage {
             }
         }
     }
+    localSceneViewContents: Component {
+        LocalSceneView {
+            id: view
+
+            //! [Set up Coordinate Conversion QML]
```
ldanzinger: same comment here about the qdoc tag. I would just remove it

uitools/examples/qml/demos/NorthArrowDemoForm.qml
```diff
@@ -42,6 +42,26 @@ DemoPage {
             }
         }
     }
+    localSceneViewContents: Component {
+        LocalSceneView {
+            id: view
+
+            //! [Set up North Arrow QML]
```
ldanzinger: suggest removing

uitools/examples/qml/demos/OverviewMapDemoForm.qml
```diff
@@ -41,6 +41,27 @@ DemoPage {
             }
         }
     }
+    localSceneViewContents: Component {
+        LocalSceneView {
+            id: view
+
+            //! [Set up Overview Map QML]
```
ldanzinger: suggest removing

uitools/examples/src/proxies/LocalSceneQuickViewProxy.cpp
```diff
@@ -0,0 +1,55 @@
+/*******************************************************************************
+ *  Copyright 2012-2022 Esri
```
ldanzinger: new files should have 2026 copyright date

uitools/common/src/OverviewMapController.h
```diff
@@ -62,14 +62,17 @@ namespace Esri::ArcGISRuntime::Toolkit
...
-    void applyMapNavigationToInset(MapViewToolkit* view);
-    void applySceneNavigationToInset(SceneViewToolkit* view);
+    void applyMapViewNavigationToInset(MapViewToolkit* view);
+    void applySceneViewNavigationToInset(SceneViewToolkit* view);
+    void applyLocalSceneViewNavigationToInset(LocalSceneViewToolkit* view);
 
     void disableInteractions();
 
-    void setupInsetMapForMap(MapViewToolkit* mapView);
-    void setupInsetMapForScene(SceneViewToolkit* sceneView);
```
ldanzinger, line 72: this is grey area that James pointed out in scrum. Our controllers are technically public API in the sense that they can be used, but we don't document them. We have made changes to them in the past when we really needed to. I think we should consider avoiding breaking if we can reasonably avoid it. I think these changes are  simply just for naming preferences, so I'd probably opt for leaving them as-is instead of changing the public interface

uitools/common/src/BookmarksViewController.cpp
tanneryould: Strange indeed 😅

uitools/common/src/OverviewMapController.h
tanneryould: Good call!

tanneryould:
@ldanzinger and @JamesMBallard -- this is ready for re-review if you'd like to look it over once again. I know the scope of this PR has inflated quite a bit, so let me know if you'd like me to set up a meeting where we can cover all of this.

ldanzinger (approved):

uitools/common/src/CoordinateConversionController.cpp
```diff
@@ -231,14 +262,21 @@ namespace Esri::ArcGISRuntime::Toolkit
 
   void CoordinateConversionController::zoomToCurrentPoint()
   {
-    if (auto sceneView = qobject_cast<SceneView*>(m_geoView))
+    if (auto* sceneView = qobject_cast<SceneViewToolkit*>(m_geoView))
```
JamesMBallard, line 265: We should be able to solve this more elegantly than expanding all the if-else checks for another type. However, that is a job for another day.

JamesMBallard (approved):

PR #810, opened 2026-04-02
Fix version for ASSERT: "isValidSegment(majorVersion)"  error.

min14751:
300.0 -> 3.0


tanneryould (approved):
Oh, good catch, thank you!

GuillaumeBelzPro (approved):

PR #811, opened 2026-04-03
QFuture state alone is not enough to prevent a feedback loop

JamesMBallard:
We were using QFuture::isRunning() as the factor to decide whether a `viewpointChanged` callback was from an interaction update, or from a controller-driven setViewpointAsync(...).

That caused a feedback loop. When one view was updated programmatically to follow the other, its own `viewpointChanged` signal could arrive after the future no longer reported as running. At that point the controller tried to sync back in the other direction.

These changes solidify that there's only one driver at a time:
 * If the main view is being interacted with by the user, the inset view only listens and reacts - it does not try to sync the main view with it.
 * And the same in reverse if the user is interacting with the inset view.

### DIFF
--- a/uitools/common/src/OverviewMapController.cpp
+++ b/uitools/common/src/OverviewMapController.cpp
@@ -322,7 +322,22 @@ namespace Esri::ArcGISRuntime::Toolkit
 
   bool OverviewMapController::shouldApplyNavigationUpdate(GeoView* geoView) const
   {
-    return geoView && geoView->isNavigating() && !m_isUpdatingGeoViewFromInset && !m_isUpdatingInsetFromGeoView;
+    if (!geoView || !geoView->isNavigating())
+    {
+      return false;
+    }
+
+    if (geoView == m_geoView)
+    {
+      return !m_isUpdatingGeoViewFromInset;
+    }
+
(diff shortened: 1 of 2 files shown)


uitools/common/src/OverviewMapController.cpp
```diff
@@ -309,14 +320,52 @@ namespace Esri::ArcGISRuntime::Toolkit
     }
   }
 
+  bool OverviewMapController::shouldApplyNavigationUpdate(GeoView* geoView) const
```
GuillaumeBelzPro: Not required, but I found that confusing to have geoView, m_geoView (which is a QObject, not a GeoView), and m_insetView (which can be geoView too). Maybe a comment can help to understand this function.

JamesMBallard: You are right. In fact I think remove this entire helper will make it easier to follow. I tried to have a helper since all the calls into this use isNavigating() from the GeoView base, but it doesn't follow the pattern of the rest of the file.

GuillaumeBelzPro (approved):
Looks good

har13205 (approved):

aji13187 (approved):

PR #814, opened 2026-04-06
Allow map rotation by touch.

JamesMBallard:
In order to make sure the slider stays in sync,
we update the slider value when the map
rotation has changed and the slider is not
pressed by the user.


uitools/examples/src/demos/NorthArrowDemo.cpp
```diff
@@ -16,18 +17,37 @@
 
 #include "NorthArrowDemo.h"
 
+#include <MapQuickView.h>
+#include <MapViewTypes.h>
+
 #include <QFuture>
```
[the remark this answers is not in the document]
JamesMBallard: Close, but it is used.

bri12415 (approved):

bri12415:
> In order to make sure the slider stays in sync,
we update the slider value when the map
rotation has changed and the slider is not
pressed by the user.

Is this a haiku?

uitools/examples/src/demos/NorthArrowDemo.cpp
```diff
@@ -16,18 +16,43 @@
...
+  if (!mapView)
+  {
+    return;
+  }
+
+  auto future = mapView->setViewpointRotationAsync(degrees);
+  Q_UNUSED(future);
+}
+
+void NorthArrowDemo::setup()
```
GuillaumeBelzPro: maybe "setup" is not a very explicit/informative name?

GuillaumeBelzPro (approved):
looks good

uitools/examples/src/demos/NorthArrowDemo.cpp
JamesMBallard: Good point, but I was following the pattern of another demo I just worked on. I'll use a better name.

JamesMBallard:
Upon retesting I found a problem. To allow the slider and touch rotation to both rotate, and always keep the slider in sync with the current rotation value, we need to have some additional state so that slider changes don't end up reacting to itself.

Or in Haiku courtesy of copilot:
 
Retest found trouble
State guards slider from itself
Both now stay in sync

uitools/examples/src/demos/NorthArrowDemo.cpp
```diff
@@ -16,18 +16,73 @@
...
+  auto* mapView = static_cast<MapQuickView*>(view);
+  mapView->setRotationByPinchingEnabled(true);
+  setDrawingComplete_(mapView->drawStatus() == DrawStatus::Completed);
+  m_drawStatusChangedConnection = connect(mapView, &MapQuickView::drawStatusChanged, this, [this](DrawStatus drawStatus)
+  {
+    setDrawingComplete_(drawStatus == DrawStatus::Completed);
+  });
+}
+
+void NorthArrowDemo::setDrawingComplete_(bool completed)
```
bri12415, line 79: Is it possible to have the QML just rely on the `&MapQuickView::drawStatusChanged` signal rather than us propagating that in C++ through `drawingCompleteChanged()` ? I.e., just check in QML if the `&MapQuickView::drawStatusChanged` is `completed`.

Not sure if that's any better than what you have here though, or if I'm missing some logic that requires us to do it this way.

bri12415 (approved):

tanneryould (approved):
LGTM -- I tested this out on Windows in the Demo app and Map, Scene, and LocalScene Views look good. I know this was a change on touch interaction, but I wanted to make sure there were no regressions on desktop.

uitools/examples/src/demos/NorthArrowDemo.cpp
JamesMBallard: Yes, it's possible but we'd need to make the enum available to QML which has some additional boilerplate that comes along with it. I don't want to overcomplicate what should be a simple demo. A half-way option is to just represent the enum as an int, but then it doesn't read as nicely.

PR #815, opened 2026-04-08
Jared/fix uitools app2

jared-2016:
So the backend guts is nearly identical to what we do with the SampeViewer. 

cache ArcGISAuthenticationChallengeHandler at the very beginning
https://github.com/Esri/arcgis-maps-sdk-samples-qt/blob/2dfe28360a1d151630915141d359537835c3cd52/SampleViewer/SampleManager.cpp#L423-L430

resetAuthenticationState
which will set the arcGISAuthenticationChallengeHandler to it's initial state for all demos
https://github.com/Esri/arcgis-maps-sdk-samples-qt/blob/2dfe28360a1d151630915141d359537835c3cd52/SampleViewer/SampleManager.cpp#L490-L510

Added a member variable  to cache the api key in the ArcGISRuntimeEnvironment proxy so that we don't have to re-prompt the user for the API key when switching between demos.

The workflow is
1. Cache ArcGISAuthenticationChallengeHandler(only the first time)
2. set api key
3. reset authentication state
4. switch demo
5. repeat steps 2 - 4

QML was updated to orchestrate the above.

There is one thing to note with this fix. It only works because of Qt.callLater when switching the demos and I am not exactly sure why. Looking at the doc it states:

> The function passed as the first argument to Qt.[callLater](https://doc.qt.io/qt-6/qml-qtqml-qt.html#callLater-method)() will be called later, once the QML engine returns to the event loop.

Without the above the creation/destruction of Demos overlaps which prevents us from being able to restore the arcGISAuthenticationChallengeHandler because the new Demo will be created before the previous demo has been destructed. I am unsure why using Qt.callLater fixes that ordering and also results in this resetting of the authentication state after the destruction and before the construction of the new demo. I did quite a bit of testing with logging and the proposed changes works.

This is not an issue with the SampleViewer because we handle the Sample switching differently which doesn't have this odd overlapping state issue.


uitools/examples/qml/DemoApp.qml
```diff
@@ -21,6 +21,28 @@ import Esri.ArcGISRuntime.Toolkit
...
+        currentDemoSource = "";
+
+        Qt.callLater(function () {
+            const nextIndex = pendingDemoIndex;
+            pendingDemoIndex = -1;
+
+            ArcGISRuntimeEnvironment.prepareForDemoSwitch();
+            viewList.currentIndex = nextIndex;
+            currentDemoSource = listModel.get(nextIndex).url;
+        });
```
jared-2016, line 43: Using this makes sure the existing demo has been unloaded and destructed prior to calling `ArcGISRuntimeEnvironment.prepareForDemoSwitch()` which sets the api key and resets the authentication state. Then it sets the demo source which initiates the creation of the next demo. As mentioned in the PR description, I am not certain why this works. https://doc.qt.io/qt-6/qml-qtqml-qt.html#callLater-method

uitools/examples/src/proxies/ArcGISRuntimeEnvironmentProxy.cpp
```diff
@@ -36,17 +45,52 @@ ArcGISRuntimeEnvironmentProxy::~ArcGISRuntimeEnvironmentProxy()
...
 {
-  using namespace Esri::ArcGISRuntime;
-  return ArcGISRuntimeEnvironment::apiKey();
+  return Esri::ArcGISRuntime::ArcGISRuntimeEnvironment::apiKey();
 }
 
 void ArcGISRuntimeEnvironmentProxy::setApiKey(const QString& apiKey)
 {
   using namespace Esri::ArcGISRuntime;
+  const auto normalizedApiKey = apiKey.trimmed();
```
jared-2016, line 52: This is pretty cool. It trims accidental leading and trailing whitespace. I guess that could be intentional but seems really unlikely.

uitools/examples/src/proxies/ArcGISRuntimeEnvironmentProxy.cpp
```diff
@@ -16,17 +17,25 @@
 
 #include "ArcGISRuntimeEnvironmentProxy.h"
 
+#include "OAuthUserConfigurationManager.h"
+
 #include <ArcGISRuntimeEnvironment.h>
+#include <Authentication/ArcGISCredentialStore.h>
+#include <Authentication/AuthenticationManager.h>
+#include <Authentication/NetworkCredentialStore.h>
```
JamesMBallard: This is probably overkill, I don't think any part of the demo is dealing with network credentials.

uitools/examples/src/proxies/ArcGISRuntimeEnvironmentProxy.cpp
```diff
@@ -36,17 +45,58 @@ ArcGISRuntimeEnvironmentProxy::~ArcGISRuntimeEnvironmentProxy()
...
+  resetAuthenticationState();
+}
+
+void ArcGISRuntimeEnvironmentProxy::resetAuthenticationState()
+{
+  using namespace Esri::ArcGISRuntime;
+
+  ArcGISRuntimeEnvironment::authenticationManager()->arcGISCredentialStore()->removeAll();
+  auto removeAllFuture = ArcGISRuntimeEnvironment::authenticationManager()->networkCredentialStore()->removeAllAsync();
+  Q_UNUSED(removeAllFuture)
```
JamesMBallard: Probably overkill to do anything with network credentials. Even the authentication demo is just token (ArcGIS authentication).

JamesMBallard (commented):
This seems fine. The obvious solution is to have the UN demo do all this work and restore the previous challenge handler, but I remember we tried that with the samples and it didn't work.

JamesMBallard (approved):

uitools/examples/src/proxies/ArcGISRuntimeEnvironmentProxy.cpp
```diff
@@ -1,3 +1,4 @@
+
```
GuillaumeBelzPro: remove empty line

GuillaumeBelzPro (approved):
looks good

uitools/examples/src/proxies/ArcGISRuntimeEnvironmentProxy.cpp
jared-2016: git hooks likes to do this too me. 
image

PR #841, opened 2026-08-12
Guil8553/clang tidy

GuillaumeBelzPro:
Fix Clang-tidy and CodeQL warnings

### DIFF
--- a/uitools/common/src/BookmarksViewController.cpp
+++ b/uitools/common/src/BookmarksViewController.cpp
@@ -37,65 +37,60 @@
 
 namespace Esri::ArcGISRuntime::Toolkit
 {
-
-  namespace
+  static void setupBookmarks(BookmarkListModel* sourceModel, GenericListModel* targetModel)
   {
-    void setupBookmarks(BookmarkListModel* sourceModel, GenericListModel* targetModel)
+    QObject::connect(sourceModel, &BookmarkListModel::rowsInserted, targetModel,
+                     [sourceModel, targetModel](const QModelIndex& parent, int first, int last)
     {
-      QObject::connect(sourceModel, &BookmarkListModel::rowsInserted, targetModel,
-                       [sourceModel, targetModel](const QModelIndex& parent, int first, int last)
+      if (parent.isValid())
       {
-        if (parent.isValid())
-        {
-          return;
-        }
-
-        if (!targetModel->insertRows(first, last - first + 1))
-        {
-          return;
-        }
-
-        for (auto i = first; i <= last; ++i)
-        {
-          QModelIndex j = targetModel->index(i);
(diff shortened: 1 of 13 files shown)


uitools/common/src/BookmarksViewController.cpp
```diff
@@ -37,107 +37,60 @@
 
 namespace Esri::ArcGISRuntime::Toolkit
 {
-
-  namespace
+  static void setupBookmarks(BookmarkListModel* sourceModel, GenericListModel* targetModel)
```
GuillaumeBelzPro, line 40: llvm-prefer-static-over-anonymous-namespace

uitools/common/src/BookmarksViewController.cpp
```diff
@@ -37,107 +37,60 @@
...
-      Will continue to call `f` every time a mapChanged/sceneChanged signal is triggered on
-      the GeoView.
-     */
-    template<typename GeoViewToolkit, typename Func>
-    void connectToGeoView(GeoViewToolkit* geoView, BookmarksViewController* self, Func&& f)
-    {
-      static_assert(std::is_same<GeoViewToolkit, MapViewToolkit>::value
-                      || std::is_same<GeoViewToolkit, SceneViewToolkit>::value
-                      || std::is_same<GeoViewToolkit, LocalSceneViewToolkit>::value,
-                    "Must be connected to a SceneView, LocalSceneView, or MapView");
```
JamesMBallard, line 112: The new lambda approach drops the static_assert. Any concerns? I wouldn't think so since there are only three possible types.

GuillaumeBelzPro: The lambda is an inner function and can't be used outside the existing code. 

It's possible to keep the static_assert using decltype.

JamesMBallard (approved):

uitools/common/src/BookmarksViewController.cpp
```diff
@@ -208,6 +161,37 @@ namespace Esri::ArcGISRuntime::Toolkit
...
+      auto connectToGeoModel = [this, typedGeoView, f]()
+      {
+        auto model = getGeoModel(typedGeoView);
+        if (!model)
+        {
+          return;
+        }
+
+        // Call `f` once the GeoModel is loaded.
+        auto c = doOnLoaded(model, this, [f]()
```
bri12415, line 180: This is pedantic, but I think `c` and `f` could be named better.

bri12415 (approved):

uitools/common/src/BookmarksViewController.cpp
GuillaumeBelz: This comes from the orignal code.

PR #844, opened 2026-09-21
Toolkit should be importing WebEngineQuick as it brings in the dependency

jared-2016:
added WebEngineQuick to the toolkit QT modules and cmake equivalent.

Moved Qt6::WebView to Public for CMake which results in the usage requirements to be propagated to the consumer e.g. 
```c++
// initialize needs to be called immediately before QGuiApplication instance is created
// https://doc.qt.io/qt-6/qtwebview.html#initialize
QtWebView::initialize();
```


uitools/toolkitcpp/CMakeLists.txt
```diff
@@ -19,6 +19,7 @@ include(../common/CMakeLists.txt)
 set(CMAKE_INCLUDE_CURRENT_DIR ON)
 
 find_package(Qt6 REQUIRED COMPONENTS Core Gui QuickControls2 Svg WebView Graphs NetworkAuth)
+find_package(Qt6 QUIET COMPONENTS WebEngineQuick)
```
ldanzinger, line 22: why QUIET?

uitools/toolkitcpp/CMakeLists.txt
```diff
@@ -43,12 +44,23 @@ target_link_libraries(libtoolkitcpp PRIVATE
...
     Qt6::Svg
-    Qt6::WebView
     Qt6::Graphs
     Qt6::NetworkAuth
     ArcGISRuntime::Cpp
 )
 
+# Consumers must initialize the Qt Quick web modules before constructing QGuiApplication.
+# https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html#target-usage-requirements
+target_link_libraries(libtoolkitcpp PUBLIC
```
ldanzinger, line 54: why does this not need to be conditionalized like the WebEngineQuick check does?

ldanzinger (commented):
couple questions for you @jared-2016

uitools/toolkitcpp/CMakeLists.txt
jared-2016: WebEngineQuick is not available for all platforms(i.e. iOS, Android) if we make it `REQUIRED` this will fail when running cmake for those target platforms. So this won't display any warnings or errors when trying to find the pacakge. This was mostly modeled after how we approached it in our code previously for qmake which just silently did nothing when not found. Maybe we should consider making this `REQUIRED` but just make it platform gated where it's expected so if it's not there, the user is notified as they are with the other packages.

I am in favor of the platform gating that makes this required so it doesn't go by unnoticed.

uitools/toolkitcpp/CMakeLists.txt
jared-2016: WebView is expected to there for all platforms as it was before. WebEngineQuick will not be available for all platforms(i.e. iOS, Android) as mentioned in the other comment.

jared-2016:
converted back to draft to test on all platforms.

